PR #18, opened 2017-02-02
write: missing buffer reuse in float32, cleanup

vavrusa:
Just cleanup, implemented the use of new buffer for float32 and cleaned up repetitive code.


kshvakov:
thank you

PR #19, opened 2017-02-10
block: fixed partial write in 8f94710

vavrusa:
The loop returned nil in case of success, which terminated the write loop for all columns too early if the written enum wasn’t the last column in the set.


kshvakov:
Thank you

PR #29, opened 2017-04-20
Support multi-insert

Upliner:


Upliner:
This patch adds support for statements like INSERT INTO table(a,b,c) VALUES (?,?,?), (?,?,?)

kshvakov:
That is unnecessary

kshvakov:
@jakesyl Hi, see an example https://github.com/ClickHouse/clickhouse-go/blob/master/examples/simple.go#L38

jakesyl:
@kshvakov Appreciate the help. Is this faster than columnar?

kshvakov:
No, but it’s more convenient

kshvakov:
I created [this benchmark]( https://github.com/kshvakov/go-and-clickhouse/tree/master/benchmark) for [Altinity](https://www.altinity.com/)

PR #40, opened 2017-06-03
columnar writer interface

vavrusa:
Hi, this does some optimisations and implements a columnar writer interface for high throughput insertions. See the commit messages for more information, but here's the gist:

* writeBuffer growing strategy is now O(1) and uses just normal doubling of previous chunk size
* new interface for statement to support inserts directly into columns

For my bulk insertions case it's about 2x faster overall with half the memory usage and less gc pauses.

### DIFF
--- a/block.go
+++ b/block.go
@@ -188,20 +188,25 @@ func (b *block) write(revision uint64, w io.Writer) error {
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


kshvakov:
Great, thank you

PR #44, opened 2017-06-07
write_buffer: recycle buffer chunks, not buffers

vavrusa:
since aae5960 only the slice holding memory
chunks was recycled, not the data backing the
buffers

this moves memory recycling to write_buffer and
recycles memory chunks backing the buffer.
to play nice with memory growing strategy,
smallest chunks are released instead of recycling,
which makes memory pool keep larger chunks when
the memory usage rises

### DIFF
--- a/block.go
+++ b/block.go
@@ -5,15 +5,7 @@ import (
 	"fmt"
 	"io"
 	"strings"
-	"sync"
 )
-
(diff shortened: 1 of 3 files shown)


vavrusa:
The CI build fails because gocoveralls is misconfigured I think.

kshvakov:
thank you

PR #46, opened 2017-06-27
fixed writing arrays in long inserts, support []byte in arrays

vavrusa:


vavrusa:
The tests pass, coverage collection script fails for some reason.

PR #50, opened 2017-07-10
additional int conversions in writeBuffer.write

Vespertinus:
and integers to Date & DateTime conversion

That changes allow to almost transparently transmit data from msgpack & cbor decoders to clickhouse


kshvakov:
@Vespertinus thank you, good job!

PR #54, opened 2017-08-18
Refactoring

kshvakov:


connect.go
```diff
@@ -19,16 +19,18 @@ func dial(network string, hosts []string, noDelay bool, r, w time.Duration, logf
 			return v
 		}
 		conn  net.Conn
-		index = abs(int(atomic.AddInt32(&tick, 1)))
+		ident = abs(int(atomic.AddInt32(&tick, 1)))
```
vavrusa: It'd be nice not to make tick global, or to not force it's use. I'm happy as a caller to provide a list of hosts sorted by my preference, but the library will not actually use them in the order because of the advancing ticker.

write_column.go
```diff
@@ -14,20 +15,13 @@ type ColumnarStatement interface {
...
 func (stmt *stmt) ColumnWriterEnd(rows uint64) error {
-	writtenBlocks := stmt.counter / stmt.ch.blockSize
-	stmt.counter += int(rows)
-	stmt.ch.data.numRows += rows
-
-	var err error
-	if stmt.counter/stmt.ch.blockSize > writtenBlocks {
-		err = stmt.ch.data.write(stmt.ch.serverRevision, stmt.ch.conn)
-	}
-	return err
```
vavrusa: Is there a way how to force chunking of input into blocks? The `blockSize` put an upper bound on how much memory would data would be in buffers before sending it to server. Is there a way to control when a block should be written and its structure recycled? Something like:

```
stmt.ColumnWriterEnd(1)
if stmt.NumRows > 100000 {
    stmt.WriteAndReset()
}
```

kshvakov: Oops, I'm sorry, I'll restore this feature )

connect.go
kshvakov: Do you mean to use alt_host only for failover and try to connect to the first server of the list for each attempt to open driver?

vavrusa: Yes, exactly.

write_column.go
vavrusa: Actually, it might be better to just allow multiple blocks so caller can control both size and write them concurrently with the columnar interface, like:

```
b1 := stmt.Block() // Default "ch.block"
b2 := stmt.CreateBlock() // Create extra block and return it
wg := sync.WaitGroup()
wg.Add(2)
go func() {
  b1.WriteUInt8(0, 42)
  wg.Done()
}
go func() {
  b2.WriteUInt8(0, 43)
  wg.Done()
}
wg.Wait()
stmt.Commit() // Write both blocks, followed by an empty block to signalise EOF
```

What do you think? @kshvakov

kshvakov: Yes, that makes sense, but I would like to make the second interface for the driver (without support database/sql) to provide this functionality. I think, will be better to create new issue for it )

PR #56, opened 2017-08-20
Use Docker image we actually pulled

AlekSi:


kshvakov:
Already in PR https://github.com/kshvakov/clickhouse/pull/54

PR #62, opened 2017-08-23
write blocks concurrently with columnar writer

vavrusa:
This changes the interface to enable accessing clickhouse connection and block interfaces directly, so the caller can create and write additional blocks concurrently. This replaces the `blockSize` functionality for me, as the maximum size of blocks can be controlled by the caller.


kshvakov:
Great thanks!

PR #96, opened 2018-03-29
enum: allow implicit conversion from uint8, uint16

### LINKED ISSUE #95: Enum types are missing conversion from uint8 when in arrays
The enum column does have a converter function for int8, but it's missing it for uint8 which is a bit counter-intuitive, so `WriteUInt8(id, 0)` into an enum column succeeds, but `WriteArray(id, types.NewArray([]uint8{0, 1}))` fails.

vavrusa:
The width must be matching, but implicit conversion of unsigned types
to enums should be allowed.

fixes #95


kshvakov:
Thank you!

PR #107, opened 2018-05-19
Mention batch inserts as key feature in README

agaurav:
So someone like me don't get confused and ends up exploring `http` interface with [clickhouse-bulk](https://github.com/nikepan/clickhouse-bulk)

and to avoid https://github.com/kshvakov/clickhouse/issues/28


kshvakov:
Thank you!

PR #110, opened 2018-07-11
Add Nullable support for writing data in blocks.

bocharov:
cc @vavrusa @dqminh @kvap @twalwyn

### DIFF
--- a/lib/binary/encoder.go
+++ b/lib/binary/encoder.go
@@ -19,12 +19,26 @@ type Encoder struct {
 	scratch [binary.MaxVarintLen64]byte
 }
 
+func (enc *Encoder) Nullable() error {
+	if _, err := enc.output.Write([]byte{0}); err != nil {
+		return err
+	}
+	return nil
+}
+
 func (enc *Encoder) Uvarint(v uint64) error {
 	ln := binary.PutUvarint(enc.scratch[:binary.MaxVarintLen64], v)
 	if _, err := enc.output.Write(enc.scratch[0:ln]); err != nil {
 		return err
 	}
 	return nil
+}
+
+func (enc *Encoder) UvarintNullable(v uint64) error {
+	if err := enc.Nullable(); err != nil {
+		return err
+	}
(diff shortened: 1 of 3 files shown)


lib/binary/encoder.go
```diff
@@ -19,6 +19,13 @@ type Encoder struct {
 	scratch [binary.MaxVarintLen64]byte
 }
 
+func (enc *Encoder) Nullable() error {
+	if _, err := enc.output.Write([]byte{0}); err != nil {
+		return err
+	}
+	return nil
```
kvap: This can just be `err := Write(...)` and then `return err` without the `if`.

bocharov: That is how everywhere else is done in the file, so I'm keeping consistency by being more explicit what we return..

dqminh:
Dont we have to detect if the value is null and write `1` instead of only write `0 + serialized_data` as now ?

Also doesnt clickhouse require us to write the null map first, before actually writing data out just like how array works now ? See https://github.com/yandex/ClickHouse/blob/d1785d92a7b067a53ef34632cad4adf93e28674a/dbms/src/DataTypes/IDataType.h#L53

I also think a better design would be something like this:

```
UvarintNullable(i *uint64) {
  isNull = i == nil
  encodeBinary(isNull)
  if !isNull { Uvarint(*i) }
}
```

which is somewhat closer to how clickhouse implemented this iirc.

kshvakov:
@dqminh is right, you must write the null map before writing data (see https://github.com/kshvakov/clickhouse/blob/master/lib/data/block.go#L114)

ahutsunshine:
@vavrusa @dqminh @kvap @twalwyn Anyone can update the PR and merge? We're using the feature, but seems it's blocked now for a long time.

jakesyl:
hey, #367 should address most of the comments here/I'm using it in prod now. Anyway, if that gets merged we can probably close this. The starting point for that pr is ceba07f7bb8e60218a2f4110a90f044d657b815c (this pull request)

PR #114, opened 2018-08-16
Support for passing limit number as bind parameter

C-Pro:
In current master this would not work:
`select * from table limit ?`
It just does not see `?` placeholder after `limit` and therefore does not substitute.
I have added support for this and corresponding tests.


kshvakov:
Thank you!

PR #117, opened 2018-09-10
Fixing the problem with alter table query execution

irukeru:
While trying to remove old partitions at clickhouse with alter table, current version gives _"unexpected packet [5] from server"_ error and crashes. With this version, problem is solved.


kshvakov:
Thank you!

PR #119, opened 2018-09-11
Adding a column handler for AggregatedFunction

joemadeus:
Note that this is based on the v1.3.4 tag, since I'm unable to get `master` to compile or run tests. Happy to move it over to `master` if that needs be.

* Added an `AggregateFunction` type (implements `Column`) to a new file, `function.go` in the `lib/column` dir. This adds support for `AggregateFunction` columns in CH, holding the name of the aggregation function in a field named `function` and the type of its rvalue in `base.valueOf`.
* Added a `case` clause to the `Factory` func that resolves `chTypes` to `column` types.
* Added a (very) basic unit test.

### DIFF
--- a/lib/column/column.go
+++ b/lib/column/column.go
@@ -129,6 +129,8 @@ func Factory(name, chType string, timezone *time.Location) (Column, error) {
 	}
 
 	switch {
+	case strings.HasPrefix(chType, "AggregateFunction"):
+		return parseAggregateFunction(name, chType, timezone)
 	case strings.HasPrefix(chType, "Array"):
 		return parseArray(name, chType, timezone)
 	case strings.HasPrefix(chType, "Nullable"):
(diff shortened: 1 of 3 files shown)


joemadeus:
The unit test, for now, does not work and I'm stumped. It errors out with this:

```
=== RUN   Test_Column_AggregateFunction
--- FAIL: Test_Column_AggregateFunction (0.00s)
	column_test.go:469: 
			Error Trace:	column_test.go:469
			Error:      	Not equal: 
			            	expected: 0xb
			            	actual  : 0x16
			Test:       	Test_Column_AggregateFunction
```

...which I simply don't understand. @kshvakov is there a hint you could give me that'd let me fix it?

joemadeus:
Closing this since the same query can be done without placeholders -- i.e., by executing the insert as a string without args in the `Exec()` func.

Issue #124, opened 2018-09-18
How do I write null values?


derekperkins:
I'm trying to insert null values, so I have columns like `*int32` and nested columns like `[]*int16`, but that results in an error like this.
`FieldName (Int32): unexpected type *int32`

Is there a different supported way to write nil values? I do this fairly often with the `mysql` driver.

derekperkins:
Should all the `/lib/column/` Write methods add type assertions for the corresponding pointer values?

https://github.com/kshvakov/clickhouse/blob/aeedb7b9d0584f393905447021238b5c6e9c5154/lib/column/int32.go#L18-L29

derekperkins:
I don't think that test covers typical usage. You are testing inserting a nil constant, where normal usage is going to be `*int32`, `*string`, etc, that may or may not be null.

https://github.com/kshvakov/clickhouse/blob/master/clickhouse_nullable_test.go#L213-L230

derekperkins:
This line isn't ever going to return nil except for a nil literal. https://github.com/kshvakov/clickhouse/blob/master/lib/column/nullable.go#L54

See example: https://play.golang.org/p/QuyrmlpqX03

derekperkins:
I think the two options are to add reflection to that if statement or to add checks on all the type switches. I think I'd prefer the type switching option, but the flow of the program makes it easier to do the first.

dstruck:
> Nullable example https://github.com/kshvakov/clickhouse/blob/master/clickhouse_nullable_test.go#L14
> But I'm not sure that `Nullable(Array(*))`is working correctly, I'll check it.

Nullable support for arrays does not seem to work yet, at least for the DateTime type.

**Default DateTime array:**
```
# clickhouse-client
CREATE TABLE test (ts Array(DateTime)) ENGINE=Memory;
INSERT INTO test VALUES (array(now(), now()));

# go client
_, err = stmt.Exec(clickhouse.ArrayDateTime([]time.Time{time.Now(), time.Now()}))
```
Works.

**Nullable DateTime array:**
```
# clickhouse-client
CREATE TABLE test (ts Array(Nullable(DateTime))) ENGINE=Memory;
INSERT INTO test VALUES (array(now(), NULL));
```
Works.
```
# go client
_, err = stmt.Exec(clickhouse.ArrayDateTime([]time.Time{time.Now(), time.Now()}))
```
Does not work and gives back the following error message:

[...]

PR #125, opened 2018-09-18
add test for nil pointers, not just nil literal

derekperkins:
This should work as a basic test case for #124. If you like this direction, I can flesh it out more.

### DIFF
--- a/clickhouse_nullable_test.go
+++ b/clickhouse_nullable_test.go
@@ -211,9 +211,9 @@ func Test_Nullable(t *testing.T) {
 					if tx, err := connect.Begin(); assert.NoError(t, err) {
 						if stmt, err := tx.Prepare(dml); assert.NoError(t, err) {
 							if _, err := stmt.Exec(
-								nil,
+								new(int8),
 								16,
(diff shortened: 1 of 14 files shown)


derekperkins:
I used reflection to check for nil, but I'm not sure how much overhead that adds. Cargo culting from this article, we could replace that code with `unsafe` for performance reasons.
https://dev.to/pauljlucas/go-tcha-when-nil--nil-hic

```
func isNilValue(i interface{}) bool {
    return (*[2]uintptr)(unsafe.Pointer(&i))[1] == 0
}
```

derekperkins:
@kshvakov Does this feel like it is heading it a good direction? It's not working for me locally, but I also don't want to spend a ton more time on it if you don't think it's going to get anywhere.

derekperkins:
I've tested this out in my setup and for some reason it seems to just lock up, but doesn't emit any error messages. I'd really appreciate any guidance you have on the best way to move forward.

kshvakov:
Hi, it's looking good. Thank you!

PR #129, opened 2018-10-04
swap order of UUID bytes 0-7 and 8-15 when reading and writing to clickhouse

Niev:
Bytes must be sent to clickhouse in reverse order to be writen in natural order


kshvakov:
Hi, I'll check it later because the current master is broken by another PR.

kshvakov:
OK, thank you!

PR #131, opened 2018-10-05
Support arithmetic operators before ? placeholders

nineinchnick:
Support arithmetic operators before ? placeholders by adding extra operators when looking for keywords in `stmt.bind()`


kshvakov:
Thank you!

PR #141, opened 2018-11-29
Date column writing with zone offset fixed

neverlee:
[issues140](https://github.com/kshvakov/clickhouse/issues/140)


kshvakov:
Thank you!

Issue #112, opened 2018-08-09
weird behavior of WriteBuffer


paruru48:
My program is encountering memory leak issue. When profiling program  with `pprof`, I notice that `(*WriteBuffer).addChunk` inuse heap is increasing along with running time of program.
 
` 398MB 32.38% 32.38%      398MB 32.38%  github.com/kshvakov/clickhouse/lib/writebuffer.(*WriteBuffer).addChunk`
...
`614.61MB 41.16% 41.16%   614.61MB 41.16%  github.com/kshvakov/clickhouse/lib/writebuffer.(*WriteBuffer).addChunk`

Generally, the program has multi routine that push data to clickhouse in batch.
I had tried to create a brand new connection on each transaction but issue still remains.

kshvakov:
Make an example to reproduce this problem. We don't have this issue on our production servers.

mnach:
@paruru48
Hi! I have the same issue, but in production only. Can you share some sample of code which triggers this memory leak?

kshvakov:
We have a better memory usage in [V2](https://github.com/ClickHouse/clickhouse-go/tree/v2)

PR #144, opened 2018-12-06
add leaky byte pool for memory management, remove prealloc on dial

c3mb0:
Our production load uses multiple transactions since the nature of the data varies a lot, along with the number of prepared statements. After many OoM machines, I profiled my application and here's the relevant part from a heap dump:

pprof

The `sync.Pool` used for `WriteBuffer` gets out of hand quite fast. I've replaced it with a leaky pool implementation that is much more easy on the memory. This does create more GC pressure, however it's barely noticeable and I can at least use the driver now. The size of this pool can be adjusted via driver string parameters, having a sensible 1000 as default.

The heap generated by `dial` happens when the driver experiences multiple consecutive errors. This is because the connection is closed and redialed in the case of an error. A large volume of errors that happen in a short time uses a lot of memory because of the preallocation made during dial. A misformed (whether it is user or library related) query can quickly fill the memory unnecessarily. This is why I've switched it to a simple `bufio.NewReader`. I think it's preferable to use more CPU (not that I've noticed) instead of jeopardizing the app/machine as a whole.

This code's been in production for 2 days and no problems so far. No more OoM machines, RAM usage dropped to 1/5.

Solves [#112](https://github.com/kshvakov/clickhouse/issues/112).

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ Golang SQL database driver for [Yandex ClickHouse](https://clickhouse.yandex/)
     * random   - choose random server from set 
     * in_order - first live server is choosen in specified order
 * block_size - maximum rows in block (default is 1000000). If the rows are larger then the data will be split into several blocks to send them to the server
+* pool size - maximum amount of preallocated byte chunks used in queries (default is 100). Decrease this if you experience memory problems at the expense of more GC pressure and vice versa.
 * debug - enable debug output (boolean value)
 
 SSL/TLS parameters:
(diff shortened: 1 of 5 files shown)


kshvakov:
Hi, thank you. I'll test these changes.

c3mb0:
Further testing revealed that 1000 is still a bit too much for varied concurrent prepared statements such as:
```
INSERT INTO table (x) VALUES (x);
INSERT INTO table (x, y) VALUES (x, y);
INSERT INTO table (x, y, z) VALUES (x, y, z);
INSERT INTO table (x, z) VALUES (x, z);
...
```
About 100 of them uses around 3GB of memory when the pool size is set to 100. While 1000 should be alright for a production machine, keeping the pool size lower at about 100 would prevent the misunderstanding of this library's memory management and enable it to run without problems on smaller machines (development environment, commodity hardware, etc).

kshvakov:
It's looking good. Let's try it ;)

PR #145, opened 2018-12-07
add lz4 compress support

sundy-li:
This pr add LZ4 Compression support(With golang implemented cityhash102)


kshvakov:
Hi, thank you. It's very important feature. Maybe the [pure Go](https://github.com/pierrec/lz4) implementation of the LZ4 codec will be better? I'll try to make some tests for the comparison these libraries.

kshvakov:
@sundy-li thank you!

PR #151, opened 2019-02-07
Allow to use pointer to variables in Scan

altush1983:
It is very useful to reduce code for example when copy data from one database to Clickhouse DB


altush1983:
See that it is already added. Sorry )

PR #152, opened 2019-02-09
Fix Readme example registering driver

bithavoc:
Just a simple change to clarify driver registration in the Readme example.

Thanks for such great project BTW 🎉

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ import (
 	"log"
 	"time"
 
-	"github.com/kshvakov/clickhouse"
+	_ "github.com/kshvakov/clickhouse" // registers "clickhouse" driver
 )
 
 func main() {


kshvakov:
`clickhouse` package is used in type cast (see clickhouse.Exception)

bithavoc:
Why close this? The Readme is still outdated, there are two examples: sqlx works, the first one doesn't. 

This was a fix to that typo.

I don't understand what you're saying about clickhouse.Exception

kshvakov:
If you export package like `_ "github.com/kshvakov/clickhouse"` you''ll receive an error about this 

```go 
if exception, ok := err.(*clickhouse.Exception); ok { // <- undefined: clickhouse
```

bithavoc:
ok ok I see :v:

PR #157, opened 2019-02-28
Correctly handle zero timestamps

Upliner:
Clickhouse displays zero timestamps as 0000-00-00 00:00:00
However, Time.Unix() returns non-zero for zero timestamps
This patch makes correct handling of golang zero timestamps


kshvakov:
Thank you!

PR #161, opened 2019-03-28
Fix Nullable support for Decimal types

### LINKED ISSUE #160: Nullable Decimal
Hi, am I missing something here https://github.com/kshvakov/clickhouse/blob/master/lib/column/decimal.go#L97 or is it currently not possible to use nullable decimals? The null itself will work (handled elsewhere), but actual values can't be passed because there is no handling for the pointer types.

grongor:
Fixes #160


kshvakov:
Thank you!

PR #163, opened 2019-04-02
Fix dataSize not being updated while data size larger than chunk freeSize

litao91:
While writing data to chunks, if the `dataSize` is larger than the `freeSize` of the current size, the data will be cut, but the `dataSize` won't be updated accordingly. This would cause slices bound out of range in the case when single record is large enough.

### DIFF
--- a/lib/writebuffer/buffer.go
+++ b/lib/writebuffer/buffer.go
@@ -22,14 +22,17 @@ func (wb *WriteBuffer) Write(data []byte) (int, error) {
 	var (
 		chunkIdx = len(wb.chunks) - 1
 		dataSize = len(data)
+		writtenSize = 0
 	)
 	for {
 		freeSize := cap(wb.chunks[chunkIdx]) - len(wb.chunks[chunkIdx])
 		if freeSize >= len(data) {
 			wb.chunks[chunkIdx] = append(wb.chunks[chunkIdx], data...)
-			return dataSize, nil
+			writtenSize += len(data)
+			return writtenSize, nil
 		}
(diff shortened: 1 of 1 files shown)


kshvakov:
Hi, thank you. Please fix the tests https://travis-ci.org/kshvakov/clickhouse/builds/514612302

cw9:
Hi @litao91 , I do not understand why this change is needed. Isn't `writtenSize` always equal to the original `dataSize`? Can you explain what case is the original code breaking? If this diff fixes some cases then there should be a regression unit test added, otherwise people can just revert this change without knowing it's breaking something.

@kshvakov do you think the regression unit test policy make sense?

litao91:
> Hi @litao91 , I do not understand why this change is needed. Isn't `writtenSize` always equal to the original `dataSize`? Can you explain what case is the original code breaking? If this diff fixes some cases then there should be a regression unit test added, otherwise people can just revert this change without knowing it's breaking something.
> 
> @kshvakov do you think the regression unit test policy make sense?

@cw9 My fault, I actually wanted to fix the leaky pool, as you've fixed here: https://github.com/kshvakov/clickhouse/commit/23823208a6b68e14b6f80a5ad1d71410fa8e4535#diff-053b713e663dcc7f46d99264eaa83043

I was using an older version prior to your fix, so I fixed it again and directly apply the patch to the master.

litao91:
@kshvakov Please revert this change, thanks.

kshvakov:
@litao91 Done )

kshvakov:
I have an error after revert https://travis-ci.org/kshvakov/clickhouse/jobs/563565890 . Can you fix it?

PR #179, opened 2019-06-06
Fix compressed data read for Read()s that span multiple compressed payloads

jtroy:
I ran into an issue when using compression which resulted in incorrect data being read when that data spanned two compressed payloads. The root cause of the issue is a missing copy() call in the code which is meant to combine decompressed data into a buffer. The behavior of the code should be:

```
copy data from the previously decompressed buffer
while the length of the read isn't satisfied {
    decompress the next payload into a buffer
    copy the decompressed data from the buffer
}
```

This PR adds the missing copy(), fixes an error message in the file, and formats that error message to be more easily read.


kshvakov:
Thank you!

PR #187, opened 2019-07-04
Set timeout properly for TLS connection 

dysosmus:
The 'timeout' option isn't properly set for TLS connection, it's currently hard-coded to 5 seconds.
This PR set the timeout of the connection to the `timeout` option value like for insecure connections.


kshvakov:
Thank you!

Issue #188, opened 2019-07-05
Add support for inserting IPv4 and IPv6 types


ayng:
Since [Clickhouse release 19.3.3 on 2019-02-13](https://clickhouse.yandex/docs/en/changelog/#clickhouse-release-19-3-3-2019-02-13), dedicated IPv4 and IPv6 types have been added.

When inserting to a table with columns of type IPv4, this library returns the error `column: unhandled type IPv4`. I can see that IPv4 and IPv6 are not listed in the switch-case in [lib/column/column.go](https://github.com/kshvakov/clickhouse/blob/master/lib/column/column.go#L21)

My usage of this library could benefit from inserting IPv4 and IPv6 types, since I have tables that contain IPv4 and IPv6 addresses. Are you interested in supporting insertion of IPv4 and IPv6 types? If yes, I'm happy to work with you to come up with a pull request.

PR #189, opened 2019-07-09
Support IPv4 and IPv6 data types

ayng:
For insertion, handling of net.IP is taken over by new column types IPv4
and IPv6. Values of type net.IP can no longer be inserted into columns
of type FixedString(16).

For selection, columns of type IPv4 and IPv6 are returned as values with
type net.IP.

Addresses issue #188

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,8 @@ tcp://host1:9000?username=user&password=qwerty&database=clicks&read_timeout=10&w
 * FixedString(N)
 * Date 
 * DateTime
+* IPv4
+* IPv6
 * Enum
 * UUID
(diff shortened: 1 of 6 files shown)


kshvakov:
Thank you!

ayng:
Sorry, I just realized that I got the endianness wrong for both read and write for IPv4. The test looks reasonable but it only works because insertion reverses the value and selection reverses the value again, resulting in the correct value, but the value in Clickhouse is backwards.

I'm working on a fix.

ayng:
Just created a PR with the fix (above). Again, sorry for introducing this bug into master 😓

PR #193, opened 2019-07-10
Fix cancel not closing if write error

fastest963:
If there was an error writing the cancel message then `cancel` wouldn't close the connection. Additionally, if the write did succeed, it was never flushed, so in my debugging tcpdump's it was never actually sent to the server.

I made `cancel` return either the error from `Close` or the error from the write/flush. Though this ever is never actually used in `watchCancel`.


kshvakov:
Thank you!

Issue #202, opened 2019-08-07
Alter table Drop Partition hangs reading from Clickhouse


abraithwaite:
Not sure what's going on but it looks like it's stuck reading from the network.

Using version: 1d87030de2a38bbe8c05e1b13f5b2f27466fd5a8

Connected to ClickHouse server version 19.9.2 revision 54421.

This looks related: https://github.com/kshvakov/clickhouse/commit/4ca5dc1073e63661819d546d662ac8c19848e42d

```
^\SIGQUIT: quit
PC=0x45b103 m=0 sigcode=128

goroutine 9 [syscall]:
runtime.notetsleepg(0xb0eee0, 0xdf8470cfe, 0x1414)
        /usr/lib/go/src/runtime/lock_futex.go:227 +0x34 fp=0xc000039f60 sp=0xc000039f30 pc=0x40c014
runtime.timerproc(0xb0eec0)
        /usr/lib/go/src/runtime/time.go:311 +0x2ea fp=0xc000039fd8 sp=0xc000039f60 pc=0x44a9aa
runtime.goexit()
        /usr/lib/go/src/runtime/asm_amd64.s:1337 +0x1 fp=0xc000039fe0 sp=0xc000039fd8 pc=0x4592a1
created by runtime.(*timersBucket).addtimerLocked
        /usr/lib/go/src/runtime/time.go:169 +0x10e

goroutine 1 [IO wait]:
internal/poll.runtime_pollWait(0x7fbee951b310, 0x72, 0xffffffffffffffff)
        /usr/lib/go/src/runtime/netpoll.go:182 +0x56
internal/poll.(*pollDesc).wait(0xc000144618, 0x72, 0x1000, 0x1000, 0xffffffffffffffff)
        /usr/lib/go/src/internal/poll/fd_poll_runtime.go:87 +0x9b
internal/poll.(*pollDesc).waitRead(...)
        /usr/lib/go/src/internal/poll/fd_poll_runtime.go:92
internal/poll.(*FD).Read(0xc000144600, 0xc000170000, 0x1000, 0x1000, 0x0, 0x0, 0x0)
        /usr/lib/go/src/internal/poll/fd_unix.go:169 +0x19b
net.(*netFD).Read(0xc000144600, 0xc000170000, 0x1000, 0x1000, 0x528dbf, 0xc000144600, 0xc000170000)
        /usr/lib/go/src/net/fd_unix.go:202 +0x4f
[...]

PR #203, opened 2019-08-08
Fix alter table #202

kshvakov:


kshvakov:
https://github.com/kshvakov/clickhouse/issues/202#issuecomment-519314023

PR #207, opened 2019-09-12
Support inserting [][]byte to Array(String) column

cw9:
Previously the driver only supports writing []string to Array(String) column.

This diff enables writing [][]byte to Array(String) column


cw9:
Please advice if there is a more efficient way to do this.

kshvakov:
Thank you!

PR #212, opened 2019-09-19
Add WriteArrayWithValue function to block

cw9:
This diff adds a `Value` interface and a `WriteArrayWithValue` function to block package. 

This allows user to provide custom implementation of `Value` and avoid unnecessary reflection cost.


cw9:
@kshvakov Can you take a look at this? The performance in our benchmark increased 30% by switching `WriteArray` to `WriteArrayWithValue `

kshvakov:
Hi! Thank you. It's simple and real fast!

PR #219, opened 2019-10-09
Add a better example for WriteArrayWithValue

cw9:
By using `WriteArrayWithValue` one could reduce the cost in reflection, but a simple `data.Value` implementation can still cause `runtime.convT2I` cost, this diff adds an example of how to avoid `runtime.convT2I` cost.

For most use cases the simple implementation should be good enough.

### DIFF
--- a/clickhouse_columnar_test.go
+++ b/clickhouse_columnar_test.go
@@ -28,8 +28,9 @@ func Test_ColumnarInsert(t *testing.T) {
 				enum8      Enum8 ('a' = 1, 'b' = 2),
 				enum16     Enum16('c' = 1, 'd' = 2),
 				array      Array(String),
-				array2     Array(UInt64),
-				arrayArray Array(Array(String))
+				arrayArray Array(Array(String)),
+				arrayWithValue Array(UInt64),
+				arrayWithValueFast Array(UInt64)
 			) Engine=Memory
 		`
 		dml = `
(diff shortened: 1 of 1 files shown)


clickhouse_columnar_test.go
```diff
@@ -108,8 +110,9 @@ func Test_ColumnarInsert(t *testing.T) {
 						block.WriteUInt8(10, 1)
 						block.WriteUInt16(11, 2)
 						block.WriteArray(12, []string{"A", "B", "C"})
-						block.WriteArrayWithValue(13, uint64SliceValue{value: []uint64{1, 2, 3}})
-						block.WriteArray(14, [][]string{[]string{"A", "B"}, []string{"CC", "DD", "EE"}})
+						block.WriteArray(13, [][]string{[]string{"A", "B"}, []string{"CC", "DD", "EE"}})
```
golangcibot: Error return value of `block.WriteArray` is not checked (from `errcheck`)

clickhouse_columnar_test.go
```diff
@@ -108,8 +110,9 @@ func Test_ColumnarInsert(t *testing.T) {
...
+						block.WriteArrayWithValue(14, newUint64SliceValueSimple([]uint64{1, 2, 3}))
```
golangcibot: Error return value of `block.WriteArrayWithValue` is not checked (from `errcheck`)

clickhouse_columnar_test.go
```diff
@@ -108,8 +110,9 @@ func Test_ColumnarInsert(t *testing.T) {
...
+						block.WriteArrayWithValue(15, newUint64SliceValueFast([]uint64{10, 20, 30}))
```
golangcibot: Error return value of `block.WriteArrayWithValue` is not checked (from `errcheck`)

PR #225, opened 2019-10-31
add more uint64 case for uint column

sundy-li:
we use code like this to parse the string to int:

```
	case "UInt8", "UInt16", "UInt32", "UInt64":
		ui, err := strconv.ParseUint(ifnull, 10, 64)
		if err != nil {
			return 0
		}
		return ui
```

it will panic when using UInt64 for UInt32 column, this's a simple fix for that case.

I'd like to switch all int case `int,int8,int16,int32,int64,  uint8,uint16,uint32,uint64`  in   `column/int*.go`  and `column/uint*.go`  . But got too much dirty paste,   golang really need template~~. Or we could do it by code gen.

### DIFF
--- a/lib/column/uint16.go
+++ b/lib/column/uint16.go
@@ -20,6 +20,8 @@ func (u *UInt16) Write(encoder *binary.Encoder, v interface{}) error {
 		return encoder.UInt16(v)
 	case int64:
 		return encoder.UInt16(uint16(v))
+	case uint64:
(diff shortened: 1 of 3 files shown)


kshvakov:
Thank you! 
> golang really need template 
Yes!) 

I was thinking about сodegen, and that is a really good thought. We really need a template for ClickHouse types

PR #228, opened 2019-11-12
Better write array with benchmark

vahid-sohrabloo:
For better performance, I add a write length function. For check it I also provide a benchmark.
this is a result of my benchmark
```
Benchmark_BlockWriteArrayWithLen4-4   	32752696	        45.5 ns/op	      49 B/op	       0 allocs/op
Benchmark_BlockWriteArray4-4          	 4509878	       264 ns/op	     113 B/op	       4 allocs/op
```

### DIFF
--- a/lib/data/block.go
+++ b/lib/data/block.go
@@ -95,13 +95,13 @@ func (block *Block) Read(serverInfo *ServerInfo, decoder *binary.Decoder) (err e
 	return nil
 }
 
-func (block *Block) writeLen(l, num, level int) {
+func (block *Block) writeLen(length, num, level int) {
 	if len(block.offsets[num]) < level {
-		block.offsets[num] = append(block.offsets[num], []int{l})
+		block.offsets[num] = append(block.offsets[num], []int{length})
 	} else {
 		block.offsets[num][level-1] = append(
 			block.offsets[num][level-1],
-			block.offsets[num][level-1][len(block.offsets[num][level-1])-1]+l,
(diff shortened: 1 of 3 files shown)


lib/data/block.go
```diff
@@ -95,20 +95,24 @@ func (block *Block) Read(serverInfo *ServerInfo, decoder *binary.Decoder) (err e
 	return nil
 }
 
+func (block *Block) writeLen(l, num, level int) {
```
cw9: use a more intuitive name instead of `l`?

lib/data/block_benchmark_test.go
```diff
@@ -0,0 +1,46 @@
...
+	block.Columns = append(block.Columns, c)
+	block.Reserve()
+	b.ReportAllocs()
+
+	for i := 0; i < b.N; i++ {
+		block.WriteArrayLen(4, 0, 1)
+		for i := 0; i > 4; i++ {
+			block.WriteInt32(0, 1)
+		}
+	}
```
cw9: This only works for one level arrays? Does it work for Nested Arrays as well?

lib/data/block.go
vahid-sohrabloo: I changed it

lib/data/block_benchmark_test.go
vahid-sohrabloo: I add another benchmark

PR #230, opened 2019-11-12
Fix Block Reset and remove not needed code

vahid-sohrabloo:
As I know in Golang we don't need to reset data and then nil child of an object and then nil object and ...
Only nil data is enough and GC removes data from memory.

### DIFF
--- a/clickhouse.go
+++ b/clickhouse.go
@@ -129,10 +129,7 @@ func (ch *clickhouse) beginTx(ctx context.Context, opts txOptions) (*clickhouse,
 func (ch *clickhouse) Commit() error {
 	ch.logf("[commit] tx=%t, data=%t", ch.inTransaction, ch.block != nil)
 	defer func() {
-		if ch.block != nil {
-			ch.block.Reset()
-			ch.block = nil
-		}
+		ch.block = nil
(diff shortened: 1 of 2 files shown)


develar:
Sorry, but have your read source code of deleted method? `Block` under the hood uses pool, and `Reset` returns data to pool for further usage. Please see https://github.com/ClickHouse/clickhouse-go/blob/6b1ba590788873fd6993f50f823bb90274a23ba5/lib/writebuffer/buffer.go#L89

`Block.Reset` -> `buffer.reset`-> `WriteBuffer.Reset` (both — `offsetBuffer` and `columnBuffer `).

vahid-sohrabloo:
> Sorry, but have your read source code of deleted method? `Block` under the hood uses pool, and `Reset` returns data to pool for further usage. Please see
> 
> https://github.com/ClickHouse/clickhouse-go/blob/6b1ba590788873fd6993f50f823bb90274a23ba5/lib/writebuffer/buffer.go#L89
> 
> `Block.Reset` -> `buffer.reset`-> `WriteBuffer.Reset` (both — `offsetBuffer` and `columnBuffer `).

Ops Your right. sorry

PR #239, opened 2019-12-16
add more detail docs on block_size

sundy-li:
according  to [this](https://github.com/ClickHouse/ClickHouse/blob/604a31bb66953f397eab3bd006e3bb68c8889543/dbms/programs/server/TCPHandler.cpp#L982)


kshvakov:
Thank you!

PR #240, opened 2019-12-19
do not allocate compress buffer when compress not used to lowdown the memory

sundy-li:


sundy-li:
I can't remove the OffsetWriter..

PR #253, opened 2020-02-11
implement RowsColumnTypeNullable and RowsColumnTypePrecisionScale

sobolevsv:
Implement RowsColumnTypeNullable and RowsColumnTypePrecisionScale interfaces by rows struct to get decimal's precision and scale via func (*Rows) ColumnTypes

https://golang.org/pkg/database/sql/#Rows.ColumnTypes

interfaces are defined here https://github.com/golang/go/blob/master/src/database/sql/driver/driver.go

### DIFF
--- a/clickhouse_decimal_test.go
+++ b/clickhouse_decimal_test.go
@@ -1,0 +1,102 @@
+package clickhouse_test
+
+import (
+	"database/sql"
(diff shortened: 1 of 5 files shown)


kshvakov:
Thank you

PR #255, opened 2020-02-13
Add connOpenTimeRandom strategy to distribute connections equally across shards. 

rebrendov:
I have 10 instances of application that is writing to a sharded cluster of clickhouse. There are more than 20 hosts and i want to distribute connections equally across all of them. In dsn i pass them as alt_hosts.

Unfortunately I couldn't archive desirable outcome with currently available strategies for opening connections. The `random` strategy, which is used by default, chooses host depending on [the number of connections opened before](https://github.com/ClickHouse/clickhouse-go/blob/master/connect.go#L64). So the connections are opened in the same order for each app instance and that leads to an uneven spread of connections across hosts.

I added new strategy for choosing a host to connect. Using this strategy I was able to archive somewhat even spread of connections across shards. 

It's not perfect but it does the job. If you have any thoughts on this, or know a better way to do that - let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,8 +18,9 @@ Golang SQL database driver for [Yandex ClickHouse](https://clickhouse.yandex/)
 * no_delay   - disable/enable the Nagle Algorithm for tcp socket (default is 'true' - disable)
 * alt_hosts  - comma separated list of single address host for load-balancing
 * connection_open_strategy - random/in_order (default random).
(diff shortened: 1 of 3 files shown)


kshvakov:
Hi, it's looking good! Please add info about the new strategy to the README

rebrendov:
Done! Hope my description is good.

PR #266, opened 2020-04-07
date and datetime can be unsigned

pikot:
subj, if i pack data/datetime  as unsigned int, error will happen,


kshvakov:
thank you

PR #271, opened 2020-04-28
Add DateTime64 support

troyanov:
PR to add support for [DateTime64](https://clickhouse.tech/docs/en/sql_reference/data_types/datetime64/) data type.

### DIFF
--- a/clickhouse_test.go
+++ b/clickhouse_test.go
@@ -90,6 +90,7 @@ func Test_Insert(t *testing.T) {
 				fString FixedString(2),
 				date    Date,
 				datetime DateTime,
+				datetime64 DateTime64,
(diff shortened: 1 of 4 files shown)


troyanov:
Build failure caused by https://github.com/ClickHouse/clickhouse-go/pull/270

mdelvecchio-wasabi:
I am trying to convert a DateTime column to DateTime64.  The DB seems fine, with nonsecond-precision values in the column.

But I can't get any queries to work; the first query on this table always produces zero rows, and the second produces "[process] unexpected packet [11] from server".

I am hoping that this PR will fix these problems.  Is this what this PR is intended to do?

Thanks.

Issue #224, opened 2019-10-22
Support selecting Array(Array(T)) column


cw9:
Hi,

Is `Array(Array(T))` column supported in select queries? 

I saw [this](https://github.com/kshvakov/clickhouse/blob/4ca0da588f7f562145a0bca218f4d89d174d5ac9/clickhouse_test.go#L1247) test has been commented out and it seems to be failing when I uncomment it.

Thanks,
Chao

kshvakov:
Yes, see [V2](https://github.com/ClickHouse/clickhouse-go/blob/v2/tests/array_test.go#L75)

PR #282, opened 2020-06-14
Add ability to read multidimensional arrays

afterdusk:
Resolves issue #224 . `TestArrayArrayT` was updated to cover this change. Please let me know if there needs to be more tests, or if any specific case should be covered.

CI also seems to be [failing](https://travis-ci.org/github/ClickHouse/clickhouse-go/jobs/693388356#L253) on `Test_CompressCorrect`, although I do not have the expertise to help with that.


kshvakov:
Good job. Thank you!

Issue #285, opened 2020-06-24
invalid Enum value: 0


veqryn:
Our organization is starting to use golang to read from our clickhouse instances, and immediately encountered this error/bug:
`invalid Enum value: 0`

We have a table that looks roughly like:
```sql
CREATE TABLE IF NOT EXISTS analytics.stuff_local (
    foo     Nullable(Enum8('bar' = 1, 'qux' = 2, 'corge' = 3))
    -- and lots of other fields...
```

Queries we run look roughly like:
```sql
SELECT foo, count(*) FROM stuff_local GROUP BY foo
```

We can run these queries from command line or through other drivers, giving us good results:
```
NULL 12313
bar 38532
qux 2329
corge 14003
```

But with this golang driver, we get the error: `invalid Enum value: 0`

When I look at what is happening in debug mode, I can see that data being read is NULL and has a value of zero.
Since zero is not an allowed enum value, we get the above error.

PR #286, opened 2020-06-24
Fixes: invalid Enum value: 0 for valid null records

veqryn:
Fixes https://github.com/ClickHouse/clickhouse-go/issues/285

Test is included.

---
This change is [Reviewable](https://reviewable.io/reviews/clickhouse/clickhouse-go/286)

### DIFF
--- a/lib/column/enum.go
+++ b/lib/column/enum.go
@@ -15,7 +15,7 @@ type Enum struct {
 	baseType interface{}
 }
 
-func (enum *Enum) Read(decoder *binary.Decoder) (interface{}, error) {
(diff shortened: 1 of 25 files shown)


lib/column/enum.go
```diff
@@ -30,7 +30,7 @@ func (enum *Enum) Read(decoder *binary.Decoder) (interface{}, error) {
 			return nil, err
 		}
 	}
-	if ident, found := enum.vi[ident]; found {
+	if ident, found := enum.vi[ident]; found || isNull {
```
veqryn, line 33: This is the fix

kshvakov:
Thank you for your contribution!

PR #293, opened 2020-07-15
Allow int64 switch to float

caiwc:
I think float type should allow more type conversion.

### DIFF
--- a/lib/column/float32.go
+++ b/lib/column/float32.go
@@ -16,12 +16,16 @@ func (Float32) Read(decoder *binary.Decoder, isNull bool) (interface{}, error) {
 
 func (float *Float32) Write(encoder *binary.Encoder, v interface{}) error {
 	switch v := v.(type) {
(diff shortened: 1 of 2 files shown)


kshvakov:
Hi, this is not a good idea

caiwc:
We are implementing a function that insert data into a table of unknown field type. Since I just can use `proto.Struct` as the type of inserted data and `proto.Value` just support `float64` for number , so the number type data I got is all `float64`. 
my solution is  convert the float to `int64` if it is a whole number. But when I insert a the whole number to Float type column will be fail because of the limitation of type conversion. 
The ideal solution for us would be to float and int can be converted to each other and rasie errors if overflows

kshvakov:
Use ClickHouse metadata (like `DESC table_name`) and check/convert types before inserting. That logic must be on application side, IMHO

caiwc:
os,I see, thank you !

Issue #329, opened 2020-11-24
Queries that filter on time don't work for the date 2059-09-19 and after


veqryn:
We have a table that looks roughly like:
```sql
CREATE TABLE IF NOT EXISTS analytics.stuff_local (
    eventday           Date DEFAULT toDate(created),
    created            DateTime('UTC'),
    user_id            UInt32,
    -- and lots of other fields...
    sign               Int8,
    version            UInt16
) ENGINE = VersionedCollapsingMergeTree(sign, version)
      PARTITION BY toYYYYMM(eventday)
      ORDER BY (user_id, eventday, created)
      TTL eventday + toIntervalDay(2) DELETE;
```

Queries we run look roughly like:
```sql
SELECT COUNT(*)
FROM analytics.stuff_local
WHERE eventday >= '2000-01-01'
  AND eventday <= '2100-01-01'
  AND created >= '2000-01-01 00:00:00'
  AND created < '2100-01-01 00:00:00'
```

We can run these queries from command line or through other drivers, giving us good results:
```
2318
```

But with this golang driver, we get zero (as in, zero rows match the 'where' filter).

If I change the query to be the following, it works with this driver:
```sql
SELECT COUNT(*)
FROM analytics.stuff_local
WHERE eventday >= '2000-01-01'
  AND eventday <= '2037-01-01'
  AND created >= '2000-01-01 00:00:00'
  AND created < '2037-01-01 00:00:00'
```

This is a small reproducer of our golang:
```golang
func main() {
	conn, err := newClickhouse(context.TODO(), env.SeedsClickhouseUser, env.SeedsClickhousePass, env.SeedsClickhouseDatabase, env.SeedsClickhouseHost, env.SeedsClickhousePort)
	if err != nil {
		panic(err)
	}

	err = conn.Ping()
	if err != nil {
		panic(err)
	}

	sqlStr := `
		SELECT COUNT(*)
		FROM inbox.campaign
		WHERE eventday >= ?
		AND eventday <= ?
		AND created >= ?
		AND created < ?`

	start := time.Date(2000, 1, 1, 0, 0, 0, 0, time.UTC)

	//end := time.Date(2030, 1, 1, 0, 0, 0, 0, time.UTC) // Works (large count)
	//end := time.Date(2059, 9, 18, 0, 0, 0, 0, time.UTC) // Works (large count)
	end := time.Date(2059, 9, 19, 0, 0, 0, 0, time.UTC) // Does not work (zero count)
[...]

PR #330, opened 2020-11-24
Fix int16 overflow preventing correct queries with time.Time

veqryn:
Fixes https://github.com/ClickHouse/clickhouse-go/issues/329

At first I mostly fixed the problem by changing the casting from int16 to uint16, but then I realized the better solution didn't need any casting at all, as Clickhouse accepts very large values in toDateTime.


kshvakov:
Thank you

PR #332, opened 2020-12-04
Add uint support for uint columns

jinzhu:


kshvakov:
I think we need to check for integer overflow

PR #335, opened 2020-12-09
fix travis test error

zxc111:
1. fix travis test error
in travis.yml, `go get github.com/pierrec/lz4` not lock version
2. fix: lib/column/column_test.go
DateTime64 test error.


kshvakov:
Thank you!

PR #338, opened 2020-12-23
Add query_id in query context

Zhile:


kshvakov:
Hi. Please resolve conflicts

Zhile:
@kshvakov conflicts are resolved.

kshvakov:
Thank you!

Zhile:
@kshvakov Sorry for the inconvenience.

clickhouse_send_query.go
```diff
@@ -1,22 +1,30 @@
...
 
-func (ch *clickhouse) sendQuery(query string, externalTables []ExternalTable) error {
+func (ch *clickhouse) sendQuery(ctx context.Context, query string, externalTables []ExternalTable) error {
 	ch.logf("[send query] %s", query)
 	if err := ch.encoder.Uvarint(protocol.ClientQuery); err != nil {
 		return err
 	}
-	if err := ch.encoder.String(""); err != nil {
+	var queryId string
+	queryIdValue := ctx.Value("query_id")
```
olomix, line 15: It is not recommended to use context keys as plain strings. It may cause collisions with other packages.
Better to do something like this
```golang
type key uint8
const keyQueryID key

// wrap query_id into context
func WithQueryID(ctx context.Context, queryID string) context.Context {
    return context.WithValue(ctx, keyQueryID, queryID)
}

func (ch *clickhouse) sendQuery(...) {
    ...
    queryIdValue := ctx.Value(keyQueryID)
    ...
}
```
Mode details in golang blog https://blog.golang.org/context#TOC_3.2.

Zhile: This is a good point. I change it according to your suggestion. 
@kshvakov I'm not sure if it's ok to put the WithQueryID in stmt.go. Please help to find the best place or if you have a better idea about this.

sundy-li (approved):
LGTM

sundy-li:
This pr looks great, why not merge this @kshvakov  ?

Issue #336, opened 2020-12-16
driver: bad connection


necryin:
Hi. I found inconsistent with database/sql in common case:
```
...
tx, _ := conn.Begin()
stmt, _ := tx.Prepare("Some insert query") <-- ErrBadConn
...
```
In other drivers (where connection checked in Begin()) I will get this error in Begin() and it will be retried (this logic is in database/sql). But in clickhouse driver (where as i see no real Begin() queries) I got ErrBadConn in Prepare() where no retry logic in database/sql.

Astone-Chou:
same issue

Astone-Chou:
`db.SetConnMaxLifetime(15*time.Minute)` seems working well.

yzx-fish:
same issue.

sashayakovtseva:
same issue. here are debug logs from CH server (names in sql queries are cut out):

```
[clickhouse][connect=5][hello] -> Golang SQLDriver 1.1.54213
[clickhouse][connect=5][hello] <- ClickHouse 19.16.54427 (Etc/UTC)

// the rest of app logs

[clickhouse][connect=5][begin] tx=false, data=false
[clickhouse][connect=5][cancel] <- finished
[clickhouse][connect=5][prepare]
		INSERT INTO ... VALUES (? ...)

[clickhouse][connect=5][send query]
		INSERT INTO
			...
	 VALUES
[clickhouse][connect=5][read meta] <- data: packet=1, columns=6, rows=0
[clickhouse][connect=5][commit] tx=true, data=true
[clickhouse][connect=5][process] <- end of stream
[clickhouse][connect=5][stmt] close

// the rest of app logs

[clickhouse][connect=5][begin] tx=false, data=false
[clickhouse][connect=5][cancel] <- finished
[clickhouse][connect=5][prepare]
		INSERT INTO ... VALUES (? ...)

[clickhouse][connect=5][send query]
		INSERT INTO
			...
	 VALUES
[clickhouse][connect=5][connect] read error: EOF
[clickhouse][connect=5][rollback] tx=true, data=false
// at this point I have "driver: bad connection" during stmt.Prepare() in app logs
```

scarbo87:
@sashayakovtseva did you find a solution with HAProxy? I have the same problem.

sashayakovtseva:
> @sashayakovtseva did you find a solution with HAProxy? I have the same problem.

nope, I switched to direct connection.

PR #350, opened 2021-01-31
fix the "driver: bad connection"  when calling tx.Prepare()

ahmczsy:
#336


kshvakov:
Hi, thank you for your PR. We have implemented `SessionResetter` for `database/sql`

Issue #358, opened 2021-03-11
sending nil to array column causes panic


romberli:
when inserting data, sending clickhouse.Array(nil) to nullable array column causes panic

how to repeat:
1. create table
```
create table if not exists t01
(
	id               Int64,
	groups            Array(LowCardinality(String))
)
	engine = MergeTree PARTITION BY id
	ORDER BY id SETTINGS index_granularity = 8192;
```
2. insert data
```
sql = `insert into t01(id, groups) values(?, ?)`
... ...
_, err = stmt.Exec(1, nil)
```
3. client panic
the client will panic when calling Exec() function, although nullable array is not supported by clickhouse, maybe it's better to return error then panic the whole program.

romberli:
I've created a PR to fix this. #359

PR #359, opened 2021-03-11
fix crash bug when sending nil to array column

romberli:
fix: when inserting data, sending nil to array column, it will panic due to calling value.Interface() in block.AppendRow() function, this commit could avoid this


romberli:
#358

romberli:
as it's been a long time that no one take care of pr, I'll close this pr.

PR #371, opened 2021-04-20
Add a column handler for AggregatedFunction

m9rco:
Since I'm unable to get master to compile or run tests. Happy to move it over to master if that needs be.

Added an AggregateFunction type (implements Column) to a new file, function.go in the lib/column dir. This adds support for AggregateFunction columns in CH, holding the name of the aggregation function in a field named function and the type of its rvalue in base.valueOf.
Added a case clause to the Factory func that resolves chTypes to column types.
Added a (very) basic unit test.

### DIFF
--- a/lib/column/aggregate_function.go
+++ b/lib/column/aggregate_function.go
@@ -1,0 +1,47 @@
+package column
+
+import (
+	"fmt"
+	"github.com/ClickHouse/clickhouse-go/lib/binary"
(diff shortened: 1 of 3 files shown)


m9rco:
retry rebuild

roshaans:
I feel like `AggregateFunction` are a core functionality of Clickhouse and should be supported

PR #372, opened 2021-04-26
Add RegisterDial function

c3mb0:
Add a function to enable the user to optionally register a custom dial function. Implementation mostly inspired by [mysql driver](https://github.com/go-sql-driver/mysql/blob/46351a8892976898935f653f5333782579a96fa5/driver.go#L49).

### DIFF
--- a/clickhouse_test.go
+++ b/clickhouse_test.go
@@ -1555,3 +1555,13 @@ func Test_LikeQuery(t *testing.T) {
 		}
 	}
 }
+
+func Test_RegisterDial(t *testing.T) {
+	clickhouse.RegisterDial(func(network, address string, timeout time.Duration, config *tls.Config) (net.Conn, error) {
+		return net.DialTimeout(network, address, timeout)
+	})
+	if connect, err := sql.Open("clickhouse", "tcp://127.0.0.1:9000?debug=true"); assert.NoError(t, err) {
+		assert.NoError(t, connect.Ping())
+	}
+	clickhouse.DeregisterDial()
+}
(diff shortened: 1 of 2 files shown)


kshvakov:
And what reasons can there be for register a custom Dial function?

c3mb0:
We use it in production to cache DNS resolutions via [dnscache](https://github.com/rs/dnscache). Here's a sample implementation:

```
r := &dnscache.Resolver{Timeout: 10 * time.Second}
clickhouse.RegisterDial(func(network, address string, timeout time.Duration, config *tls.Config) (conn net.Conn, err error) {
	host, port, err := net.SplitHostPort(address)
	if err != nil {
		return nil, err
	}
	ips, err := r.LookupHost(context.TODO(), host)
	if err != nil {
		return nil, err
	}
	for _, ip := range ips {
		if config != nil {
			conn, err = tls.DialWithDialer(&net.Dialer{Timeout: timeout}, network, net.JoinHostPort(ip, port), config)
		} else {
			conn, err = net.DialTimeout(network, net.JoinHostPort(ip, port), timeout)
		}
		if err == nil {
			break
		}
	}
	return
})
```
It can also be used to register custom dialers or troubleshoot connection issues. The implementation is non-intrusive and useful for advanced use cases.

kshvakov:
Can you rebase PR to V1?  I'll add this feature to V2

kshvakov:
Added in V2 and [V1.5.4](https://github.com/ClickHouse/clickhouse-go/releases/tag/v1.5.4)

PR #373, opened 2021-04-28
Fix decimal encoding to allow writing Nullable Decimal32

veqryn:
This is a major bug we've run into: Clickhouse does not allow writing a `sql.NullInt32` to a `Nullable(Decimal32(x))` field.

The error you get is: `narrowing type conversion from int64 to int32`

Why? Because `sql.NullInt32.Value()` returns an `int64`, like so:
```golang
func (n NullInt32) Value() (driver.Value, error) {
	if !n.Valid {
		return nil, nil
	}
	return int64(n.Int32), nil
}
```

And clickhouse's golang driver won't let you write an int64 to a Decimal32 field:
```golang

func (d *Decimal) write32(encoder *binary.Encoder, v interface{}) error {
	switch v := v.(type) {
	// lots of other cases...
	case int64:
		return errors.New("narrowing type conversion from int64 to int32")
	}
}
```

The answer is to allow writing it, so long as it doesn't actually overflow.
This is what my PR does.

---
This change is [Reviewable](https://reviewable.io/reviews/clickhouse/clickhouse-go/373)

### DIFF
--- a/lib/column/decimal.go
+++ b/lib/column/decimal.go
@@ -3,6 +3,7 @@ package column
 import (
 	"errors"
 	"fmt"
+	"math"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


kshvakov:
Thank you!

Issue #382, opened 2021-06-24
The func formatTime() will lost timezone info when it is not zero


EthanZ0111:
I'm using clickhouse-go at version of 1.4.5, and find following code:

```
func formatTime(value time.Time) string {
	// toDate() overflows after 65535 days, but toDateTime() only overflows when time.Time overflows (after 9223372036854775807 seconds)
	if days := value.Unix() / 24 / 3600; days <= math.MaxUint16 && (value.Hour()+value.Minute()+value.Second()+value.Nanosecond()) == 0 {
		return fmt.Sprintf("toDate(%d)", days)
	}
	return fmt.Sprintf("toDateTime(%d)", value.Unix())
}
```
So if we do following query,
```
tz, err := time.LoadLocation("Asia/Shanghai")
assert.NoError(t, err)
create_time, err := time.ParseInLocation("2006-01-02 15:04:05", "2021-06-24 00:00:00", tz)
assert.NoError(t, err)
rows, err := connect.Query("SELECT t FROM example where create_time>?",create_time)
```
then clickhouse-go will use toDate(days) instead of real time with timezone, so the result will not correct.

kshvakov:
Fixed in [V2](https://github.com/ClickHouse/clickhouse-go/tree/v2)

PR #383, opened 2021-06-25
Avoid the influence of timezone when parse time query conditions

EthanZ0111:
Try to fix the issues [382](https://github.com/ClickHouse/clickhouse-go/issues/382).
When query condition is midnight with type of time.Time and non-zero timezone, the real sql exec in clickhouse is not expected.
For example:
```
action_time, err := time.ParseInLocation("2006-01-02 15:04:05", "2021-06-25 00:00:00", tz)
connect.Query("SELECT country_code, os_id, browser_id, categories, action_day, action_time FROM example where action_day >?",action_time)
```
then the sql's condition will looks like `action_day > toDate(18802)` with wrong days(SELECT toDate(18802) will be "2021-06-24 ").

### DIFF
--- a/helpers.go
+++ b/helpers.go
@@ -134,8 +134,11 @@ func quote(v driver.Value) string {
 }
 
 func formatTime(value time.Time) string {
+	// Avoid the influence of timezone when parse midnight, because value.Hour() will be
+	// the hour under special timezone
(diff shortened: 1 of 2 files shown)


kshvakov:
Fixed in [v1.5.2 (format time)](https://github.com/ClickHouse/clickhouse-go/releases/tag/v1.5.2)

PR #384, opened 2021-06-29
Don't overflow Unix seconds by using signed integers.

miguelff:
On 19 January, 2038, Unix timestamp will overflow int32. Thus, reading `DateTime` values that are later than that breaks.

I stumbled across this  problem while reading a stored date of `2040-01-01 10:00:00 +0000 UTC` as  `1903-11-26 03:31:44 +0000 UTC`.

This PR parses DateTime using UInt32 decoders, thus supporting native DateTime32 ranges.

### DIFF
--- a/lib/column/column_test.go
+++ b/lib/column/column_test.go
@@ -438,20 +438,20 @@ func Test_Column_Date(t *testing.T) {
 
 func Test_Column_DateTime(t *testing.T) {
 	var (
-		buf     bytes.Buffer
-		timeNow = time.Now().Truncate(time.Second)
-		encoder = binary.NewEncoder(&buf)
-		decoder = binary.NewDecoder(&buf)
+		buf           bytes.Buffer
+		inTwentyYears = time.Now().Add(20 * 365 * 24 * time.Hour).Truncate(time.Second)
+		encoder       = binary.NewEncoder(&buf)
+		decoder       = binary.NewDecoder(&buf)
 	)
 	if column, err := columns.Factory("column_name", "DateTime", time.Local); assert.NoError(t, err) {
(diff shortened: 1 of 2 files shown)


miguelff:
Closing this. I will research a better alternative that applies the same encoding semantics as Clickhouse

miguelff:
Actually , I will reopen this:

Even when [new changes were added to support extended ranges in DateTime64](https://github.com/ClickHouse/ClickHouse/pull/9404/files) (released on 24.1.1), the exposed behavior for DateTime32 was not changed, and still represent ranges for unsigned 32 bit integers. So I think the PR is relevant. According to the following test

```sql
Connected to ClickHouse server version 21.5.6 revision 54448.

clickhouse-s0-r0 :) CREATE TABLE ts (ts DateTime32) ENGINE = TinyLog()

clickhouse-s0-r0 :) INSERT INTO ts values (toDateTime(0x0));

INSERT INTO ts VALUES

Query id: c3649a08-ff66-4a40-8421-4cee0d360a44

Ok.

1 rows in set. Elapsed: 0.004 sec.

clickhouse-s0-r0 :) INSERT INTO ts values (toDateTime(0xFFFFFFFF));

INSERT INTO ts VALUES

Query id: 82f674c3-20c3-4cb4-ae39-d8fb49861ddc

Ok.

1 rows in set. Elapsed: 0.003 sec.

clickhouse-s0-r0 :) select * from ts;

SELECT *
FROM ts

Query id: 4674c8df-3993-4e9e-b886-e98c4fff89dc

┌──────────────────ts─┐
│ 1970-01-01 00:00:00 │
│ 2106-02-07 06:28:15 │
└─────────────────────┘
```

PR #404, opened 2021-08-16
Improve CPU utilization by reducing allocation overhead

florianl:
Signed-off-by: Florian Lehner <flehner@optimyze.cloud>

This package builds around the struct `clickhouse` [0] which holds  multiple private elements. One of those is `block` of type `*data.Block`. The struct `Block` [1] manages a private element `buffers` of type `[]*buffer]`. The private type `buffer` [2] is a wrapper around `WriteBuffer` [3] which is based on the package `leakypool` that should provide reusable memory.

The problem with the element `block` of the initial `clickhouse` struct is that it is reset quite often [5] - e.g. in `(*clickhouse)beginTx(...)`, `(*clickhouse)Commit(...)`, `(*clickhouse)Rollback()`. So this potential reusable memory is freed and reallocated [6] on a regular basis. Creating a new element of type `WriteBuffer` [7] increases the pressure on needed garbage collection management.

So instead of preallocating memory that may or may not be reused later on this PR uses instead a `bytes.Buffer`. This results in reduced garbage collection overhead.

Kind regards,
 Florian

[0] https://sourcegraph.com/github.com/ClickHouse/clickhouse-go/-/blob/clickhouse.go?L46
[1] https://sourcegraph.com/github.com/ClickHouse/clickhouse-go/-/blob/lib/data/block.go?L17
[2] https://sourcegraph.com/github.com/ClickHouse/clickhouse-go/-/blob/lib/data/block.go?L289
[3] https://sourcegraph.com/github.com/ClickHouse/clickhouse-go/-/blob/lib/writebuffer/buffer.go?L17
[4] https://sourcegraph.com/github.com/ClickHouse/clickhouse-go/-/blob/lib/leakypool/leaky_pool.go
[5] https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/ClickHouse/clickhouse-go%24+content:%22.block+%3D+nil%22
[6] https://sourcegraph.com/github.com/ClickHouse/clickhouse-go/-/blob/lib/data/block.go?L162
[7] https://sourcegraph.com/github.com/ClickHouse/clickhouse-go/-/blob/lib/writebuffer/buffer.go?L11


filimonov:
It looks like a revert of https://github.com/ClickHouse/clickhouse-go/pull/144 :)

florianl:
It is not a revert of #144 as it replaces the pkg `writebuffer` with `bytes.Buffer`. #144 added `leakypool` to `writebuffer.

kshvakov:
let's try it

PR #411, opened 2021-08-30
Add Decimal128 support

markandrus:
This PR adds support for Decimal128. I've included a test which inserts Decimal128 values (nullable and non-nullable) using Block. The test then reads these values back using the SQL interface, and asserts their equivalence.

Go doesn't support an `int128` or `decimal128` type natively. Rather than choose a particular Decimal128 library to use — which may also imply unnecessary work copying bytes around — we simply deal with the 16 underlying little-endian bytes directly.

This is convenient for me in my application, where I'm copying `decimal128` values as byte arrays out of a Parquet file and into a ClickHouse Block.

Note that the same approach can be extended to Decimal256 and similar types.


markandrus:
thanks for the quick merge! ❤️

williamfeng323:
Appreciate for this PR, it helps me solved my problem.  Many thanks! @markandrus

Issue #415, opened 2021-09-16
Example use of SimpleAggregateFunctions?


ewanwalk:
I find the mention of SimpleAggregateFunctions in `column.go` but cannot seem to find any examples of its usage and everything seems to throw an error for me:

```
column: invalid SimpleAggregateFunction type (SimpleAggregateFunction(sum, Decimal(38, 8)))
```

Could we get an example on how to declare and use an aggregate function?

ewanwalk:
This may be related to #412

Issue #412, opened 2021-09-03
Failed to parse Column type: SimpleAggregateFunction(max, DateTime64(3, 'UTC'))


utkarsh-sib:
When reading from a table with column type `SimpleAggregateFunction(max, DateTime64(3, 'UTC'))` the following code tries to split `max, DateTime64(3, 'UTC')` by `,`. Which results in an array of length 3. With parts `max`,  `DateTime64(3`, `UTC')`.

https://github.com/ClickHouse/clickhouse-go/blob/1ae00d34c574d8706a8fa2f9277865fb3c41617f/lib/column/column.go#L189-L200

This function returns an error if the length of the splitted array is not equal to 2. 
Please check.
Thanks.

kshvakov:
Fixed in [V2](https://github.com/ClickHouse/clickhouse-go/tree/v2)

PR #416, opened 2021-09-16
Fix not supporting SimpleAggregateFunction types not working on types with more than one arguement

ewanwalk:
This change addresses any `SimpleAggregateFunction` which references a type that has more than one argument as outlined in #415 and #412.


kshvakov:
Thank you!

PR #429, opened 2021-11-05
fix Read buffer in binary.Decoder

nikandfor:
In decoder we expect full buffer was read, but in Go Read method can return successfully if it read less than buffer size. I've added io.ReadFull

### DIFF
--- a/lib/binary/binary_test.go
+++ b/lib/binary/binary_test.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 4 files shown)


nikandfor:
The same issue in compressReader.

kshvakov:
Hi, that is unnecessary. See  https://github.com/ClickHouse/clickhouse-go/blob/master/connect.go#L141

nikandfor:
This is against [convention](https://pkg.go.dev/io#Reader) and so is not reusable.

PR #432, opened 2021-11-10
Expose driver for tracing

alvarotuso:


alvarotuso:
Closed this since I intended to open it in the forked repository first

timson:
@alvarotuso did you succeed to expose driver for tracing? is it work with datadog tracer?

alvarotuso:
@timson unfortunately even when we did that it didn't work:

The trace is recorded right after calling QueryContext in the datadog tracer: https://github.com/DataDog/dd-trace-go/blob/13167e6409946881c5706d984562a819b8d82cf2/contrib/database/sql/conn.go#L125

But the row fetch is done async here in the CH driver: https://github.com/ClickHouse/clickhouse-go/blob/ed3cd74e27fb80ed31300afe5d04fb6418beea30/stmt.go#L107

And it only blocks when we try to scan it: https://github.com/ClickHouse/clickhouse-go/blob/b596500be6e91a0a52451641a69c361369a39daf/rows.go#L44

The way we worked around this was to do manual tracing:
```
package tracing

import (
	"context"

	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace"
	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/ext"
	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
)

type TracedFunction func(ctx context.Context) error

// WithSpan starts a span, executes the provided function and handles setting possible errors as tags in the span
// a new context is provided for the traced function
func WithSpan(ctx context.Context, spanType string, serviceName string, resourceName string,
	operationName string, f TracedFunction) error {
	startSpanOpts := []ddtrace.StartSpanOption{
		tracer.SpanType(spanType),
		tracer.ServiceName(serviceName),
		tracer.ResourceName(resourceName),
	}
	span, ctx := tracer.StartSpanFromContext(ctx, operationName, startSpanOpts...)
	err := f(ctx)

	finishSpanOpts := []ddtrace.FinishOption{
		tracer.WithError(err),
	}
	span.Finish(finishSpanOpts...)
	return err
}

// WithSQLSpan starts a span to run a query. serviceName is the name of the database
func WithSQLSpan(ctx context.Context, serviceName string, query string, f TracedFunction) error {
	return WithSpan(ctx, ext.SpanTypeSQL, serviceName, query, "db.sql", f)
}
```

We use this like so:
```
         yourObj := YourObj{}
	err := tracing.WithSQLSpan(ctx, "serviceName", query,
		func(ctx context.Context) error {
			row := yourDb.QueryRowContext(ctx, query, params...)
			return row.Scan(&yourObj)
		},
	)
```
It's more inconvenient but the traces look pretty good

Issue #433, opened 2021-11-11
Return real errors from user-facing methods and implement `ResetSession`


joewreschnig:
The `connect` `Read` and `Write` methods return `driver.ErrBadConn` if any error occurs during reading or writing. However, this masks the real error, which can only be seen if you enable the debug log. However, as noted in several other issue reports `ErrBadConn` must be returned as part of the `driver.Conn` interface to invalidate the connection rather than return it to the pool.

To deal with this Go supports [a `ResetSession` method](https://pkg.go.dev/database/sql/driver#SessionResetter). If the `driver.Conn` implements it, it's called any time a connection will be reused and it can return `ErrBadConn` to the same effect, allowing e.g. the `Commit` method to instead return a more informative error to the user.

A `ResetSession` implemented on `clickhouse` may also be a better place for some of the cleanup logic currently implemented in `Commit` and `Rollback`.

joewreschnig:
If there's no objection to this design I'd be willing to work on it.

Slach:
Great, please feel free to make PR

joewreschnig:
I've put a draft PR for review up at #436. There are three tests I'm having trouble with (also before my changes) and would appreciate help from someone more familiar with the test suite.

williamfeng323:
@joewreschnig are you still working on this PR? It is a good fix because if the ResetSession can throw the ErrBadConn, the BeginTx method could create another tx with a new connection which might resolve the #336

joewreschnig:
As far as I know the PR is ready but I had two test failures that I could not trace to anything in my changes, and one failure related to my changes that also failed on master. I've taken the PR out of draft state to hopefully attract more attention, as well as raise one other minor issue I found. If you're more familiar with the code you could look at #436.

kshvakov:
Added in [V2](https://github.com/ClickHouse/clickhouse-go/blob/v2/clickhouse_std.go#L49)

PR #436, opened 2021-11-15
Implement `SessionResetter` on `*clickhouse` (#433)

joewreschnig:
This allows returning a more helpful error when the connection `Read`
and `Write` fail, while still maintaining the connection pool properly.
(See also issue #433.)

`Test_ConnCheckNegative`, `Test_Context_Timeout`,
`Test_Ping_Context_Timeout`, and `Test_Timeout` have been updated to
check against the new expected error rather than `ErrBadConn`.

I've marked this as a draft because I'm still having issues with three tests, but the failures appear both before and after my changes so I'm not sure I affected them in any way (my testing procedure is `docker-compose up; make test; docker-compose down`).

### `Test_Negative_OpenConnectAndPing`

I'm a little worried about this since what error type is returned is directly related to my changes, but in this case it doesn't seem related; it's `ErrBadConn` before my changes, `EOF` after, but the test isn't expecting either of those:

```
    clickhouse_negative_test.go:34: 
                Error Trace:    clickhouse_negative_test.go:34
                Error:          Should be true
                Test:           Test_Negative_OpenConnectAndPing
--- FAIL: Test_Negative_OpenConnectAndPing (0.00s)
```

### `Test_ArrayT`
Here, I've got no idea, this test seems to just be broken in general:

```
    clickhouse_test.go:724: 
                Error Trace:    clickhouse_test.go:724
                Error:          Received unexpected error:
                                sql: Scan error on column index 14, name "ipv4": unsupported Scan, storing driver.Value type []net.IP into type *[]column.IP
                Test:           Test_ArrayT
--- FAIL: Test_ArrayT (0.02s)
```

### `TestQuerySettings`

This sometimes stalls and times out. As with the other two, this happens both before and after my changes. However, my changes do deal with connection lifetimes and based on frequency I may have made the stalls more likely. Maybe there is some indeterminism leading to a leaked connection in an earlier test?

### DIFF
--- a/clickhouse.go
+++ b/clickhouse.go
@@ -117,6 +117,21 @@ type txOptions struct {
 	ReadOnly  bool
 }
 
+func (ch *clickhouse) ResetSession(ctx context.Context) error {
+	ch.logf("[ResetSession] tx=%t, data=%t", ch.inTransaction, ch.block != nil)
+	switch {
+	case ch.inTransaction:
+		return sql.ErrTxDone
+	case ch.conn.closed:
+		return driver.ErrBadConn
+	}
+	if ch.block != nil {
+		ch.block.Reset()
+		ch.block = nil
+	}
+	return nil
+}
+
 func (ch *clickhouse) beginTx(ctx context.Context, opts txOptions) (*clickhouse, error) {
 	ch.logf("[begin] tx=%t, data=%t", ch.inTransaction, ch.block != nil)
 	switch {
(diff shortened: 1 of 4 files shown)


joewreschnig:
I've taken this PR out of draft state but I'm still unsure about the three issues above.

Specifically regarding `Test_Negative_OpenConnectAndPing` and the eventual call to `(conn *connect) connCheck() error`, I'm also wondering if `io.ErrUnexpectedEOF` would be more appropriate than `io.EOF`, since the expected case is not EOF but open-but-no-data.

kshvakov:
Added in [V2](https://github.com/ClickHouse/clickhouse-go/tree/v2)

PR #439, opened 2021-11-23
adding new types makes it easier for other ORM to use

ChuckWe:
Add partial array types to register in database/sql

type TestModel struct {
    TestArray ArrayInt64 `gorm:"type:Array(Int64)"`
}

### DIFF
--- a/clickhouse.go
+++ b/clickhouse.go
@@ -21,9 +21,14 @@ import (
 )
 
 type (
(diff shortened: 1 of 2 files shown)


kshvakov:
Version 1 is no longer supported and [version 2 ](https://github.com/ClickHouse/clickhouse-go/tree/v2)works with Go types

Issue #440, opened 2021-12-04
fix formateTime error when application timezone is defferent from Clickhouse


tsingsun:
If the time zone of the application server is different from that of CK , helpers.go converts the time format may causes a conditional error.
for example, America/New_York 2006-01-02 19:00 = UTC 2006-01-03
CK server time zone: UTC
,which is treated as toDate instead of expected toDateTime

kshvakov:
Fixed in [V2](https://github.com/ClickHouse/clickhouse-go/tree/v2)

PR #442, opened 2021-12-09
add shopspring/decimal support

tsingsun:
It's not developer-friendly in handling decimal when use clickhouse-go, this PR's new fearture:

1. decimal Scan return [shopspring/decimal]( github.com/shopspring/decimal) object, take the database column's precision
2. decimal `Write` can accept and auto convert a decimal with different precision from database 

 use `shopspring/decimal` directly in internal decimal's Read /Write for the following reasons.
1. [shopspring/decimal]( github.com/shopspring/decimal) is extremely good and  very popular. use it output Is much friendlier than the built-in column.decimal.
2. Read method return value can lazy calculate is better than return native type, and need no more tedious processing
3. The accuracy with which non-raw values are handled correctly require application sync database column's precision,but that is heavy work.

This new is enabled through add `parse_decimal=true` configuration in DSN, false default value keeping the origin usage

[shopspring/decimal]( github.com/shopspring/decimal) Scan method does not support receive selft instance,  you can use one more layer of packaging, a simple way to rewrite the Scan. until another PR for it.
```
type Decimal struct {
    decimal.Decimal
}

func (d *Decimal) Scan(value interface{}) error {
    if v, ok := value.(decimal.Decimal); ok {
        d.Decimal = v
        return nil
    }
    return d.Decimal.Scan(value)
}
```

ps:
In this PR, include a problem fix: the client time zone and the server time zone inconsistency are also included in the time conversion problem see #440


kshvakov:
Hi, @tsingsun. I think it would be great to add `shopspring/decimal` by default in [V2](https://github.com/ClickHouse/clickhouse-go/tree/v2). Can you prepare a PR for the V2 version?

tsingsun:
ok, I need more time to Learn v2

kshvakov:
@tsingsun I added an empty Decimal column with no logic https://github.com/ClickHouse/clickhouse-go/commit/ae91118d89c87b932366291324759b67f129229d

kshvakov:
Decimal32/64 is supported

kshvakov:
Done!

Issue #92, opened 2018-02-23
Expose progress metadata


arctica:
It would be very nice if your driver somehow could expose the progress metadata of a query. I saw in the code that it already handles the necessary part of the ClickHouse protocol (and logs the data when run in debug mode).

I am not sure how the API can expose the data but my first thought was to use a context.WithValue() to provide an unbuffered channel of progress structs to the query executing functions. The driver can then check if the context specified such a channel and attempt to send the progress update to it (but not block) whenever it receives an update from the server.

Example:

```
progressChan := make(chan clickhouse.Progress)

go func() {
    for progress := range progressChan {
        fmt.Printf("Progress: %v / %v rows, %v bytes.\n", progress.Rows, progress.TotalRows, progress.Bytes)
    }
}()

ctx := context.WithValue(context.Background(), clickhouse.ProgressKey, progressChan)
conn.QueryContext(ctx, "SELECT foo FROM bar")
```

kshvakov:
It's a good idea, I'll investigate and implement this feature

arctica:
Happy to hear that the idea was accepted, thanks :)

mikelovato:
@kshvakov excuse me, Is this feature ready to use?

kshvakov:
Added in [V2](https://github.com/ClickHouse/clickhouse-go/blob/v2/examples/native/simple.go#L34)

PR #453, opened 2022-01-11
Implement feature #92[Expose query progress in query context]

lingpeng0314:
This PR is for issue: [Expose progress metadata](https://github.com/ClickHouse/clickhouse-go/issues/92)
#https://github.com/ClickHouse/clickhouse-go/issues/92
Usage:
 ```
        ctx := clickhouse.WithQueryProgress(context.Background())
	rows, err :=connect.QueryContext(ctx, "select * from users limit 4")
        ......(handle rows)
	p:=clickhouse.QueryProgress(ctx)
```

### DIFF
--- a/clickhouse.go
+++ b/clickhouse.go
@@ -267,14 +267,15 @@ func (ch *clickhouse) process() error {
 			ch.logf("[process] <- exception")
 			return ch.exception()
 		case protocol.ServerProgress:
-			progress, err := ch.progress()
+			progress := &Progress{}
+			err := progress.update(ch)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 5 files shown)


kshvakov:
Hi, it already implemented in [V2](https://github.com/ClickHouse/clickhouse-go/tree/v2). We will release V2 beta1 soon

lingpeng0314:
@kshvakov Seems when we use the WithProgress func, we could provide a func to handle the Process object. But we cannot return it out right using `func(*Progress)`? 

func WithProgress(fn func(*Progress)) QueryOption {
	return func(o *QueryOptions) error {
		o.events.progress = fn
		return nil
	}
}

lingpeng0314:
Could i file a PR to expose Progress out which could be fetched by client based on v2?

PR #469, opened 2022-01-27
Add defer release conn in clickhouse PrepareBatch method

igorkim:
Add defer release conn in clickhouse PrepareBatch method


kshvakov:
There is no need to release the connection as it is used in the batch request

f0rtung:
May be Batch should have Release/Close method? How can i release connection if something go wrong after PrepareBatch and before Send/Append?

kshvakov:
@f0rtung The `abort` method is added in v2.0.10

PR #487, opened 2022-02-05
Add DialContext option

slaskis:
It allows for using a cancellable context in a connection. My particular use-case for this is cancelling a long running query when an HTTP-request has been aborted (ex. `ctrl-c` a curl). 

This implementation is backwards compatible and opt-in by using a `DialContext`-option. And since the context is always already provided this was very easy to implement.

Another approach I considered was to add a `*net.Dialer`-option, but it was a bit harder to test the re-dials with. Or using `(*net.Dialer).DialContext` by default but that would not be backwards compatible if code rely on the non-cancellable contexts. Let me know if you think that would be a better fit though and I'll be happy to update this PR.

Btw, I really like working with the new v2 client, thank you!

### DIFF
--- a/clickhouse.go
+++ b/clickhouse.go
@@ -181,13 +181,13 @@ func (ch *clickhouse) Stats() driver.Stats {
 	}
 }
 
-func (ch *clickhouse) dial() (conn *connect, err error) {
+func (ch *clickhouse) dial(ctx context.Context) (conn *connect, err error) {
 	connID := int(atomic.AddInt64(&ch.connID, 1))
 	for num := range ch.opt.Addr {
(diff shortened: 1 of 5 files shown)


kshvakov:
It looks definitely better than the current Dial function. I think it's possible to remove the `Dial` from `Options` since it won't be breaking changes. Can you remove it in this PR?

slaskis:
Did you mean like this? [7401485](https://github.com/ClickHouse/clickhouse-go/pull/487/commits/74014855de4c0aba300290785f84d4c7776e74e7)

kshvakov:
Yes

kshvakov:
Thank you!

PR #495, opened 2022-02-16
Add ability to pass custom type that has ToSQLArgument method as a query argument

f0mster:
Add ability to pass custom type that has `ToSQLArgument() string` method as a query argument.


kshvakov:
The driver should not convert custom user types. You should do this in your app

f0mster:
If so, maybe the driver should return an error if it's receive a type that it doesn't know?

kshvakov:
No, if you want to convert something to a string, you must use the `Stringer` interface https://github.com/ClickHouse/clickhouse-go/blob/v2/bind.go#L140

gingerwizard:
This does seem like something the driver should be doing.

PR #498, opened 2022-02-16
Fix nullable column for batch insertion

grizio:
When using `batch.AppendStruct` with a nullable int field, the insertion does not insert `null` but `0`.

This commit fixes the logic, so it will correctly insert `null` and updates the test ensuring the behavior.


kshvakov:
Thank you!

kshvakov:
There are some problems with this PR. See https://github.com/ClickHouse/clickhouse-go/runs/5218882474?check_suite_focus=true

grizio:
Oops, I will fix them. I only run the test I changed, thinking all tests would be run in PR before merging.

PR #500, opened 2022-02-16
Fix nullable column for batch insertion (fixed)

grizio:
When using `batch.AppendStruct` with a nullable int field, the insertion does not insert `null` but `0`.

This commit fixes the logic, so it will correctly insert `null` and updates the test ensuring the behavior.

### DIFF
--- a/lib/column/nullable.go
+++ b/lib/column/nullable.go
@@ -93,10 +93,9 @@ func (col *Nullable) Append(v interface{}) ([]uint8, error) {
 }
 
 func (col *Nullable) AppendRow(v interface{}) error {
-	switch {
-	case v == nil:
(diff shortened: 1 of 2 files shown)


lib/column/nullable.go
```diff
@@ -93,10 +93,9 @@ func (col *Nullable) Append(v interface{}) ([]uint8, error) {
 }
 
 func (col *Nullable) AppendRow(v interface{}) error {
-	switch {
-	case v == nil:
+	if v == nil || (reflect.ValueOf(v).Kind() == reflect.Ptr && reflect.ValueOf(v).IsNil()) {
```
grizio, line 96: Added `reflect.ValueOf(v).Kind() == reflect.Ptr &&` here

PR #507, opened 2022-02-22
Support []byte in String and FixedString fields

miguelff:
From the commit message:

```
A common use case in clickhouse is to store json in String data.
Golang functions that marshal json return a []byte rather than a
string.

The string type has a few behaviors attached that are handy for dealing
with UTF-8 text, but it'll hold whatever bytes you put in it.

Text-handling behavior in the standard library is often written to work with
[]bytes too, e.g. package bytes mirrors package strings and regexp deals in
either.
```

This was motivated after seeing a braking change in v2 that might have been prevented by this patch. And decided to go ahead and implement it.

### DIFF
--- a/lib/column/string.go
+++ b/lib/column/string.go
@@ -18,6 +18,7 @@
 package column
 
 import (
+	"encoding"
 	"fmt"
 	"reflect"
 
@@ -52,8 +53,14 @@ func (col *String) ScanRow(dest interface{}, row int) error {
 	case *string:
 		*d = v[row]
 	case **string:
-		*d = new(string)
-		**d = v[row]
+		*d = &v[row]
+	case *[]uint8:
+		*d = []uint8(v[row])
+	case **[]uint8:
+		data := []uint8(v[row])
+		*d = &data
+	case encoding.BinaryUnmarshaler:
+		return d.UnmarshalBinary([]uint8(v[row]))
 	default:
 		return &ColumnConverterError{
 			Op:   "ScanRow",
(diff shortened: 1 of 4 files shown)


lib/column/string.go
```diff
@@ -52,8 +53,14 @@ func (col *String) ScanRow(dest interface{}, row int) error {
...
 	case **string:
-		*d = new(string)
-		**d = v[row]
+		*d = &v[row]
+	case *[]uint8:
+		*d = []uint8(v[row])
+	case **[]uint8:
+		data := []uint8(v[row])
+		*d = &data
+	case encoding.BinaryUnmarshaler:
```
kshvakov, line 62: I think it's (`encoding.BinaryUnmarshaler`) too much for a String. And you should use byte instead of uint8^^^

miguelff: So, what do you want me todo:
- Remove the case altogether? (FWIW, this is symmetric with FixedString)
- Amend the line to use []byte, instead of uint8 (although I don't understand very well why, given uint8 is used everywhere else, and they are type aliases)
- A different thing than any of the above?

miguelff:
Also, I don't know if this is the right place to ask this, but as it's related with implicit type conversions I will go ahead:

Currently, if I have an in-memory value that's signed (albeit positive) and try to pass marshal/unmarshal to/from a column that's unsigned, it will panic, and force me to cast the value in memory. This is a change from v1 that might or might not be desired.  

* Would you like to lean on API ergonomics or type safety?
* If leaning on ergonomics, do you want me to open a patch for this type of automatic conversions? 
* Also, are there other automatic type conventions do you want to support?

miguelff:
Stale

PR #510, opened 2022-02-25
Update to allow *int64 for Datetime64 type

rtkaratekid:


rtkaratekid:
If other changes need to be made to allow pointers for other types, please feel free to make them. I've only tested this one though.

lib/column/datetime64.go
```diff
@@ -141,6 +141,8 @@ func (dt *DateTime64) AppendRow(v interface{}) error {
 	switch v := v.(type) {
 	case int64:
 		datetime = v
+	case *int64:
+		datetime = *v
```
kshvakov, line 145: The value can be nil. It's potentially `invalid memory address or nil pointer dereference`

rtkaratekid:
any update on this?

kshvakov:
Also, you should add `*int64` to the Append method

rtkaratekid:
Anything else I need to do to get this approved?

rtkaratekid:
Just checking in to see if I need to do anything else.

gingerwizard:
I've tested this and im satisfied it works. I will follow up with a test enhancement to cover this. In general we will be lenient on types moving forward but not allow any conversions which result in a loss of precision.

Thanks @rtkaratekid for this. Apologies it took soo long. Change of maintainers.

rtkaratekid:
@gingerwizard No problem, I'm just glad it's accepted.

PR #512, opened 2022-02-27
Prepared statement added for AsyncWrite

emrekasg:


kshvakov:
This is not very convenient, since bind parameters are very limited. Therefore, it is assumed that for asynchronous inserts, the user will generate SQL code on the side of his application

emrekasg:
But as I know prepared statements prevents SQL injection, am I wrong? @kshvakov

kshvakov:
Not for ClickHouse. ClickHouse does not support prepared queries

emrekasg:
So, what's the way to prevent SQL injections in ClickHouse? Does this driver already prevent injections automatically?

emrekasg:
I'm closing this pull request. Just learned this driver does use stdlib to escape string

PR #520, opened 2022-03-16
map as value of map

trueifnotfalse:


gingerwizard:
I think we can merge this. I'll follow up with a test enhancement though.

gingerwizard:
@trueifnotfalse thanks for this. I think we can do this alittle more elegantly - i've raised a PR https://github.com/ClickHouse/clickhouse-go/pull/554 with a test.

trueifnotfalse:
@gingerwizard  nested if, are not elegantly , but whatever

PR #530, opened 2022-03-25
rename external/ package to ext to avoid name conflict in repos managed by bazel

beebeeep:
This is a fairly ridiculous issue, but nevertheless - some (usually big) golang projects are managed by the tool named bazel ([link](https://bazel.build/)), which uses some sort of urls (looking like "//path/to/some/package:some_item") to refer entities like packages etc.  

The problem is that in bazel there is a special entity "//external" for external packages (for example, if you import clickhouse-go to your bazel-managed repo, you can refer it as "//external:com_github_clickhouse_clickhouse_go"), and this name ("external") conflicts with "external" subpackage of clickhouse-go.

I understand that renaming library is a breaking change, but unfortunately I don't see any other way to allow projects managed by bazel to use clickhouse-go.

### DIFF
--- a/context.go
+++ b/context.go
@@ -21,7 +21,7 @@ import (
 	"context"
 	"time"
 
-	"github.com/ClickHouse/clickhouse-go/v2/external"
+	"github.com/ClickHouse/clickhouse-go/v2/ext"
 	"go.opentelemetry.io/otel/trace"
 )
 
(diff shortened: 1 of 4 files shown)


alexeypavlenko:
Maybe something more unique than just `ext`, e.g. `external_clickhouse_go_v2`? 
Yes, this is ridiculous, but well ...

ananthc:
I was wondering if there is a timeline to merge this fix. We use Bazel as our build tool and depend on the recent versions of the this project to meet our performance requirements. Any pointers regarding expected timelines for merging this fix would be highly appreciated

Issue #535, opened 2022-04-01
FormatTime broken with v1.5.2+ when TZ is not set (Local)


SLoeuillet:
### Issue description
With 1.5.1, FormatTime didn't use timezone at all, inserting a time.Time like time.Now() worked even if TZ was not set (Local)
With 1.5.2+, FormatTime would write "Local" in the timestamp send to clickhouse that would "fail to load Local timezone"

### How v2 handles that :
```bind.go:134, method format 

        case time.Time:
                switch v.Location().String() {
                case "Local":
                        return fmt.Sprintf("toDateTime(%d)", v.Unix())
                case tz.String():
                        return v.Format("toDateTime('2006-01-02 15:04:05')")
                }
                return v.Format("toDateTime('2006-01-02 15:04:05', '" + v.Location().String() + "')")
```
### How v1.5.2 handles that :

```helpers.go:142

func formatTime(v time.Time) string {
        return v.Format("toDateTime('2006-01-02 15:04:05', '" + v.Location().String() + "')")
}
```

SLoeuillet:
Something like 
```
diff --git a/helpers.go b/helpers.go
index 7e10d06..782c401 100644
--- a/helpers.go
+++ b/helpers.go
@@ -140,5 +140,11 @@ func quote(v driver.Value) string {
 }
 
 func formatTime(v time.Time) string {
+       switch v.Location().String() {
+       case "Local":
+               return fmt.Sprintf("toDateTime(%d)", v.Unix())
+       case tz.String():
+               return v.Format("toDateTime('2006-01-02 15:04:05')")
+       }
        return v.Format("toDateTime('2006-01-02 15:04:05', '" + v.Location().String() + "')")
 }
```

SLoeuillet:
1.5.2+ gives that error with no TZ set :
failed to query CH: code: 1000, message: Exception: Cannot load time zone Local

gingerwizard:
Closed via #536

PR #536, opened 2022-04-01
Try to fix formatTime when TZ is not set (Local)

SLoeuillet:
Try to fix issue #535

### DIFF
--- a/helpers.go
+++ b/helpers.go
@@ -140,5 +140,9 @@ func quote(v driver.Value) string {
 }
 
 func formatTime(v time.Time) string {
+	switch v.Location().String() {
+	case "Local":
+		return fmt.Sprintf("toDateTime(%d)", v.Unix())
+	}
 	return v.Format("toDateTime('2006-01-02 15:04:05', '" + v.Location().String() + "')")
 }


SLoeuillet:
#535

helpers.go
```diff
@@ -140,5 +140,9 @@ func quote(v driver.Value) string {
 }
 
 func formatTime(v time.Time) string {
```
alexey-milovidov, line 142: @gingerwizard I don't understand what it is looking at the diff.
What is `2006-01-02 15:04:05`? Would this change affect performance?

Need your help.

gingerwizard: So if the user sets the timezone to Local in their insert e.g. 

```connect.QueryRow("SELECT COUNT(*) FROM clickhouse_test_select WHERE datetime = ?", time.Date(2017, 1, 20, 14, 0, 0, 0, time.Local))```

Currently we serialized "Local" into the string sent to ClickHouse - wrapped in the toDateTime function. Proposal here is to send epoch instead.
Few options:

1. Send epoch at this PR does
2. Find the timezone t.Zone() and do the current strategy - i.e. dyanmic sys lookup. Likely slower.

I wonder what is most efficient for ClickHouse? @alexey-milovidov  i.e. why not just always send epoch to CH?

Note this is v1 and tests for this are failing - question remains re support for v1.

SLoeuillet: Well, I did try to mimic what v2 does (see https://github.com/ClickHouse/clickhouse-go/issues/535)

gingerwizard: Yeh trying to figure out the authors original intention here..this is for queries..unclear as to the benefit of sending as a string and not just a unit32 - the former CH has to the conversion. Checking with core.

gingerwizard: This fixes the issue and v1 is patches only at this point.

PR #537, opened 2022-04-06
Fix go vet issue: possible misuse of reflect.SliceHeader

forsaken628:
Make compiler happy.

Link https://github.com/golang/go/issues/40701


gingerwizard:
thanks for this @forsaken628 it LGTM.

@genzgd we should add `go vet` to the lint check - which has alot of issues needing resolution.

PR #538, opened 2022-04-06
make it possible to use bool type instead of uint8

hulb:
To mek it possible to use `bool`  instead of `uint8` when a read bool is needed.


alexey-milovidov (approved):
LGTM

PR #543, opened 2022-04-12
clickhouse std use []byte for string to support database/sql scan

tsingsun:
```go
line 239 
func (r *stdRows) Next(dest []driver.Value) error {
....
}
```
if column is a base type of string, in Next method set dest string value,  database/sql will cause a panic err :  unsupported Scan, storing driver.Value type *string into type *string.

and database/sql get []byte value for string value.


gingerwizard:
Thanks @tsingsun your contributions are always hugely appreciated - especially since the original author has stepped away.

I spent quite some time looking into this issue - its reproducible with the following test (which we should add):

```
func TestStdStr(t *testing.T) {
	if conn, err := sql.Open("clickhouse", "clickhouse://127.0.0.1:9000"); assert.NoError(t, err) {
		if err := checkMinServerVersion(conn, 21, 12); err != nil {
			t.Skip(err.Error())
			return
		}

		ctx := clickhouse.Context(context.Background(), clickhouse.WithSettings(clickhouse.Settings{
			"max_block_size": 10,
		}))

		const ddl = `
			CREATE TABLE test_str (
				    Col1 String
				  , Col2 String
				  , Col3 Array(String)
				  , Col4 Nullable(String)
				  , Col5 Array(Nullable(String))
			) Engine Memory
		`
		defer func() {
			conn.Exec("DROP TABLE test_str")
		}()
		if _, err := conn.Exec(ddl); assert.NoError(t, err) {
			scope, err := conn.Begin()
			assert.NoError(t, err)
			batch, err := scope.PrepareContext(ctx, "INSERT INTO test_str (Col1, Col2, Col3, Col4, Col5)")
			assert.NoError(t, err)

			for i := 0; i < 10; i++ {
				var val string
				_, err := batch.Exec("col1", "col2", []string{"col3a", "col3b", "col3c"}, nil, []*string{&val, nil, &val})
				assert.NoError(t, err)
			}

			if err := scope.Commit(); assert.NoError(t, err) {
				rows, err := conn.QueryContext(ctx, "SELECT * FROM test_str")
				if assert.NoError(t, err) {

					for rows.Next() {
						var (
							col1 string
							col2 string
							col3 []string
							col4 string
							col5 []*string
						)
						if err := rows.Scan(&col1, &col2, &col3, &col4, &col5); assert.NoError(t, err) {
							var val string
							assert.Equal(t, "col1", col1)
							assert.Equal(t, "col2", col2)
							assert.Equal(t, []string{"col3a", "col3b", "col3c"}, col3)
							assert.Nil(t, nil, col4)
							assert.Equal(t, []*string{&val, nil, &val}, col5)
						}
					}
				}
			}

		}
	}
}
```

The challenge here is that if a col is a nullable string it will be mapped to a Nullable column on reading the first block. The scan then tries to convert a *string to a *string is you noticed. 
```
  sql: Scan error on column index 3, name "Col4": converting NULL to string is unsupported
```
There is alot indirection and abstraction here and im new to the driver but it appears this is actually correct. A nullable is represented as a *string which allows the value nil. If a reference to a string is passed (deference on the Scan), that means the variable is a string - presumably Scan is preventing this as not all *string values can be represented with string. If you were to have a standard String field it would be fine.

If we merged this we would actually break the case were the column is nil. For example, this fails with your patch applied:

```
func TestStdStr(t *testing.T) {
	if conn, err := sql.Open("clickhouse", "clickhouse://127.0.0.1:9000"); assert.NoError(t, err) {
		if err := checkMinServerVersion(conn, 21, 12); err != nil {
			t.Skip(err.Error())
			return
		}

		ctx := clickhouse.Context(context.Background(), clickhouse.WithSettings(clickhouse.Settings{
			"max_block_size": 10,
		}))

		const ddl = `
			CREATE TABLE test_str (
				    Col1 String
				  , Col2 String
				  , Col3 Array(String)
				  , Col4 Nullable(String)
				  , Col5 Array(Nullable(String))
			) Engine Memory
		`
		defer func() {
			conn.Exec("DROP TABLE test_str")
		}()
		if _, err := conn.Exec(ddl); assert.NoError(t, err) {
			scope, err := conn.Begin()
			assert.NoError(t, err)
			batch, err := scope.PrepareContext(ctx, "INSERT INTO test_str (Col1, Col2, Col3, Col4, Col5)")
			assert.NoError(t, err)

			for i := 0; i < 10; i++ {
				var val string
				_, err := batch.Exec("col1", "col2", []string{"col3a", "col3b", "col3c"}, nil, []*string{&val, nil, &val})
				assert.NoError(t, err)
			}

			if err := scope.Commit(); assert.NoError(t, err) {
				rows, err := conn.QueryContext(ctx, "SELECT * FROM test_str")
				if assert.NoError(t, err) {

					for rows.Next() {
						var (
							col1 string
							col2 stringc
							col3 []string
							col4 string
							col5 []*string
						)
						if err := rows.Scan(&col1, &col2, &col3, &col4, &col5); assert.NoError(t, err) {
							var val string
							assert.Equal(t, "col1", col1)
							assert.Equal(t, "col2", col2)
							assert.Equal(t, []string{"col3a", "col3b", "col3c"}, col3)
							assert.Nil(t, nil, col4)
							assert.Equal(t, []*string{&val, nil, &val}, col5)
						}
					}
				}
			}

		}
	}
}

``` 

with 

```
 sql: Scan error on column index 3, name "Col4": converting NULL to string is unsupported
```

So in summary i think the current behaviour is by design.

gingerwizard (commented):
See comment

tsingsun:
@gingerwizard Thank you for praise.
and I will spent some time on the issue as soon as I get through busy  period.

PR #548, opened 2022-04-22
conn: fix Close() on rows when a timeout during processing

vincentbernat:
There is a race condition when calling rows.Close() after a query was
stopped due to an error (notably when hitting a timeout). The
goroutine processing the query result tries to send the error to the
error channel, before closing both the error and the stream channel
while the Close() method is trying to read from stream before
processing errors. This leads to a deadlock.

This could be fixed by closing the stream channel before sending the
error, but it seems more future-proof to drain both stream and errors
channel in parallel in the Close() method.

### DIFF
--- a/clickhouse_rows.go
+++ b/clickhouse_rows.go
@@ -95,11 +95,26 @@ func (r *rows) Columns() []string {
 }
 
 func (r *rows) Close() error {
-	for range r.stream {
-	}
-	for err := range r.errors {
-		if err != nil {
-			r.err = err
+	active := 2
+	for {
+		select {
+		case _, ok := <-r.stream:
+			if !ok {
+				active--
+				if active == 0 {
+					return r.err
+				}
+			}
+		case err, ok := <-r.errors:
+			if err != nil {
+				r.err = err
+			}
+			if !ok {
+				active--
+				if active == 0 {
+					return r.err
+				}
+			}
 		}
 	}
 	return r.err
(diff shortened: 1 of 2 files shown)


gingerwizard:
Thanks @vincentbernat for raising this and nicely spotted. This indeed can be simulated with the following. Swapping the close(stream) and close(error) is not even full proof in this case - although it admittedly is an extreme case.

```
package issues

import (
	"context"
	"testing"
	"time"

)

func Test548(t *testing.T) {
	var (
		ctx, cancel = context.WithTimeout(context.Background(), time.Second)
		conn, err   = clickhouse.Open(&clickhouse.Options{
			Auth: clickhouse.Auth{
				Database: "default",
				Username: "default",
				Password: "",
			},
			DialTimeout: time.Second,
			Compression: &clickhouse.Compression{
			},
			//Debug: true,
		})
	)
	defer cancel()
	assert.NoError(t, err)
	timeout := time.After(3 * time.Second)
	done := make(chan bool)
	go func() {
		rows, _ := conn.Query(ctx, "SELECT sleepEachRow(0.001) as Col1 FROM system.numbers LIMIT 1000 SETTINGS max_block_size=10;")
		rows.Close()
		done <- true
	}()

	select {
	case <-timeout:
		t.Fatal("Close() deadlocked")
	case <-done:
	}
}
```

gingerwizard:
Happy this solves the issue and will follow up with the test. Thanks @vincentbernat again

PR #554, opened 2022-05-02
Fix map of maps

gingerwizard:
In looking at https://github.com/ClickHouse/clickhouse-go/pull/520 this seemed like a more elegant solution.


gingerwizard:
and credit to @trueifnotfalse for spotting this issue

genzgd (approved):
As discussed there are edge cases where a more sophisticated parsing of the Map column type arguments may be necessary, but this addresses the immediate Map of Map case.

Issue #506, opened 2022-02-22
Data race accessing structMap.cache


flyingmutant:
clickhouse-go v2.0.12, build with `-race` enabled:

```
[clickhouse][conn=2][10.2.46.136:9000][read data] compression=false. block: columns=0, rows=0
WARNING: DATA RACE
Read at 0x00c0031fc330 by goroutine 98:
  runtime.mapaccess2()
      /usr/local/go/src/runtime/map.go:452 +0x0
  github.com/ClickHouse/clickhouse-go/v2.(*structMap).Map()
      .../vendor/github.com/ClickHouse/clickhouse-go/v2/struct_map.go:60 +0x3e4
  github.com/ClickHouse/clickhouse-go/v2.(*rows).ScanStruct()
      .../vendor/github.com/ClickHouse/clickhouse-go/v2/clickhouse_rows.go:79 +0x84
  github.com/ClickHouse/clickhouse-go/v2.(*clickhouse).Select()
      .../vendor/github.com/ClickHouse/clickhouse-go/v2/scan.go:57 +0x826
```

As we can see in [struct_map.go](https://github.com/ClickHouse/clickhouse-go/blob/202ca18e6dd39d9720f3fd60530ab1569c78d378/struct_map.go#L60), access to `cache` is not synchronized. In [conn_query.go](https://github.com/ClickHouse/clickhouse-go/blob/202ca18e6dd39d9720f3fd60530ab1569c78d378/conn_query.go#L79) we can see that `structMap` is shared between different queries.

gingerwizard:
@flyingmutant can i confirm you detect this when you parallel inserts on the same client? The race detector doesn't detect on our current test suite.

flyingmutant:
This race is triggered when doing parallel `SELECT`s, not `INSERT`s (as you can see from the race detector report, which starts with `v2.(*clickhouse).Select()`).

The bug is quite obvious from the code itself: as I wrote in the original report, `structMap` is shared between different queries without synchronizing access to its internal state (`cache`).

gingerwizard:
I believe the gap in our tests is the race detector doesn't detect concurrent access if the queries are using the same struct. We need each query to use a different struct on the same connection to induce this. We will fix and add a test.

PR #567, opened 2022-05-09
Fix 506

gingerwizard:
Thanks to @genzgd for sync map fix and colloboration.


struct_map.go
```diff
@@ -20,17 +20,15 @@ package clickhouse
 import (
 	"fmt"
 	"reflect"
+	"sync"
 )
 
 type structMap struct {
-	cache map[reflect.Type]map[string][]int
+	cache sync.Map
```
gingerwizard, line 27: wonder if its better for this to be a pointer...

gingerwizard:
Closes https://github.com/ClickHouse/clickhouse-go/issues/506

Issue #542, opened 2022-04-11
support mysql placeholder


tsingsun:
The clickhouse statement supports mysql protocol, but v2 go driver placeholder only supports $,@, while mysql uses '? 'As placeholder.  

I hope to support the mysql placeholder

gingerwizard:
I assume you are referring to ClickHouse's ability to communicate via the [MySql protocol](https://clickhouse.com/docs/en/interfaces/mysql/).

This is not something a dedicated ClickHouse driver is likely to support.

tsingsun:
@gingerwizard as @itzikiusa said, V1 supported.backwards compatibility would be great

gingerwizard:
k seems reasonable. It looks like a trivial change if its limited to this.

gingerwizard:
Closed by https://github.com/ClickHouse/clickhouse-go/pull/577

PR #577, opened 2022-05-12
Support for positional arguments

gingerwizard:
Closes https://github.com/ClickHouse/clickhouse-go/issues/542

Benchmark suggests this is actually more performant than numeric.

### DIFF
--- a/bind.go
+++ b/bind.go
@@ -36,7 +36,7 @@ func Named(name string, value interface{}) driver.NamedValue {
 }
 
 var bindNumericRe = regexp.MustCompile(`\$[0-9]+`)
-var bindPositionalRe = regexp.MustCompile(`[?]`)
+var bindPositionalRe = regexp.MustCompile(`[^\\][?]`)
 
 func bind(tz *time.Location, query string, args ...interface{}) (string, error) {
 	if len(args) == 0 {
@@ -71,6 +71,8 @@ func bind(tz *time.Location, query string, args ...interface{}) (string, error) 
 	return bindPositional(tz, query, args...)
 }
 
+var bindPositionCharRe = regexp.MustCompile(`[?]`)
+
 func bindPositional(tz *time.Location, query string, args ...interface{}) (_ string, err error) {
 	var (
 		unbind = make(map[int]struct{})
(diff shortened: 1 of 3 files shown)


bind.go
```diff
@@ -35,32 +35,70 @@ func Named(name string, value interface{}) driver.NamedValue {
...
-var bindNumericRe = regexp.MustCompile(`\$[0-9]+`)
+func bindPositional(tz *time.Location, query string, args ...interface{}) (_ string, err error) {
+	var (
+		unbind = make(map[int]struct{})
+		params = make([]string, len(args))
+	)
+	for i, v := range args {
+		if fn, ok := v.(std_driver.Valuer); ok {
+			if v, err = fn.Value(); err != nil {
+				return "", nil
```
genzgd, line 82: I don't understand why we return an empty query string and no error here?  Won't that cause weirdness and confusion in the caller?

bind.go
```diff
@@ -35,32 +35,70 @@ func Named(name string, value interface{}) driver.NamedValue {
...
-	return bindNumeric(tz, query, args...)
+	if haveNumeric {
+		return bindNumeric(tz, query, args...)
+	}
+	return bindPositional(tz, query, args...)
 }
 
-var bindNumericRe = regexp.MustCompile(`\$[0-9]+`)
+func bindPositional(tz *time.Location, query string, args ...interface{}) (_ string, err error) {
+	var (
```
genzgd, line 75: Can this be simplified just by using a param array?  Since we don't have string keys just indexes seem sufficient.

genzgd (approved):

bind.go
```diff
@@ -35,32 +35,70 @@ func Named(name string, value interface{}) driver.NamedValue {
 	}
 }
 
+var bindNumericRe = regexp.MustCompile(`\$[0-9]+`)
+var bindPositionalRe = regexp.MustCompile(`[?]`)
```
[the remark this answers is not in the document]
gingerwizard: +1

bind.go
gingerwizard: yes this was be consistent with the other methods..i'd need to dig into under what conditions this is used and why we handle it this way. I'll work on a PR to address all 3 methods.

PR #589, opened 2022-05-25
Run test in all subdirectories

orloffv:


orloffv:
@gingerwizard current problem is - tests from subdirectories(issues, std) https://github.com/ClickHouse/clickhouse-go/tree/v2/tests never run in CI
Screen Shot 2022-05-25 at 17 53 35

gingerwizard:
Nice catch. It seems we have a failure on an older version of ClickHouse - let me ascertain if that's an LTS we still want to support. We probably need to revisit the full test matrix as well - periodically on releases.

gingerwizard:
I'm going to update the test matrix now and will then merge this @orloffv

gingerwizard:
https://github.com/ClickHouse/clickhouse-go/pull/601 passes on latest versions. merging.

PR #590, opened 2022-05-25
JSON Support

gingerwizard:
Draft only for JSON support. Relies on https://github.com/ClickHouse/ClickHouse/pull/37482

This support comes in a few steps:

Insert time

- [ ] Insert structs as JSON. Included here.
- [ ] Insert map[string]interface{} as JSON or as a sub field of a struct. Pending
- [ ] Insert JSON as string. Pending. Forces inference to the server.

Query time
- [ ] Response into struct. Included here.
- [ ] Response to map[string]interface{} or as sub field of struct. Pending.

Users left to serialize to string here.

Changes are significant as we need names with the sub columns at query and insert time.

cc @CurtizJ


gingerwizard:
This is now ready for review.

We support all of the above-noted functionality. Some known limitations:

- We don't support inserting structs and maps which have pointers inside them. Whilst easy to add it complicates the code a little. I'd like to refactor and get more feedback before we support.
- if inserting structs or maps, types must be consistent within a batch. Outside of batch we obviously can't check and defer to CH. 
- Within a batch, we don't allow mixed formats e.g. strings with maps/structs. Maps and structs are equivalent (i.e. they are marshaled to tuples so can be mixed).
- []interface{} must be of a consistent type at insert time - we don't downcast types or coerce types like inside CH - mainly as we don't know the context outside of the batch.  For example, strings and ints cant be mixed inside a []interface{}.   If users need flexible type conversion, insert as a string. In future we may do more coercion client-side within the batch but this flexible typing is more of a wider client issue.
- Dimensions and types within an []interface{} must be the same e.g. an interface{} slice with a struct and list is not supported. This is the same behaviour as CH where lists of objects with different dimensions are not supported.
- At query time we do best-effort filling structs passed. Any data for which there is no field is currently ignored.

**This feature is experimental and requires 22.6.1 of CH**

cc @genzgd

ernado:
Why not something more generic, like
```go
type Named struct {
   Name string
   Column Interface
}
```
Or I'm missing something important?

lib/column/json.go
```diff
@@ -0,0 +1,888 @@
...
+	}
+	return nil
+}
+
+func appendStructOrMap(jCol *JSONObject, data interface{}) error {
+	vData := reflect.ValueOf(data)
+	kind := vData.Kind()
+	if kind == reflect.Struct {
+		return iterateStruct(vData, jCol, 0)
+	} else if kind == reflect.Map {
```
genzgd, line 463: nit -- no else after return to avoid a level of indentation (this is a matter of style)

lib/column/tuple.go
```diff
@@ -102,6 +121,260 @@ func (col *Tuple) Row(i int, ptr bool) interface{} {
...
+			field.Set(reflect.ValueOf(uuid))
+			return nil
+		}
+	}
+
+	// check if our target is a string
+	if field.Kind() == reflect.String {
+		field.Set(reflect.ValueOf(fmt.Sprint(value.Interface())))
+		return nil
+	} else if value.CanConvert(field.Type()) {
```
genzgd, line 182: same else after return nit

lib/column/tuple.go
```diff
@@ -111,6 +384,30 @@ func (col *Tuple) ScanRow(dest interface{}, row int) error {
...
+		kind := jType.Kind()
+		if kind == reflect.Struct {
+			rStruct := reflect.New(jType.Type()).Elem()
+			err := col.scanJSONStruct(rStruct, row)
+			if err != nil {
+				return err
+			}
+			jType.Set(rStruct)
+			return nil
+		} else if kind == reflect.Map {
```
genzgd, line 397: same

genzgd (approved):
Amazing amount of work.  At some point it probably needs another pass to reduce some duplication but LGTM for now

ernado:
Great work!

Issue #584, opened 2022-05-19
map data type doesn't work with Exec


alifpay:
### Issue description
Error in map data type

when insert into table throws error

### Example code
```go

type Service struct {
	Id        string            `json:"id"`
	Name      string            `json:"name"`
	Apikey    string            `json:"apikey"`
	DelayTime uint32            `json:"delay_time"`
	Interval  uint32            `json:"interval"`
	Timeout   uint32            `json:"timeout"`
	Info      map[string]string `json:"info"`
	RegDate   time.Time         `json:"regdate"`
	Active    bool              `json:"active"`
}

err = d.conn.Exec(ctx,
		`INSERT INTO polldb.services(id, name, apikey, delayTime, interval, timeout, info, active) VALUES($1, $2, $3, $4, $5, $6, $7, $8)`,
		p.Id, p.Name, p.Apikey, p.DelayTime, p.Interval, p.Timeout, p.Info, p.Active)

```

### Error log
```
message: Cannot parse expression of type Map(String, String) here: map[test:gap nest], true)
```

### Configuration

*Interface:* E.g. `native`
github.com/ClickHouse/clickhouse-go/v2

*Driver version: github.com/ClickHouse/clickhouse-go/v2 v2.0.14

*Go version:* run `go version` in your console

*ClickHouse Server version:*

alifpay:
batch is working, but Exec gives error

```go

package main

import (
	"context"
	"fmt"
	"log"
	"time"

	"github.com/ClickHouse/clickhouse-go/v2"
)

func mapToString(m map[string]string) string {
	if len(m) == 0 {
		return "NULL"
	}
	str := "{"
	for k, v := range m {
		str += fmt.Sprintf("'%s':'%s', ", k, v)
	}
	str = str[:len(str)-2]
	str += "}"
	return str
}

func main() {

	/* mp := make(map[string]string)

	mp["k1"] = "val1"
	mp["k2"] = "val2"
	mp["k3"] = "val3"

	str := mapToString(mp)
	fmt.Println(str)
	return */

	conn, err := clickhouse.Open(&clickhouse.Options{
		Addr: []string{"127.0.0.1:9001"},
		Auth: clickhouse.Auth{
			Database: "default",
			Username: "default",
			Password: "",
		},
		Compression: &clickhouse.Compression{
			Method: clickhouse.CompressionLZ4,
		},
		Settings: clickhouse.Settings{
			"max_execution_time": 60,
		},
[...]

PR #603, opened 2022-06-06
Support maps in values

gingerwizard:
This closes https://github.com/ClickHouse/clickhouse-go/issues/584

I'm not sure this is something we want to support long term. This means user can send maps in params but it seems fragile - mainly as for more complex structures it can't tell the differences between tuples and arrays i.e. the existing code assumes []interface{} is a tuple.

For simple statements this is probably fine but its still susceptible to injection and users should use the new JSON feature.

### DIFF
--- a/bind.go
+++ b/bind.go
@@ -230,7 +230,10 @@ func format(tz *time.Location, v interface{}) (string, error) {
 	case reflect.Map: // map
 		values := make([]string, 0, len(v.MapKeys()))
 		for _, key := range v.MapKeys() {
-			name := key.Interface().(string)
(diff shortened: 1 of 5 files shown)


genzgd (approved):
We should probably put a version minimum on the test at this point.  I think it's safe to assume that people inserting into Map with integer keys have a 22.x version.

gingerwizard:
@genzgd https://github.com/ClickHouse/ClickHouse/pull/24589 prior to 21.8 we need a space before `:`. I've added this so we're backwards compatible. All tests now pass.

Issue #517, opened 2022-03-09
Experimental Watch Query does not work with limit.


effusion:
### Issue description
I created a table and live view(experimental) to watch for updates on the table. It seems that the client does not adhere to the "Limit" clause. This causes the following queries to return all the previous updates to be returned as well.  Over the HTTP Interface of the Clickhouse server, the query works.

### Example code
SQL for creating the database
```sql
SET allow_experimental_live_view = 1;
create table sensor_data ( name varchar(100), time datetime, value Float32)engine=Log;
create live view lv_endplate with refresh 5 as select * from sensor_data sd where sd.name = 'Endplate' order by sd.time desc limit 1;
```

Query with limit
```sql
watch lv_endplate limit 1"
```

Clickhouse client code
```go
package clickhouse

/*
	Example Code for driver
	https://github.com/ClickHouse/clickhouse-go/blob/v2/examples/native/scan_struct/main.go
*/
import (
	"context"
	"eatplanted.com/dashboard-backend/config"
	"fmt"
	"github.com/ClickHouse/clickhouse-go/v2"
	"github.com/ClickHouse/clickhouse-go/v2/lib/driver"
	"io/ioutil"
	"net/http"
	"strings"
	"time"
)

type DbConfig struct {
	UserName string
	Password string
	Url      string
	Database string
}

type Clickhouse struct {
	context.Context
	driver.Conn
}

func Init(config config.AppConfig) *Clickhouse {
	var dbConfig DbConfig
	err := config.LoadConfigByPath("server.clickhouse", &dbConfig)

	if err != nil {
		panic(fmt.Errorf("unable to load clickhouse config: %s", err))
	}
	conn, err := clickhouse.Open(&clickhouse.Options{
		Addr: []string{dbConfig.Url},
		Auth: clickhouse.Auth{
			Database: dbConfig.Database,
			Username: dbConfig.UserName,
			Password: dbConfig.Password,
		},
		Debug:           false,
		DialTimeout:     time.Second,
		MaxOpenConns:    10,
		MaxIdleConns:    5,
		ConnMaxLifetime: time.Hour,
		Compression: &clickhouse.Compression{
			Method: clickhouse.CompressionLZ4,
		},
	})

	if err != nil {
		panic(fmt.Errorf("Unable to get connection to clickhouse %s\n", err))
	}

	ctx := clickhouse.Context(context.Background(), clickhouse.WithSettings(clickhouse.Settings{
		"max_block_size":               10,
[...]

PR #614, opened 2022-06-16
Correction of SELECT work

ortyomka:
Issue #517

The problem was not in WATCH query. The problem can be reproduced with two consecutive SELECT queries with same `dest` parameter. 

In `Select`, a result is simply added to `dest`. Therefore, after the first selection, `dest` will be filled, then the second selection will write to the filled `dest`. In result, `dest` will be a union of the two results. 

 I have added a check for an empty slice. If the slice is not empty, then it is cleared.

сс @gingerwizard @genzgd


gingerwizard:
Awesome catch @ortyomka appreciated.

Issue #574, opened 2022-05-11
Supported type conversions in v2


gingerwizard:
v1 of this driver allowed different types to be passed for a column at insertion type, with this necessary type conversion supported implicitly. Whilst this was convenient, it also allowed the potential loss of precision. Our aspiration for v2 is to support all possible implicit conversions that can be done safely with no precision loss.

Currently, v2 has a number of functional gaps where no precision would be lost and conversion could be safely performed. This issue will be used to track current gaps and the eventual compatibility matrix.

**matrix to be added with pending support**

gwvandesteeg:
Can you add String to UUID and vice versa, especially for those of us that don't use the google UUID library (github.com/google/uuid), but alternatives such as github.com/gofrs/uuid, which for us has a much cleaner API and more expansive functionality.

quan-nguyen2:
I have the same issue with `Date` type, in v1, we are able to use `string` to represent date; but in v2, it returns the error `converting string to Date is unsupported`

fiftin:
Hi @gingerwizard ,

What should be done to close this issue?

gingerwizard:
@fiftin we need to go through and compare v1 and v2 to see where type conversions are missing that also don't loose precision. Then we need to address gaps.
Right now I'm addressing as users report.

Noting @quan-nguyen2 issue I'll fix in next release.

fiftin:
@gingerwizard 

So: time and date to/from string conversion? That's all?

gingerwizard:
Currently supported conversions now documented [here](https://github.com/ClickHouse/clickhouse-go/blob/examples/TYPES.md). This will be merged to main. We welcome PRs to add `X`'s.

DGuang21:
hi! @gingerwizard 
I want to complete the type conversions of string to datetime, and I have completed part of it now, but I am worried that it will not meet the code specifications. Can you tell me whether the implementation of this PR is feasible? https://github.com/DGuang21/clickhouse-go/pull/1

The principle is to match the text through regular rules, and if it is a text time type that conforms to the rules, it will be converted.

gingerwizard:
[...]

PR #616, opened 2022-06-16
Support implicit type conversions

fiftin:
Relates to https://github.com/ClickHouse/clickHouse-go/issues/574

Currently, v2 has a number of functional gaps where no precision would be lost and conversion could be safely performed. This PR adding following implicit conversions:

* UUID <-> string
* IPv4 <-> string
* IPv6 <-> string


lib/column/decimal.go
```diff
@@ -92,6 +92,11 @@ func (col *Decimal) Row(i int, ptr bool) interface{} {
 
 func (col *Decimal) ScanRow(dest interface{}, row int) error {
 	switch d := dest.(type) {
+	case *float64:
```
gingerwizard, line 95: This potentially results in precision loss e.g. float64 can't represent numbers exactly like decimal. This is against a philosophy we apply in v2 that I think v1 got wrong i.e. it allowed precision loss. The user should accept this in their code explicitly vs the driver doing it for them.

gingerwizard:
@fiftin thanks for this. Comfortable with the ip/uuid conversions- they seem useful and we can reuse them elsewhere (@genzgd thinking JSON on read). Not sure on whether allowing float64 and Decimal to be interchangeable is wise - since the former can't represent all values of the latter - see the comment.

lib/column/decimal.go
fiftin: Ok, what about IP and UUID?

lib/column/ipv4_test.go
```diff
@@ -0,0 +1,211 @@
...
+
+	// appending string
+	err := col.AppendRow(invalidIPv4Str)
+	if err == nil {
+		t.Fatalf("AppendRow didn't IP")
+	}
+	switch e := err.(type) {
+	case *ColumnConverterError:
+		if e.Hint != "invalid IP format" {
+			t.Fatalf("expected error hint %s, got %s", "AppendRow didn't IP", e.Hint)
```
gingerwizard, line 21: can we use `require` instead and avoid indentation? See other tests. e.g. https://github.com/ClickHouse/clickhouse-go/blob/v2/tests/json_test.go#L269

lib/column/ipv4_test.go
```diff
@@ -0,0 +1,211 @@
...
+	// appending strings
+	col := IPv4{}
+	err := col.AppendRow("127.0.0.1") // add 1 valid IP
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	_, err = col.Append(strIps)
+	if err == nil {
+		t.Fatal("Append must return error if invalid IP provided")
```
gingerwizard, line 43: see above using require and assert packages in test.

lib/column/ipv4.go
```diff
@@ -51,6 +51,11 @@ func (col *IPv4) Row(i int, ptr bool) interface{} {
 
 func (col *IPv4) ScanRow(dest interface{}, row int) error {
 	switch d := dest.(type) {
+	case *string:
+		*d = col.row(row).String()
+	case **string:
```
gingerwizard, line 56: where is the need for support for this coming from? can you give me an example as to when we need  **string support?

lib/column/ipv4_test.go
fiftin: Done

lib/column/ipv4_test.go
fiftin: Done

lib/column/ipv4.go
fiftin: I relied on String column's ScanRow method.

https://github.com/ClickHouse/clickhouse-go/blob/a27ceead7ef92930e47841cec6d08cc24912c4dd/lib/column/string.go#L61

lib/column/decimal.go
fiftin: Removed

fiftin:
@gingerwizard ,

What do you think about Geo types conversion to GeoJSON?

lib/column/ipv4.go
gingerwizard: yes i noticed this, let me see if that's actually something useful..i might be missing a use case.

lib/column/ipv4.go
```diff
@@ -66,39 +71,93 @@ func (col *IPv4) ScanRow(dest interface{}, row int) error {
...
+			Op:   "Append",
+			To:   "IPv4",
+			From: "IPv6",
+			Hint: "invalid IP version",
+		}
+	}
+	return append(data, IPv4ToBytes(ip)...), nil
+}
+
+func (col *IPv4) appendIPv4(ip net.IP) (err error) {
```
gingerwizard, line 103: Do we need these one liner methods?  Not sure they really refactor a huge amount and just make the code harder to follow

fiftin: Removed

fiftin:
All done, please check.

Issue #597, opened 2022-06-01
Support HTTP  in Driver


gingerwizard:

ortyomka:
Hello, @gingerwizard. I want to clarify this issue.

Does it mean to add HTTP-interface? So, dev could choose Native, Std or HTTP to use.

or

Expand Native Driver to accept both Native and HTTP interfaces? So, dev write addresses with prefix protocol(http://foo:port and clickhouse://foo:port) and driver will select interface by its own.

futuarmo:
@gingerwizard, hello! Is it appropriate to adapt https://github.com/mailru/go-clickhouse implementation?

gingerwizard:
@ortyomka has started this work for the std interface. @ortyomka, ideally we would continue to send the data in native format since this will give us the best performance. Happy to share the work or let you own.

ortyomka:
@gingerwizard Do I get it right that you want to pass requests in native format via HTTP Interface?

crashiura:
@gingerwizard I have another solution, you can see it please, if the solution is good, I will finalize it

gingerwizard:
@ortyomka 

> Do I get it right that you want to pass requests in native format via HTTP Interface?

yes

ortyomka:
@gingerwizard 
Interesting, I didn't see this option in the documentation, but I'll research it.

If you already know how to do this, I would be happy to see an example.

crashiura:
@gingerwizard ok i will raise PR

gingerwizard:
Discussed with @genzgd 

1. We prefer the usage pattern of @ortyomka approach for the user 
2. httpBatch is a nice improvement from @crashiura 
3. `type transport interface {` is a bit ugly..maybe @crashiura idea of a batch would allow this?

Lets proceed with @ortyomka with the batch approach and see if it cleanly separates.

ortyomka:
@crashiura Yes, I will

gingerwizard:
@ortyomka can you summarize current state here and remanining TODOs.

ortyomka:
Now:
- HTTP supported by std driver
- HTTP works with native format queries
- HTTP support Batch, Exec, Query

TODO:
- HTTP URL options (like "query_id")
- HTTP Async Insert
- HTTP support by clickhouse driver V2

This is how I see this, and I will open PR soon

cc @gingerwizard

gingerwizard:
Data race in new http code @ortyomka 

```
make test
==================
[...]

Issue #587, opened 2022-05-20
Default values are ignored in AppendStruct


leonidboykov:
### Issue description
Altering table to add a new column causes an error

```
[AppendStruct]: missing destination name "<column_name>" in <struct_name>"
```

even if I've specified a default value for the column.

Perhaps checking that column has a default value here might solve this issue.

https://github.com/ClickHouse/clickhouse-go/blob/ee79713daf7a10ba3b363797f74b0909be463b49/struct_map.go#L69-L74

### Configuration
*OS:* unrelated

*Interface:* E.g. `native`

*Driver version:* v2.0.12

*Go version:* unrelated

*ClickHouse Server version:* 20.10.6.27

gingerwizard:
Thanks @leonidboykov i'll need to dig into this as I'm not certain we get the default value back over the native protocol (@genzgd do you know?) - the initial INSERT gets the columns back and this is in the information which is matched against the struct.

You might want to consider using JSON support - see https://github.com/ClickHouse/clickhouse-go/pull/590

gingerwizard:
@leonidboykov so this information isn't available or returned by the native protocol by default. We'll therefore need to issue a precursor query to DESCRIBE the table, prior to any INSERT. This has some merit - allowing columns to be committed on insert if they have default values would be a nice feature - I'm just a little hesitant about the additional call.

leonidboykov:
@gingerwizard what about `MustSend` method which disables checks? Or some "unsafe" flags for driver settings?

dovefi:
Is there any new progress

dovefi:
> i got the same problem here. i have upgrade clickhouse table first
> 
> ```
> ALTER TABLE [db_name.]table ADD COLUMN col_name DEFAULT emptyArrayString();
> ```
> 
> and the code have not add this column to struct yet. when i AppendStruct(), it return error
> 
> ```
> clickhouse [AppendStruct]: missing destination name \"xxxxx\" in *xxxxx\
> ```

actually, we can't upgrade both table and production code together

jkaflik:
@khanakia you can work around this behaviour by specifying columns in INSERT clause

khanakia:
Yes, that's true but there are hundreds of columns I have.

I thought Clickhouse would handle it automatically.

jkaflik:
[...]

PR #622, opened 2022-06-20
HTTP Part 1

ortyomka:
Issue #597 

### HTTP Connection Update
Add Query and Execution over Native Format
Query reuse `rows` structure
Execution return text of error

### HTTP Batch
Add HTTP Batch
Implement all methods
It works with Native Format

### Tests
Expand tests with types to check HTTP connection and batch

### Drawbacks
HTTP Format Native has 0 revision, so solution has 2 Drawbacks.
1) Batch should recover table schema. [Link](https://github.com/ClickHouse/clickhouse-go/compare/v2...ortyomka:http/1?expand=1#diff-e23f0c1c92cf54758691c16198e28ad7511ff5c18c3542c343db1e8c4d9ccb11R55)
2) Clear DateTime don't have a Time Zone. https://github.com/ClickHouse/ClickHouse/issues/38209

If you have any ideas how to solve them, I will be grateful

cc @gingerwizard


clickhouse_std.go
```diff
@@ -272,8 +293,12 @@ func (std *nativeTransport) PrepareContext(ctx context.Context, query string) (d
 
 func (std *nativeTransport) Close() error { return std.conn.close() }
 
+type IBatch interface {
+	Append(v ...interface{}) error
+}
+
 type stdBatch struct {
-	batch *batch
+	batch IBatch
```
ortyomka, line 301:
```suggestion
	batch driver.Batch
```
Here is can be `Batch` interface from `lib` directory.

If that's a better solution.

clickhouse_std.go
```diff
@@ -147,51 +148,71 @@ func (d *stdDriver) Open(dsn string) (_ driver.Conn, err error) {
 var ErrHttpNotSupported = errors.New("HTTP: not supported")
 
 type httpTransport struct {
```
gingerwizard, line 150: @ortyomka starting to wonder if we really need the transport - the implementations of the respective transports are identical. Could we just not interface the connection itself instead?

gingerwizard:
@ortyomka 

> Batch should recover table schema. [Link](https://github.com/ClickHouse/clickhouse-go/compare/v2...ortyomka:http/1?expand=1#diff-e23f0c1c92cf54758691c16198e28ad7511ff5c18c3542c343db1e8c4d9ccb11R55)

I think we should probably use a DESCRIBE here. This is what the Python driver does it and I means we don't rely on system table structures which might be more subject to change.


@genzgd 
Whats your solution to the following in the Python driver?

> Clear DateTime don't have a Time Zone. HTTP with Native Format: DateTime without TimeZone  ClickHouse#38209

clickhouse_std.go
gingerwizard: yes i think so

clickhouse_std.go
ortyomka: Yeah, it is what I want to do next

clickhouse_std.go
```diff
@@ -272,8 +293,12 @@ func (std *nativeTransport) PrepareContext(ctx context.Context, query string) (d
 
 func (std *nativeTransport) Close() error { return std.conn.close() }
 
+type IBatch interface {
```
gingerwizard, line 296: I'm curious if these interface introductions increase allocations and impact performance? Id like to see before and after comparisons if possible.

gingerwizard: i.e. its impact on the native interface.

conn_http.go
```diff
@@ -51,14 +61,24 @@ func dialHttp(ctx context.Context, addr string, num int, opt *Options) (*httpCon
 			ResponseHeaderTimeout: opt.ReadTimeout,
 			TLSClientConfig:       opt.TLS,
 		},
-		url:      url,
-		location: time.UTC, // TODO: make configurable
+		url: u,
```
gingerwizard, line 64: Why only MaxIdleConns=1?

gingerwizard: ok reasonable, we pool at a higher level.

conn_http.go
```diff
@@ -51,14 +61,24 @@ func dialHttp(ctx context.Context, addr string, num int, opt *Options) (*httpCon
...
 	}
 
-	if err := connect.ping(ctx); err != nil {
+	rows, err := connect.query(ctx, "SELECT timeZone()")
```
gingerwizard, line 67: @genzgd, presumably you do something similar in the python driver?

clickhouse_std.go
ortyomka: Do you have measuring tests?

conn_http.go
```diff
@@ -69,15 +89,47 @@ type httpConnect struct {
...
+	if len(extra) != 0 {
+		query := u.Query()
+		for key, value := range extra {
+			query.Set(key, value)
+		}
+		u.RawQuery = query.Encode()
+	}
 
-	req, err := http.NewRequest(http.MethodPost, h.url.String(), reader)
+	req, err := http.NewRequest(http.MethodPost, u.String(), reader)
```
gingerwizard, line 132: Can we use the context here? i.e. NewRequestFromContext - thus allowing time outs etc to be propagated?

clickhouse_std.go
ortyomka: Found benchmark dir

conn_http.go
```diff
@@ -93,32 +145,103 @@ func (h *httpConnect) executeRequest(ctx context.Context, req *http.Request) (io
 		return nil, err
 	}
 	if resp.StatusCode != http.StatusOK {
-		return nil, fmt.Errorf("clickhouse [execute]:: got no 200 code('%d')", resp.StatusCode)
+
+		msg, err := readResponse(resp)
+
+		if err != nil {
+			return nil, fmt.Errorf("clickhouse [execute]:: failed to read the response (status code %d): %w", resp.StatusCode, err)
```
gingerwizard, line 152: errors.Wrap?

conn_http.go
```diff
@@ -30,13 +36,17 @@ import (
...
+	query := u.Query()
+	query.Set("default_format", "Native")
+	u.RawQuery = query.Encode()
+
 	if opt.TLS != nil {
-		url.Scheme = "https"
+		u.Scheme = "https"
 	}
 
 	connect := &httpConnect{
```
gingerwizard, line 52: Currently, you use the transport directly vs. a client yet don't exploit this lower level abstraction as far as I can tell?

conn_http.go
```diff
@@ -93,32 +145,103 @@ func (h *httpConnect) executeRequest(ctx context.Context, req *http.Request) (io
...
+	if res != nil {
+		defer res.Close()
+		// we don't care about result, so just discard it to reuse connection
+		_, _ = io.Copy(ioutil.Discard, res)
+	}
+
+	return err
+}
+
+func (h *httpConnect) query(ctx context.Context, query string, args ...interface{}) (*rows, error) {
```
gingerwizard, line 182: can we move to conn_http_query.go to be consistent with native?

conn_http.go
```diff
@@ -93,32 +145,103 @@ func (h *httpConnect) executeRequest(ctx context.Context, req *http.Request) (io
...
+			block, err := readData(decoder)
+			if err != nil {
+				if err != io.EOF {
+					errCh <- err
+				}
+				close(stream)
+				close(errCh)
+				return
+			}
+			stream <- block
```
gingerwizard, line 220: how does the channel get closed?

conn_http_batch.go
```diff
@@ -0,0 +1,192 @@
...
+	"github.com/ClickHouse/clickhouse-go/v2/lib/proto"
+	"io"
+	"io/ioutil"
+	"os"
+	"regexp"
+	"strings"
+)
+
+var splitHttpInsertRe = regexp.MustCompile(`(?i)INSERT INTO\s([\w.]+)(\s\((.*?)\))?`)
+var queryTableSchema = "SELECT name, type FROM system.columns WHERE table = $1 ORDER BY position"
```
gingerwizard, line 36: lets use DESCRIBE

conn_http_batch.go
```diff
@@ -0,0 +1,192 @@
...
+			fmt.Println(err)
+			return
+		}
+		if err = encoder.Flush(); err != nil {
+			fmt.Println(err)
+			return
+		}
+	}()
+
+	req, err := b.conn.prepareRequest(context.Background(), r, map[string]string{"query": b.query})
```
gingerwizard, line 177: shouldn't this be using the context from the batch?

conn_http_batch.go
```diff
@@ -0,0 +1,192 @@
...
+
+	req.Header.Add("Content-Type", "application/octet-stream")
+
+	res, err := b.conn.executeRequest(context.Background(), req)
```
gingerwizard, line 181: see above

conn_http.go
```diff
@@ -69,15 +89,47 @@ type httpConnect struct {
 	location  *time.Location
 }
 
-func (h *httpConnect) prepareRequest(ctx context.Context, query string, args ...interface{}) (*http.Request, error) {
-	query, err := bind(h.location, query, args)
+func writeData(encoder *binary.Encoder, block *proto.Block, name string) error {
```
gingerwizard, line 92: name is unused

conn_http_batch.go
```diff
@@ -0,0 +1,192 @@
...
+	if err != nil {
+		return err
+	}
+
+	go func() {
+		defer w.Close()
+		encoder := binary.NewEncoder(w)
+		if b.block.Rows() != 0 {
+			if err = writeData(encoder, b.block, ""); err != nil {
+				fmt.Println(err)
```
gingerwizard, line 163: can we do better than swallow these errors and print them..maybe an err channel on that we wait on and return the err at method end?

gingerwizard:
great work @ortyomka this is coming along really nicely. Left some comments. I'd like @genzgd to take a look given his experience with ClickHouse HTTP.

Also, in clickhouse_options we have 

```
	switch dsn.Scheme {
	case "http":
		if secure {
			o.TLS = nil
		}
		o.Interface = HttpInterface
```

this seems a bit risky - user wants secure but we silently ignore? maybe error

conn_http_batch.go
ortyomka: Fixed

conn_http.go
ortyomka: Fixed

conn_http_batch.go
ortyomka: Fixed

conn_http_batch.go
ortyomka: Fixed

conn_http_batch.go
ortyomka: Fixed, but it produces more columns, that this query
https://github.com/ClickHouse/clickhouse-go/pull/622/files#diff-e23f0c1c92cf54758691c16198e28ad7511ff5c18c3542c343db1e8c4d9ccb11R62

conn_http.go
ortyomka: Fixed, with context done

conn_http.go
ortyomka: Yes, moved

conn_http.go
ortyomka: Change to Client with custom Transport

conn_http.go
ortyomka: Added

conn_http.go
ortyomka: Added

clickhouse_std.go
ortyomka: I started to do this, but it's not trivial for now. 

What should be in connect interface?
What to do with `release(connect*, error)`?
What to do with return values in PrepareContext? (for now there *batch and *httpBatch)

I have a some idea, will implement soon, but some structures will be changed to interfaces.

conn_http_batch.go
ortyomka: It was breaking change. Sometimes ttl_expression was nothing, EOF was return.

Moreover, `DESCRIBE` return 7 columns, but we need only 2.

And this query is analog of DESCRIBE, but return only needed columns

gingerwizard: it means people need access to system tables to use the driver though...

ortyomka: I think use hack with `SETTINGS describe_include_subcolumns=1`. It always returns 0 or 1

ortyomka: When `Decoder` reads a string of size 0 from `Reader` which points to the last byte. The `Reader` returns `EOF`.

ortyomka: Added check [here](https://github.com/ClickHouse/clickhouse-go/pull/622/commits/04b2026b43bfe536e8792c1b95c2d67fb152b97b#diff-a813b8fd70923e6dfea0a06c69f24da5ac5ce47583cad16f0bc90c8081ae6e21R163)

clickhouse_std.go
ortyomka: `transport` has been removed. Now `stdConnect` represntive connection in std driver. Native driver will be updated later (next PR, I guess).

ortyomka:
@gingerwizard thank you for the review. I fixed all messages, can you review again?

gingerwizard:
@ortyomka ill pick this up late tonight/tmr. Thanks for all the work again!

conn_http.go
```diff
@@ -30,102 +35,178 @@ import (
...
+}
+
+func (h *httpConnect) isBad() bool {
+	if h.client == nil {
+		return true
+	}
+	return false
+}
+
+func writeData(encoder *binary.Encoder, block *proto.Block) error {
```
gingerwizard, line 98: these two result in a new encoder and decoder being created each time? any reason we can't create and hold on the connection like std interface? seems like an unnecessary overhead

conn_http.go
```diff
@@ -30,102 +35,178 @@ import (
...
+	return block.Encode(encoder, 0)
+}
+
+func readData(decoder *binary.Decoder) (*proto.Block, error) {
```
gingerwizard, line 102: see above re encoder

conn_http.go
```diff
@@ -30,102 +35,178 @@ import (
...
+	var block proto.Block
+	if err := block.Decode(decoder, 0); err != nil {
+		return nil, err
+	}
+	return &block, nil
+}
+
+func (h *httpConnect) asyncInsert(ctx context.Context, query string, wait bool) error {
```
gingerwizard, line 110: why not? its supported

conn_http.go
```diff
@@ -30,102 +35,178 @@ import (
...
+	return errors.New("HTTP: not supported")
 }
 
-func (h *httpConnect) prepareRequest(ctx context.Context, query string, args ...interface{}) (*http.Request, error) {
-	query, err := bind(h.location, query, args)
+func readResponse(response *http.Response) ([]byte, error) {
+	var result []byte
```
gingerwizard, line 115: nit: remove spaces

conn_http.go
```diff
@@ -30,102 +35,178 @@ import (
...
 
-	reader := strings.NewReader(query)
+	return buf.Bytes(), nil
+}
+
+func (h *httpConnect) prepareRequest(ctx context.Context, reader io.Reader, extra map[string]string) (*http.Request, error) {
 
-	req, err := http.NewRequest(http.MethodPost, h.url.String(), reader)
+	u := *h.url
+
```
gingerwizard, line 133: nit: remove spaces

conn_http.go
```diff
@@ -30,102 +35,178 @@ import (
...
+		return nil, fmt.Errorf("clickhouse [execute]:: %d code: %s", resp.StatusCode, string(msg))
 	}
 
 	return resp.Body, nil
 }
 
-func (h *httpConnect) ping(ctx context.Context) error {
-	req, err := h.prepareRequest(ctx, "SELECT 1")
+func (h *httpConnect) exec(ctx context.Context, query string, args ...interface{}) error {
```
gingerwizard, line 170: any reason to even have ctx as a parameter?

conn_http.go
```diff
@@ -30,102 +35,178 @@ import (
...
+		defer res.Close()
+		// we don't care about result, so just discard it to reuse connection
+		_, _ = io.Copy(ioutil.Discard, res)
+	}
+
+	return err
+}
+
+func (h *httpConnect) ping(ctx context.Context) error {
+	rows, err := h.query(ctx, nil, "SELECT 1")
```
gingerwizard, line 192: there is a ping endpoint for http i.e. "In health-check scripts use GET /ping request. This handler always returns “Ok.” (with a line feed at the end)."

gingerwizard: I dont think this is that important btw, optional change

clickhouse_options.go
```diff
@@ -155,11 +156,9 @@ func (o *Options) fromDSN(in string) error {
 			InsecureSkipVerify: skipVerify,
 		}
 	}
+	o.Scheme = dsn.Scheme
 	switch dsn.Scheme {
 	case "http":
-		if secure {
```
gingerwizard, line 160: still wonder if we should error here..user specifies secure it should be SSL

conn_http_batch.go
```diff
@@ -0,0 +1,201 @@
...
+	"github.com/ClickHouse/clickhouse-go/v2/lib/driver"
+	"github.com/ClickHouse/clickhouse-go/v2/lib/proto"
+	"io"
+	"io/ioutil"
+	"os"
+	"regexp"
+	"strings"
+)
+
+var splitHttpInsertRe = regexp.MustCompile(`(?i)INSERT INTO\s([\w.]+)(\s\((.*?)\))?`)
```
gingerwizard, line 35: not clear what you're trying to extract with the third group..this seems a little unnecessary complicated after `(?i)INSERT INTO\s([\w.]+)` i.e. `(\s\((.*?)\))?` Maybe I'm missing a case..

Also why (?i) given the use of `ToUpper`? 

Can we anchor also with ^?

conn_http_batch.go
```diff
@@ -0,0 +1,201 @@
...
+		return ErrBatchAlreadySent
+	}
+	if err := b.block.Append(v...); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (b *httpBatch) AppendStruct(v interface{}) error {
+	values, err := b.structMap.Map("AppendStruct", b.block.ColumnsNames(), v, false)
```
gingerwizard, line 114: this is for later but use of DESCRIBE means we can extract default values and address https://github.com/ClickHouse/clickhouse-go/issues/587

clickhouse_std.go
gingerwizard: presumably similar now we have moved the interface up to the stdBatch?

gingerwizard (changes requested):
Thanks @ortyomka we're close. Minor comments - main ones are unused parameters and unnecessary creation of decoder and encoder.

conn_http.go
ortyomka: We cannot change reader and writer in Decoder and Encoder. 
So in batch, we need encoder only once to create a body for a request.
In query, we read a body from a response, so sources are changed each request.

Maybe I don't see some option

conn_http.go
ortyomka: https://github.com/ClickHouse/clickhouse-go/pull/622#discussion_r907280479

conn_http.go
ortyomka: Not supported via HTTP for now, options should be added to URL in query part. I plan to add it in next PR.

conn_http.go
ortyomka: Here `ctx` for computability interfaces, and it will be used by prepare and exec Request.

In prepare, `ctx` will be used for Query Id and Quota Key.  

In exec, `ctx` will be used  in request. `req.WithContext(ctx)`

conn_http.go
gingerwizard: Can we not create the Encoder and Decoder once, store in the connection and reuse across requests?

conn_http.go
gingerwizard: :+1:

conn_http.go
ortyomka: I think yes, if encoder and decoder will be extended with `Reset(input)` method. So we can reset `Writer` and `Reader` inside encoder and decoder, respectively

conn_http_batch.go
ortyomka: About first statement, I try to catch the list of inserted columns, to make a selective insert.
I can remove it, and leave as improvement.

Second and third are good suggestions. I will add it.

conn_http.go
ortyomka: See https://github.com/ClickHouse/clickhouse-go/pull/622/commits/dbd3b95c082962057a7557c2cc52b099ce99d625

conn_http.go
ortyomka: I have removed not used context, because we already use request with context in Prepare

clickhouse_options.go
ortyomka: Scheme will be changed to HTTPS

gingerwizard: This is going to lead to confusing behaviour...a user will have specified http and a http port. We switch the scheme to https due to the secure parameter and we will later error. I'd prefer we just return an error and state "secure must be performed over https".

ortyomka:
@gingerwizard we cannot keep r and w from Pipe, because a request read all from r, so it read until w not closed.

About scheme, yes, I like your suggestion

ortyomka:
I was able to bypass the read restriction with `LimitReader`. We always write to w (Pipe) and then wrap r (Pipe) in `LimitReader` with the right size.

But this immediately brings a disadvantage: When we need to read from `response.Body`, we have to copy its contents into w (Pipe). Whereas reset gives us the option not to do this.

cc @gingerwizard

gingerwizard:
Ok im good @ortyomka merging!

gingerwizard:
@ortyomka please add yourself to this https://github.com/ClickHouse/clickhouse-go/blob/v2/contributors/list

Issue #628, opened 2022-06-23
Uint128, UInt256 integer overflow


ViktorZahar:
### Issue description
An integer overflow occurs when reading unsigned integer values into big.Int.
An error occurs if the value is greater than the maximum value of int and less or equals than the maximum value of uint.
For UInt128 if value > 170141183460469231731687303715884105727 <= 340282366920938463463374607431768211455   
For UInt256 if value > 57896044618658097711785492504343953926634992332820282019728792003956564819967 
<= 115792089237316195423570985008687907853269984665640564039457584007913129639935)

### Example code
```go
func TestBigIntUIntOverflow(t *testing.T) {
	var (
		ctx       = context.Background()
		conn, err = clickhouse.Open(&clickhouse.Options{
			Addr: []string{"127.0.0.1:9000"},
			Auth: clickhouse.Auth{
				Database: "default",
				Username: "default",
				Password: "",
			},
			Compression: &clickhouse.Compression{
				Method: clickhouse.CompressionLZ4,
			},
			MaxOpenConns: 1,
		})
	)
	if assert.NoError(t, err) {
		const ddl = `
		CREATE TABLE test_bigint_uint_overflow (
			  Col1 UInt128,
			  Col2 UInt128,
			  Col3 Array(UInt128),
			  Col4 UInt256,
			  Col5 UInt256,
			  Col6 Array(UInt256)
		) Engine Memory
		`
		defer func() {
			conn.Exec(ctx, "DROP TABLE test_bigint_uint_overflow")
		}()
		if err := conn.Exec(ctx, ddl); assert.NoError(t, err) {
			if batch, err := conn.PrepareBatch(ctx, "INSERT INTO test_bigint_uint_overflow"); assert.NoError(t, err) {
				bigUint128Val := big.NewInt(0)
				bigUint128Val.SetString("170141183460469231731687303715884105729", 10)
				maxUint128Val := big.NewInt(0)
				maxUint128Val.SetString("340282366920938463463374607431768211455", 10)
				bigUint256Val := big.NewInt(0)
				bigUint256Val.SetString("57896044618658097711785492504343953926634992332820282019728792003956564819969", 10)
				maxUint256Val := big.NewInt(0)
				maxUint256Val.SetString("115792089237316195423570985008687907853269984665640564039457584007913129639935", 10)
				var (
					col1Data = bigUint128Val
					col2Data = maxUint128Val

					col3Data = []*big.Int{
						big.NewInt(256),
[...]

PR #629, opened 2022-06-23
Fix Issue #628 Uint128, UInt256 integer overflow

ViktorZahar:
Fix Issue #256 unsigned integer overflow.
I added a field "signed" to the bigint column - maybe it's not necessary

### DIFF
--- a/lib/column/column_gen.go
+++ b/lib/column/column_gen.go
@@ -59,24 +59,28 @@ func (t Type) Column(name string) (Interface, error) {
 			size:   16,
 			chType: t,
 			name:   name,
+			signed: true,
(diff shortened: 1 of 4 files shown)


lib/column/column_gen.go
```diff
@@ -59,24 +59,28 @@ func (t Type) Column(name string) (Interface, error) {
 			size:   16,
 			chType: t,
```
gingerwizard, line 60: This code is generated, you need to also modify the template

gingerwizard:
@ViktorZahar thanks for this and great catch. I'll merge and follow up with the template consistency fix.

PR #630, opened 2022-06-23
Unify Nested, JSON and Tuple

gingerwizard:
1. Restructure tuple logic - like array we use the same code paths for JSON and normal tuples - which means we handle tuples into arrays and maps in the same place. Cleans up the code and removes duplication - but means more flexibility for the user and consistency with JSON vs manual tuples. Cons mean the user has to pass the address to Scan when using JSON - however this is now consistent with all other Scans - and is only a change for JSON (which is experimental).

Results in more reflection but given the plan to expose ch-go this seems fine.

2. Scantype now dynamic for tuples - maps for named tuples and slices for unnamed if Row is called. This is important as it means std interface (which uses Row) can be used with JSON to get a map - unlocks the ability to do JSON in Grafana.
The downside is users using std interface with named tuples have a **breaking change** - consider acceptable after discussing with @genzgd (footprint likely small).

3. ClickHouse std has to have interface{} passed for named tuples (and thus JSON) - this is usually fine but feels odd you cant pass a map - this seems a limitation of golang SQL interface which refuses to covert these types. It allows the map returned to be set to the Interface{} however.

Note i wanted Scan for std interface to use ScanRow under the hood - so the type created is the same as that passed by the user - but this isn't possible as sql.Driver doesn't pass the destination to `Next`.

4. Row methods on tuple and array now simply use the ScanType to determine the type to return vs fixed types - important for tuples as this is now dynamic and based on named or unnamed.

5. We now need to support maps for named tuples as well at insert time - needed due to the scan type change. Note this is a result of (1) and (2) - nested are just Array(Tuple). This means we can use Nested with flatten_nested=0 for free! :partying_face:  The cost is as tuples are stored as []Interface{} a lookup is required to find each col by name, when passing a map[string]interface{} - for this we use a simple index mapping the column name to its offset.

### DIFF
--- a/clickhouse_std.go
+++ b/clickhouse_std.go
@@ -346,7 +346,6 @@ func (r *stdRows) Next(dest []driver.Value) error {
 	if r.rows.Next() {
 		for i := range dest {
 			nullable, ok := r.ColumnTypeNullable(i)
-			// maybe call ScanRow
 			switch value := r.rows.block.Columns[i].Row(r.rows.row-1, nullable && ok).(type) {
 			case driver.Valuer:
 				v, err := value.Value()
--- a/lib/column/array.go
+++ b/lib/column/array.go
@@ -133,7 +133,7 @@ func (col *Array) AppendRow(v interface{}) error {
 	default:
 		elem = reflect.Indirect(reflect.ValueOf(v))
 	}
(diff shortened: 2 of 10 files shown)


genzgd (approved):
LGTM

Issue #623, opened 2022-06-20
Utilise ch-go for lower level implementation


gingerwizard:
@ernado has kindly contributed [ch-go](https://github.com/ClickHouse/ch-go) to ClickHouse. This library offers superior performance at the expense of a more complex and stricter API. It offers excellent fundamentals, however, on which we would like to build this client.

Our objective is for ch-go to represent the low-level client, with this client will provide high-level constructs. ch-go will be used for all column encoding and block formulation. Users will then have an option:

1. ch-go where performance is critical e.g. insert heavy usecases 
2. clickhouse-go when convenience is required and some performance can be satisfied e.g. query use cases.

This ticket will track this work. We aim to minimize the impact on users and make no changes to the API of this client - users using this library only experience performance improvements and lower resource overhead.

PR #646, opened 2022-07-05
Use ch-go for low level implementation

gingerwizard:
Closes https://github.com/ClickHouse/clickhouse-go/issues/623


gingerwizard:
Work in progress. Current proposed approach:

1. proto.Reader replaces the Decoder
2. proto.Buffer replaces the Encoder - this means we have to buffer and then send via Stream

Note we still use the Stream for now - eventually id like to remove this as well and switch in ch-go compression. Not the change in the Column interface `Interface`.

conn_handshake.go
```diff
@@ -25,39 +25,32 @@ import (
...
+	defer c.buffer.Reset()
 	c.debugf("[handshake] -> %s", proto.ClientHandshake{})
 	c.conn.SetDeadline(time.Now().Add(c.opt.DialTimeout))
 	defer c.conn.SetDeadline(time.Time{})
 	{
-		c.encoder.Byte(proto.ClientHello)
-		if err := (&proto.ClientHandshake{}).Encode(c.encoder); err != nil {
-			return err
-		}
+		c.buffer.PutByte(proto.ClientHello)
```
ernado, line 33: you can use chproto.ClientHello here

conn_http_query.go
```diff
@@ -62,9 +63,10 @@ func (h *httpConnect) query(ctx context.Context, release func(*connect, error),
 
 	go func() {
 		for {
-			block, err := h.readData()
+			block, err := h.readData(reader)
 			if err != nil {
-				if err != io.EOF {
+				// ch-go wraps EOF files
```
ernado, line 68: probably "ch-go wraps EOF errors"

ernado (approved):
Great job, looks good in general.

It's hard to fully review it tho.

I see that we still have proto package, hope that we will use ch-go/proto for that later.

conn_handshake.go
gingerwizard: Better to just rip out handshake code completely on next PR

PR #649, opened 2022-07-08
use make file for test + docs

gingerwizard:
Single place where tests are executed from. This also ensures we use the race detector - we missed an issue with this recently. 

and doc changes for HTTP support and ch-go usage


Makefile
```diff
@@ -10,6 +10,7 @@ test:
 	@go install -race -v
 	@go test -race -timeout 30s -count=1 -v .
```
ernado, line 11: Probably just 

```
@go test -race -timeout 30s -count=1 -v ./...
```
?

PR #650, opened 2022-07-08
Http compression

gingerwizard:
Adds compression over http - currently using lz4 at block level only per https://clickhouse.com/docs/en/interfaces/http/#compression. Will add other compression options later - mainly zstd which ch-go supports as well as gzip etc for HTTP.

Note compression can be enabled via `OpenDb` with settings or via `compress=true` in the dsn.

We also introduce the `HttpsInterface` type and fix an issue with OpenDb not working @ortyomka i.e. it wasn't setting the scheme (which we no longer expose). I think needing to specify the interface in `OpenDb` is better than a scheme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,77 @@ This can be achieved by modifying the DSN to specify the http protocol.
 http://host1:9000,host2:9000/database?dial_timeout=200ms&max_execution_time=60
 ```
 
+Alternatively, use `OpenDB` and specify the interface type.
+
+```go
+conn := clickhouse.OpenDB(&clickhouse.Options{
+	Addr: []string{"127.0.0.1:8123")},
+	Auth: clickhouse.Auth{
+		Database: "default",
+		Username: "default",
+		Password: "",
+	},
+	Settings: clickhouse.Settings{
+		"max_execution_time": 60,
+	},
+	DialTimeout: 5 * time.Second,
+	Compression: &clickhouse.Compression{
+		Method: clickhouse.CompressionLZ4,
+	},
+	Interface: clickhouse.HttpInterface,
+})
+```
+
+## Compression
+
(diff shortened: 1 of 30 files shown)


README.md
```diff
@@ -110,6 +110,77 @@ This can be achieved by modifying the DSN to specify the http protocol.
...
+		Password: "",
+	},
+	Settings: clickhouse.Settings{
+		"max_execution_time": 60,
+	},
+	DialTimeout: 5 * time.Second,
+	Compression: &clickhouse.Compression{
+		Method: clickhouse.CompressionLZ4,
+	},
+	Interface: clickhouse.HttpInterface,
```
ernado, line 130: Please rename `HttpInterface` to `InterfaceHTTP`, in go we use all caps for abbrevations and is better to prefix the constant name with type, not postfix.

Also, can we use Protocol for this? I know that Interface is from clickhouse, but protocol is more common. Anyway, I'm ok with Interface.

gingerwizard: agreed protocol make sense. Noone is using this yet so its change now.

gingerwizard: @ernado moved to `Protocol` i.e. 
```
type Protocol int

const (
	Native Protocol = iota
	Http
	Https
)
```

clickhouse_options.go
```diff
@@ -49,12 +49,12 @@ const (
...
 
-type InterfaceType int
+type Protocol int
 
 const (
-	NativeInterface InterfaceType = iota
-	HttpInterface
-	HttpsInterface
+	Native Protocol = iota
+	Http
```
ernado, line 56: Please use `HTTP` and `HTTPS` here as per https://github.com/golang/go/wiki/CodeReviewComments#initialisms

ernado (approved):

PR #656, opened 2022-07-11
Clean up compression + add zstd

gingerwizard:
This change shouldn't be breaking for users since we wrap ch-go compression types. `CompressionMethod` will expand to include HTTP-specific formats over time.

I'm a little torn on `NONE` - I've not made it `CompressionNONE` to preserve compatibility - suggestions welcome @ernado @genzgd 

This also adds zstd compression for free (from ch-go) over both HTTP and native. This is for both reading and writing - note we didn't support reading compressed blocks over HTTP before.

Last PR before 2.3.0 release


ernado:
We can use CompressionNone. NONE is non-idiomatic. CompressionZSTD, CompressionGZIP, etc are OK.

If you want to preserve backward compatibility, you can add deprecation notice:
```go
// something like that:

const CompressionNone = 12345

// Deprecated: use CompressionNone
const CompressionNONE = CompressionNone
```

Actually None and Disabled compressions should have different semantics in ClickHouse, None is "compute and verify checksums only" mode.

gingerwizard:
We don't need backward compatibility with `CompressionNONE` rather we need `NONE` - which is even worse @ernado 

I don't think this is a stronger enough case for the backwards compatibility tbh - most users will not have explicitly state NONE - rather they will have not specified compression. Moving to `CompressionNone`

ernado:
Ok, if we need "only checksums" mode we can name it explicitly, like `CompressionOnlyChecksums`

PR #660, opened 2022-07-13
Handle special column chars in JSON

gingerwizard:
closes https://github.com/ClickHouse/clickhouse-go/issues/645

For JSON is we define the column names from the data (vs the user needing to escape in DDL). We therefore need to ensure these are properly escaped.

Follows [https://clickhouse.com/docs/en/sql-reference/syntax/#identifiers](https://clickhouse.com/docs/en/sql-reference/syntax/#identifiers)

We also need to ensure ``` and `\` are correctly escaped - see tests.

Keywords currently not supported - users would need to ensure their column names don't match. If we see a demand maybe we can do something.

### DIFF
--- a/tests/tuple_test.go
+++ b/tests/tuple_test.go
@@ -139,7 +139,8 @@ func TestNamedTuple(t *testing.T) {
 		})
 	)
 	require.NoError(t, err)
-	if err := checkMinServerVersion(conn, 21, 9, 0); err != nil {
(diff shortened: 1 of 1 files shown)


gingerwizard:
Need to skip a test because of https://github.com/ClickHouse/ClickHouse/pull/36544

ernado (approved):

Issue #651, opened 2022-07-08
Support Compression Methods


gingerwizard:
Currently, we support lz4 only at a block level. ClickHouse supports more than this, specifically:

For native:

- [ ] zstd - this is supported in ch-go. Exposing should be simple.

For HTTP via `Accept-Encoding`:

- [ ] zstd
- [ ] ~lz4 (this would be a HTTP transport level, probably not needed given we support at block level)~
- [ ] br
- [ ] gz
- [ ] xz
- [ ] br2

gingerwizard:
xz compression is the remaining task here. I would rather not write a specific lib to achieve this - I don't think its higher enough value. The main option appears to be https://github.com/ulikunitz/xz. This isn't really focused on streaming use cases though and doesn't support Reset or Flush - we'd need to create a new reader and writer each time adding GC pressure. Whilst this is not critical, its also not compatible with our compression pool approach. There is a rewrite however which might address this https://github.com/ulikunitz/xz/tree/rewrite. Will wait for rewrite and reassess when its released. 

If xz is an urgent feature for your usecase please raise here and we'll reconsider.

laeg:
created an issue for xz to track need.

closing this issue out as the rest have been completed.

PR #664, opened 2022-07-15
Gzip compression

gingerwizard:
This adds gzip compression to the http layer - this isn't supported for native.

If the user requests gzip we only compress blocks currently - compressing queries seems unnecessary. 
For decompression, we request ClickHouse only send for queries - again for inserts we are unlikely to benefit. This is subject to change but this aligns with ZSTD and LZ4.

To avoid creating readers and writers for each request, we utilize a pool. This pool creation, as well as the compression and decompression, can probably be abstracted away as we add other compression formats e.g. deflate.

Note to receive compressed responses the user needs to set `enable_http_compression` - either in the connection, via  the context or in the server config. I considered setting this automatically if gzip was requested but this will error if the user doesn't have permissions.

https://github.com/ClickHouse/clickhouse-go/issues/651

### DIFF
--- a/clickhouse_options.go
+++ b/clickhouse_options.go
@@ -192,9 +192,14 @@ func (o *Options) fromDSN(in string) error {
 	}
 	o.scheme = dsn.Scheme
 	switch dsn.Scheme {
-	case "http", "https":
+	case "http":
 		if secure {
 			return fmt.Errorf("clickhouse [dsn parse]: http with TLS specify")
+		}
+		o.Protocol = HTTP
+	case "https":
+		if !secure {
+			return fmt.Errorf("clickhouse [dsn parse]: https without TLS")
 		}
 		o.Protocol = HTTP
 	default:
--- a/clickhouse_options.go
+++ b/clickhouse_options.go
@@ -75,7 +75,6 @@ type Protocol int
 const (
 	Native Protocol = iota
 	HTTP
-	HTTPS
 )
 
 func (p Protocol) String() string {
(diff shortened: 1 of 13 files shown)


clickhouse_options.go
```diff
@@ -166,12 +195,12 @@ func (o *Options) fromDSN(in string) error {
 	}
 	o.scheme = dsn.Scheme
 	switch dsn.Scheme {
-	case "http":
+	case HTTP.String():
```
ernado, line 198: Probably we can leave "http" and "https" here as string literals, and keep only `HTTP` protocol.

Because https is "secure HTTP", same as TLS is "secure native", just for consistency.

So, IMO it sould be "HTTPS, HTTP, Native, SecureNative" or "HTTP, Native" and "secure / not secure"

gingerwizard: yes this has bugged me as well. For native we use the presence of TLS config to denote a secure connection i.e. https://github.com/ClickHouse/ClickHouse-go#tlsssl
Let's drop HTTPS and detect using SSL for consistency.

clickhouse_options.go
```diff
@@ -195,16 +192,11 @@ func (o *Options) fromDSN(in string) error {
 	}
 	o.scheme = dsn.Scheme
 	switch dsn.Scheme {
-	case HTTP.String():
+	case "http", "https":
```
ernado, line 195: I think that we should leave "http" and "https" cases here, like 

```go
	case "http":
		if secure {
			return fmt.Errorf("clickhouse [dsn parse]: http with TLS")
		}
		o.Protocol = HTTP
	case "https":
		if !secure {
			return fmt.Errorf("clickhouse [dsn parse]: https without TLS")
		}
		o.Protocol = HTTP
```

ernado (approved):

PR #667, opened 2022-07-18
Add support for zlib/deflate

gingerwizard:
Refactors compression so it's easier to add methods. In the process, added deflate.

The only controversial point is that we introduce a setting to allow users to set the compression level. Valid values are 1-9. By default, a value of 0 (i.e., the user doesn't set) is interpreted as level 3 - this is the ClickHouse default.


ernado (approved):
looks ok

PR #668, opened 2022-07-18
Compress options in DSN

gingerwizard:
Allow us to specify `compress=<method>` and `compress_level=<level>` in the DSN. `compress` still accepts `true` which defaults to lz4. Compression level defaults to 3  - consistent with ClickHouse defaults.

### DIFF
--- a/clickhouse_options.go
+++ b/clickhouse_options.go
@@ -171,7 +171,7 @@ func (o *Options) fromDSN(in string) error {
 				}
 			}
 		case "compress_level":
-			if level, err := strconv.ParseInt(params.Get(v), 10, 8); err != nil {
+			if level, err := strconv.ParseInt(params.Get(v), 10, 8); err == nil {
 				if o.Compression == nil {
 					o.Compression = &Compression{
 						// a level alone doesn't enable compression
--- a/clickhouse_options.go
+++ b/clickhouse_options.go
@@ -56,6 +56,14 @@ const (
 	CompressionDeflate = CompressionMethod(0x96)
 )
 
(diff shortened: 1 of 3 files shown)


clickhouse_options.go
```diff
@@ -56,6 +56,14 @@ const (
 	CompressionDeflate = CompressionMethod(0x96)
```
ernado, line 56: nit: you can use github.com/dmarkham/enumer 

```go
//go:generate go run github.com/dmarkham/enumer -type CompressionMethod -trimprefix Compression --transform lower
```

ernado: this will require something like this:
```go
//go:generate go run github.com/dmarkham/enumer -transform lower -type CompressionMethod -trimprefix Compression -output clickhouse_options_enum.go

type CompressionMethod byte

func (c CompressionMethod) String() string {
	switch c {
	case CompressionNone:
		return "none"
	case CompressionZSTD:
		return "zstd"
	case CompressionLZ4:
		return "lz4"
	case CompressionGZIP:
		return "gzip"
	case CompressionDeflate:
		return "deflate"
	default:
		return ""
	}
}

const (
	CompressionNone    CompressionMethod = CompressionMethod(compress.None)
	CompressionLZ4     CompressionMethod = CompressionMethod(compress.LZ4)
	CompressionZSTD    CompressionMethod = CompressionMethod(compress.ZSTD)
	CompressionGZIP    CompressionMethod = CompressionMethod(0x95)
	CompressionDeflate CompressionMethod = CompressionMethod(0x96)
)
```

probably not worth it

Issue #647, opened 2022-07-05
OpenDB/Open mutate the Options struct, preventing reuse


rolinh:
### Issue description

The [Open](https://github.com/ClickHouse/clickhouse-go/blob/b100034624aa6c2110d01e8cf43307017a62e1f2/clickhouse.go#L74-L81) and [OpenDB](https://github.com/ClickHouse/clickhouse-go/blob/b100034624aa6c2110d01e8cf43307017a62e1f2/clickhouse_std.go#L88) functions mutate the [Options](https://github.com/ClickHouse/clickhouse-go/blob/b100034624aa6c2110d01e8cf43307017a62e1f2/clickhouse_options.go#L67) struct that is passed as a parameter (see [here](https://github.com/ClickHouse/clickhouse-go/blob/b100034624aa6c2110d01e8cf43307017a62e1f2/clickhouse.go#L75) and [here](https://github.com/ClickHouse/clickhouse-go/blob/b100034624aa6c2110d01e8cf43307017a62e1f2/clickhouse_std.go#L104) respectively, via [setDefaults](https://github.com/ClickHouse/clickhouse-go/blob/b100034624aa6c2110d01e8cf43307017a62e1f2/clickhouse_options.go#L177-L196)). This prevents reusing the options struct instance as shown in the code below.

This issue was initially discussed [here](https://github.com/ClickHouse/clickhouse-go/issues/570#issuecomment-1123350032). A potential non-breaking fix would consist of making a deep copy of the options struct but this is rather ugly. A better fix could be to, e.g., have a function returning default values when unset and user provided values if set.

### Example code

```go
package main

import (
        "log"

        "github.com/ClickHouse/clickhouse-go/v2"
)

func main() {
        options := &clickhouse.Options{}
        db := clickhouse.OpenDB(options)
        err := db.Ping()
        if err != nil {
                log.Fatal("db ", err)
        }
        db2 := clickhouse.OpenDB(options)
        err = db2.Ping()
        if err != nil {
                log.Fatal("db2 ", err)
        }
}
```

### Error log

```
2022/07/05 15:38:06 db2 cannot connect. invalid settings. use SetMaxIdleConns,SetMaxOpenConns,SetConnMaxLifetime (see https://pkg.go.dev/database/sql)
exit status 1
```

### Configuration
*OS:* Arch Linux

*Interface:* `native` and `database/sql`

*Driver version:* `v2.2.0`

[...]

PR #675, opened 2022-07-20
Don't modify options

gingerwizard:
closes https://github.com/ClickHouse/clickhouse-go/issues/647

Not completely happy with this - we pass options as a pointer and modify their contents, preventing reuse or at least leading to confusion.
To address we make the `setDefaults` receive a value instead of a pointer - the copy is lightweight as any internal structs are pointers themselves. This mixes semantics though - options methods sometimes have a pointer or value receiver.

@rolinh for thoughts

### DIFF
--- a/clickhouse.go
+++ b/clickhouse.go
@@ -72,11 +72,11 @@ func (e *OpError) Error() string {
 }
 
 func Open(opt *Options) (driver.Conn, error) {
-	opt.setDefaults()
+	o := opt.setDefaults()
 	return &clickhouse{
-		opt:  opt,
-		idle: make(chan *connect, opt.MaxIdleConns),
-		open: make(chan struct{}, opt.MaxOpenConns),
+		opt:  o,
+		idle: make(chan *connect, o.MaxIdleConns),
+		open: make(chan struct{}, o.MaxOpenConns),
 	}, nil
 }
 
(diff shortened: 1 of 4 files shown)


ernado:
Is the nil value really supported? Probably this should not be a pointer at all :thinking:

ernado:
Agree. Probably we should handle nil case explicitly to avoid null pointer dereference.

gingerwizard:
> Probably we should handle nil case explicitly to avoid null pointer dereference.

+1 will add to this PR

I'll also start a ticket to track v3 changes we'd like to make. Not anticipating a major bump soon so this is just unfortunate technical debt we'll track until then.

gingerwizard:
@ernado nil options means a default Options instance now. Given we don't currently have a default address, this means queries will fail with `ErrAcquireConnNoAddress`. We may wish to add a default address later which would mean `nil` would be supported.

ernado:
>  Given we don't currently have a default address

We can just use `clickhouse-client` defaults (same as ch-go), localhost and default port.-

gingerwizard:
Ok ill follow up with PR. I was hesitant but the clickhouse-client argument is a strong one.

rolinh:
@gingerwizard Sorry, I didn't have time to review this work before it got merged. I think this PR ultimately addresses #647. However, as food for thoughts, have you considered using functional options rather than an option struct? I can open a draft PR if you're interested in what this would look like with this library.

gingerwizard:
@rolinh would be interesting to see yes. We can open it up for discussion for the v3 roadmap if it requires a breaking change.

Issue #682, opened 2022-07-21
Support sql.Null* types


nickbenzoni:
Converting from V1 to V2 - we currently use several `sql.Null*` types throughout our stack and are now receiving errors.

Example using `sql.NullFloat64`:
``` sql: Scan error on column index 2, name "max_value": converting driver.Value type *float64 ("0x140001ec190") to a float64: invalid syntax```

Solution I would like: Support for `sql.Null*` types as in V1 driver  :)

gingerwizard:
Seems like a good addition. Will aim for next release.

gingerwizard:
First pass on strings https://github.com/ClickHouse/clickhouse-go/pull/686
If we're good with the approach rest should be trivial.

gingerwizard:
Closed by #686 

Will cover all sql.Null* except sql.NullBytes - no obvious need for this yet. Also each sql.Null* just aligned with its obvious equivalent e.g. we don't support sql.NullInt16 for Int64 - sql.NullInt16 works with Int16.

PR #686, opened 2022-07-22
support sql.nullables

gingerwizard:
This starts to address https://github.com/ClickHouse/clickhouse-go/issues/682

@ernado need your feedback on this one. To handle `Nullable` fields correctly I check to see if the variable is a `sql.Scanner` at Scan time and `driver.Valuer` at append time - to ensure they are encoded and decoded correctly. If we're good with this approach I can add the remaining sql.Null* types for their respective fields.

### DIFF
--- a/lib/column/bool.go
+++ b/lib/column/bool.go
@@ -18,6 +18,7 @@
 package column
 
 import (
+	"database/sql"
 	"fmt"
 	"github.com/ClickHouse/ch-go/proto"
 	"reflect"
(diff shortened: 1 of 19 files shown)


gingerwizard:
@genzgd @ernado ready for review. I think we cover all sql.Null* except sql.NullBytes - no obvious need for this yet. Also each sql.Null* just aligned with its obvious equivalent e.g. we don't support sql.NullInt16 for Int64.

genzgd (approved):

PR #688, opened 2022-07-24
Convert byte slices to string

### LINKED ISSUE #687: Can't insert a slice of bytes into a String
### Issue description
When trying to insert a byte slice into a String ClickHouse data type I get an error. This used to be possible in v1 and I'm trying really hard to understand why this would not be allowed in v2.

From the ClickHouse documentation for the String type: "Strings of an arbitrary length. The length is not limited. **The value can contain an arbitrary set of bytes, including null bytes.** The String type replaces the types VARCHAR, BLOB, CLOB, and others from other DBMSs."

Given that the String data type is used to store "any bytes". Why would you restrict v2 to only insert "strings" into String types?

Thanks!

### Example code
```go
package main

import (
	"context"
	"fmt"
	"log"
	"time"

)

const ddl = `
CREATE TABLE benchmark (
	  Col1 String
) Engine Null
`

func benchmark(conn clickhouse.Conn) error {
	batch, err := conn.PrepareBatch(context.Background(), "INSERT INTO benchmark")
	if err != nil {
		return err
	}
	for i := 0; i < 1_000_000; i++ {
		err := batch.Append(
			[]byte("Golang SQL database driver"),
		)
		if err != nil {
			return err
		}
	}
	return batch.Send()
}

func main() {
	var (
		ctx       = context.Background()
		conn, err = clickhouse.Open(&clickhouse.Options{
			Auth: clickhouse.Auth{
				Database: "default",
				Username: "default",
				Password: "",
			},
			//Debug:           true,
			DialTimeout:     time.Second,
			MaxOpenConns:    10,
			MaxIdleConns:    5,
			ConnMaxLifetime: time.Hour,
		})
	)
	if err != nil {
		log.Fatal(err)
	}
	if err := conn.Exec(ctx, "DROP TABLE IF EXISTS benchmark"); err != nil {
		log.Fatal(err)
	}
	if err := conn.Exec(ctx, ddl); err != nil {
		log.Fatal(err)
	}
	start := time.Now()
	if err := benchmark(conn); err != nil {
		log.Fatal(err)
	}
	fmt.Println(time.Since(start))
}

```

### Error log
```
2022/07/22 14:29:50 clickhouse [AppendRow]: Col1 clickhouse [AppendRow]: converting []uint8 to String is unsupported
```

a-dot:
This PR closes #687

### DIFF
--- a/lib/column/string.go
+++ b/lib/column/string.go
@@ -85,6 +85,8 @@ func (col *String) AppendRow(v interface{}) error {
 		default:
 			col.col.Append("")
 		}
+	case []byte:
+		col.col.Append(string(v))
 	case nil:
 		col.col.Append("")
 	default:
--- a/lib/column/string.go
+++ b/lib/column/string.go
@@ -115,6 +115,11 @@ func (col *String) Append(v interface{}) (nulls []uint8, err error) {
 				nulls[i] = 1
 			}
 		}
+	case [][]byte:
+		nulls = make([]uint8, len(v))
+		for i := range v {
+			col.col.Append(string(v[i]))
+		}
 	default:
 		return nil, &ColumnConverterError{
 			Op:   "Append",


gingerwizard (changes requested):
need to support in `Append` as well

PR #695, opened 2022-07-26
consider tz in date

gingerwizard:
This passes based on the patch from 0.47.2.

We should explicitly document though that Date and Date32 are persisted as-is - they are not converted to UTC. This can lead to the retrieved date (which will be in UTC as we have no means of knowing the tz in Date) being different than the inserted - this is the only case in the entire driver we have this behavior @ernado

### DIFF
--- a/tests/issues/693_test.go
+++ b/tests/issues/693_test.go
@@ -1,0 +1,77 @@
+package issues
+
+import (
+	"database/sql"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"testing"
+	"time"
+)
+
+func Test693(t *testing.T) {
+	conn, err := sql.Open("clickhouse", "clickhouse://127.0.0.1:9000")
+	require.NoError(t, err)
+	const ddl = `
+			CREATE TABLE test_date (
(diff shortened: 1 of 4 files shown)


tests/issues/693_test.go
```diff
@@ -0,0 +1,50 @@
...
+	require.NoError(t, err)
+	require.NoError(t, scope.Commit())
+	var (
+		result1 result
+	)
+	require.NoError(t, conn.QueryRow("SELECT * FROM test_date WHERE ID = $1", 1).Scan(
+		&result1.ColID,
+		&result1.Col1,
+	))
+	require.Equal(t, date.Format("2006-01-02 15:04:05"), result1.Col1.Format("2006-01-02 15:04:05"))
```
ernado, line 48:
```go
const dateLayout = "2006-01-02"
require.Equal(t, date.Format(dateLayout), result1.Col1.Format(dateLayout))
```

ernado: The Date type has no time in it, so you should compare only date

ernado (approved):
LGTM if time format is adjusted

Issue #676, opened 2022-07-20
Respect ReadTimeout for native connections


gingerwizard:
ReadTimeout can be set, and it's respected by HTTP connections (it's used for ResponseHeaderTimeout). For native, the user has to pass a deadline through the context. We should consider respecting the ReadTimeout at a connection level i.e. set to now().Add(ReadTimeout).

PR #699, opened 2022-07-27
Respect ReadTimeout in native

gingerwizard:
We now respect `ReadTimeout` in native.

Closes https://github.com/ClickHouse/clickhouse-go/issues/676

Note we align the ReadTimeout with the[ ClickHouse default of 300s](https://clickhouse.com/docs/en/operations/settings/settings/#connect-timeout-receive-timeout-send-timeout).

We should increase dial timeout to 10s as well.

This could be considered a breaking change - since current behaviour is no timeout. I think this is risky however - and could be considered poor implementation.

### DIFF
--- a/clickhouse_options.go
+++ b/clickhouse_options.go
@@ -264,7 +264,7 @@ func (o Options) setDefaults() *Options {
 		o.DialTimeout = time.Second
 	}
 	if o.ReadTimeout == 0 {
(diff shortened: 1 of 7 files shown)


genzgd (approved):

ernado:
Agree. Probably I need to review ch-go defaults, thanks

PR #707, opened 2022-07-31
string to date/datetime and string type assertions

DGuang21:
contains improvements
- string to date
- string to datetime 
- string type assertions

### DIFF
--- a/lib/column/date.go
+++ b/lib/column/date.go
@@ -205,7 +205,7 @@ func (col *Date) AppendRow(v interface{}) error {
 			col.col.Append(datetime)
 		}
 	default:
-		s, ok := v.(iString)
+		s, ok := v.(fmt.Stringer)
 		if ok {
 			return col.AppendRow(s.String())
 		}
@@ -220,7 +220,9 @@ func (col *Date) AppendRow(v interface{}) error {
 
 func (col *Date) parseDate(str string) (datetime time.Time, err error) {
 	defer func() {
-		err = dateOverflow(minDate, maxDate, datetime, defaultDateFormat)
+		if err == nil {
+			err = dateOverflow(minDate, maxDate, datetime, defaultDateFormat)
+		}
 	}()
 	datetime, err = time.Parse(defaultDateFormat, str)
 	if err == nil {
--- a/lib/column/date.go
+++ b/lib/column/date.go
@@ -30,6 +30,10 @@ var (
 	maxDate, _ = time.Parse("2006-01-02 15:04:05", "2106-01-01 00:00:00")
 )
 
+const (
+	defaultDateFormat = "2006-01-02"
+)
+
 type Date struct {
 	col  proto.ColDate
 	name string
@@ -86,7 +90,7 @@ func (col *Date) Append(v interface{}) (nulls []uint8, err error) {
 	switch v := v.(type) {
 	case []time.Time:
 		for _, t := range v {
-			if err := dateOverflow(minDate, maxDate, t, "2006-01-02"); err != nil {
+			if err := dateOverflow(minDate, maxDate, t, defaultDateFormat); err != nil {
 				return nil, err
 			}
 			col.col.Append(t)
@@ -96,7 +100,7 @@ func (col *Date) Append(v interface{}) (nulls []uint8, err error) {
 		for i, v := range v {
 			switch {
 			case v != nil:
-				if err := dateOverflow(minDate, maxDate, *v, "2006-01-02"); err != nil {
+				if err := dateOverflow(minDate, maxDate, *v, defaultDateFormat); err != nil {
 					return nil, err
 				}
 				col.col.Append(*v)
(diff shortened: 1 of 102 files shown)


DGuang21:
https://github.com/ClickHouse/clickhouse-go/issues/574

DGuang21:
I have done all the modifications on my computer and confirmed all the test cases involved, please re-run ci and re-review the code, thx
@ernado

ernado:
This looks good for me in general, but someone probably should take a look too. 

Currently too busy to review in in-depth.

ernado (approved):

ernado:
cc @dwalters, I think we can merge it

lib/column/date.go
```diff
@@ -171,6 +218,19 @@ func (col *Date) AppendRow(v interface{}) error {
...
+	defer func() {
+		if err == nil {
+			err = dateOverflow(minDate, maxDate, datetime, defaultDateFormat)
+		}
+	}()
+	datetime, err = time.Parse(defaultDateFormat, str)
+	if err == nil {
+		return datetime, nil
+	}
+	return time.Parse(defaultDateTimeFormat, str)
```
gingerwizard, line 231: any reason we need to call this twice? can't we just return the err and a time.Time{}?

lib/column/date32.go
```diff
@@ -171,6 +214,19 @@ func (col *Date32) AppendRow(v interface{}) error {
...
+	defer func() {
+		if err == nil {
+			err = dateOverflow(minDate32, maxDate32, datetime, defaultDateFormat)
+		}
+	}()
+	datetime, err = time.Parse(defaultDateFormat, str)
+	if err == nil {
+		return datetime, nil
+	}
+	return time.Parse(defaultDateTimeFormat, str)
```
gingerwizard, line 227: same comment as above i.e.

any reason we need to call this twice? can't we just return the err and a time.Time{}?

lib/column/datetime.go
```diff
@@ -203,4 +251,13 @@ func (col *DateTime) row(i int) time.Time {
 	return v
 }
 
+func (col *DateTime) parseDateTime(str string) (datetime time.Time, err error) {
+	defer func() {
+		if err == nil {
+			err = dateOverflow(minDateTime, maxDateTime, datetime, defaultDateFormat)
+		}
+	}()
+	return time.Parse(defaultDateTimeFormat, str)
```
gingerwizard, line 260: we will probably need to handle tz here long term but for now fine

gingerwizard:
@DGuang21 A few minor comments re style and why we call `time.Parse` twice. I see you don't support Scan time - I think this is fine and not necessary.

There is some work pending here around ensuring we are consistent with timezones - I'll raise a separate issue.

Finally, be aware of we may change error handling per https://github.com/ClickHouse/clickhouse-go/issues

DGuang21:
@gingerwizard 
Thank you very much for your comments and supportive suggestions.
- Why parse the time twice
The formats of the two parsing times are defaultDateFormat and defaultDateTimeFormat, respectively, and the corresponding time formats are 2006-01-02 15:04:05 and 2006-01-02. What I'm thinking is that parsing with 2006-01-02 throws an error if the full standard time is specified but only the date is inserted. If you think this is unnecessary compatibility, I can remove it in the code

gingerwizard:
I didn't see the use of the different formats in each case! My only hesitancy here is there is a loss of precision in the case the users pass a DateTime string to a date field. We might parse it, but the time component will ultimately be lost. We've tried to avoid these transparent losses of precisions in v2 (v1 had these in several places). I think i would prefer we just be strict.
Also apologies for delayed reviews - summer holidays here.

DGuang21:
@gingerwizard 
Apologizing is unnecessary, this is a normal and reasonable thing~
It is true that you lose precision when using DateTime String to convert to date type. I will remove the redundant judgment as soon as possible and leave the complex judgment to the user, not us. (But I personally feel that the loss of precision like this should be the responsibility of the user, unless we also process additional information such as time zone in time parsing)
In addition, regarding the handling of Tz, should we deal with it together now or wait until the issue is released?

gingerwizard:
>  In addition, regarding the handling of Tz, should we deal with it together now or wait until the issue is released?

a separate issue I think. I'll raise very soon with the changes we need to make to be consistent with other ClickHouse clients.

gingerwizard:
@DGuang21 can you push the change we discuss above and I'll merge.

gingerwizard:
re timezone handling - https://github.com/ClickHouse/clickhouse-go/issues/719

DGuang21:
@gingerwizard 
The content of the discussion has been pushed, please re-run ci and review.

Regarding issue https://github.com/ClickHouse/clickhouse-go/issues/719 Can I use regex to parse time rules of multiple string types as originally discussed?

gingerwizard (approved):
LGTM

PR #711, opened 2022-08-03
Added GroupSet type for v1.

f0mster:
Added GroupSet type as it works on v2.

```
args := []GroupSet{ {'john', 26}, {'kevin', 30}, {'peter', 14}}
ch.Query("select * from users where (name, age) in (?)", args)
```


gingerwizard:
Will merge @f0mster but v1 is bug and security fixes only. We won't release for this PR.

Issue #713, opened 2022-08-08
another UUID support


Laotree:
**Is your feature request related to a problem? Please describe.**

when I define a struct use https://github.com/satori/go.uuid 
there is a column.ColumnConverterError

**Describe the solution you'd like**

https://github.com/satori/go.uuid support 

**Describe alternatives you've considered**

uuid is not a single package, please think about those uuids not in [https://github.com/google/uuid](https://github.com/google/uuid)

**Additional context**
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/16837634/183373879-a9bc987c-75d6-4af2-ac4f-625f1fd940b0.png">

DGuang21:
I don't think go mods should be introduced too much for compatibility with a certain type.

satori.uuid implements the String() method

https://github.com/satori/go.uuid/blob/b2ce2384e17bbe0c6d34077efa39dbab3e09123b/uuid.go#L115

```type iString interface```
Therefore, a judgment should be added at default here, and the value of interface should be asserted as iString.


code e.g
```
if s, ok := v.(iString); ok {
  col.col.Append(s.String())
}
```

DGuang21:
Or I can make iString a public method in https://github.com/ClickHouse/clickhouse-go/pull/707

DGuang21:
> **Is your feature request related to a problem? Please describe.**
> 
> when I define a struct use https://github.com/satori/go.uuid there is a column.ColumnConverterError
> 
> **Describe the solution you'd like**
> 
> https://github.com/satori/go.uuid support
> 
> **Describe alternatives you've considered**
> 
> uuid is not a single package, please think about those uuids not in https://github.com/google/uuid
> 
> **Additional context** Add any other context or screenshots about the feature request here. <img alt="image" width="1440" src="https://user-images.githubusercontent.com/16837634/183373879-a9bc987c-75d6-4af2-ac4f-625f1fd940b0.png">

This interface can also use the ```fmt.Stringer``` provided by the Golang fmt package

gingerwizard:
I see you raised https://github.com/ClickHouse/clickhouse-go/pull/714. Similar sentiments to @DGuang21  - we could support parsing anything which implements fmt.Stringer for the uuid column. This would work but not in column mode.

gingerwizard:
Addressed by https://github.com/ClickHouse/clickhouse-go/pull/720

uuids now need to just implement stringer and scan.

PR #714, opened 2022-08-08
column support github.com/satori/go.uuid

Laotree:
#713


gingerwizard:
Hesistant here - if we approve this we open the door to users adding any package they feel we should handle. This package does have a lot of usage however so I see the value - we are also very opinionated on google uuid (as is ch-go), so I see the argument. Is there any reason you can't just pass as a string though? this package supports `fmt.Stringer` and we could add `fmt.Stringer` to AppendRow  - this would feel more generic.

Laotree:
> Hesistant here - if we approve this we open the door to users adding any package they feel we should handle. This package does have a lot of usage however so I see the value - we are also very opinionated on google uuid (as is ch-go), so I see the argument. Is there any reason you can't just pass as a string though? this package supports `fmt.Stringer` and we could add `fmt.Stringer` to AppendRow - this would feel more generic.

I use this package for a quick convert from string, and also because it's widely used by my colleagues.

it's supported in other orm tool packages such as postgresql or mysql. So, when I got an error in ck-go, I'm confused why it's not jus fine as expected. 

Yes, I can workaround by using google-uuid when connected with ck, and using satori-uuid when come with pg/mysql. However it will be weird to convert uuid when pass uuid between different databases.

If you don't want to add package by users, you could think about how to workaround in a more graceful way. Since it's so confused to get an error when using satori-uuid.

gingerwizard:
> it's supported in other orm tool packages such as postgresql or mysql. So, when I got an error in ck-go, I'm confused why it's not jus fine as expected.

This isn't an orm package. Its a client/driver.

Are you using the `AppendRow` interface? If so, we can add support for `fmt.Stringer` - this will then transparently work. Likewise for `Scan`.

Laotree:
> 

I used AppendStruct and Scan interfaces

Thanks for the comming support.

gingerwizard:
@Laotree see https://github.com/ClickHouse/clickhouse-go/pull/720. Closing in favor of the more generic solution. Included test for satori uuid.

PR #716, opened 2022-08-10
Closes #715 (quoted table name in INSERT INTO)

wizzard0:
Closes #715 by extending the regex with optional backticks.

PR #718, opened 2022-08-11
DDL tests

gingerwizard:


gingerwizard:
Tests will fail, fixed by https://github.com/ClickHouse/clickhouse-go/pull/716 thanks to @wizzard0

Issue #698, opened 2022-07-27
Consider server timezone for Datetimes with Zone


gingerwizard:
Currently, we ensure time.Time{} are assigned the timezone of a DateTime column at retrieval time - if this is specified in column definition e.g. `DateTime('Asia/Shanghai')`. 

In the absence of this information, the tz of a column will revert to the configuration of the server. We don't consider the timezone time, however in this case - although its available in the native protocol. In this case, currently times have the local time.

Proposal is to respect server time.

@ernado

genzgd:
I'd lean toward making this a driver/client setting but then the question becomes what's the default.  Most languages have the concept of a "naive" timezone and I'm inclined to say if you read a ClickHouse DateTime column without a specified timezone you expect the result to be a "naive" epoch based time object (with the freedom to assign whatever timezone you want in the client application).  The server timezone is more decoupled than a column specified time, and it feels a little fragile to use that value to associate with table/column level data.

However I suspect most users are used to working in a single timezone for "all the things"  For most developers it's UTC, and for a lot of Chinese developers it's the single China timezone.  Personally as a developer I always use timestamps/naive objects unless actually displaying the to end user, so adding the UTC timezone automatically feels wrong to me.

Which leads back to a configuration setting -- "use_server_timezone_for_datetime_columns", which I'd default to false.

ernado:
Personally I prefer always using `time.Time` directly, client and server should handle timezones automatically for DateTime, DateTime64.

Internally (on the wire) it is always UTC, TZ is used only for:
- Displaying time
- Parsing time, if no timezone is provided

So, my points:
- Sometimes timezone info is not available at all (no tzinfo on the host/container), we should handle it
- Sometimes we can fail to parse or load server timezone, we should gracefully handle this
- We should encourage using time.Time for DateTime, DateTime64
- We should support proto.DateTime, proto.DateTime64 for those who want operate with raw values
- We should default to `clickhouse-cli` behaviour when timezone is not available in user input

gingerwizard:
We agreed to this. See https://github.com/ClickHouse/clickhouse-go/issues/719

Issue #653, opened 2022-07-08
Rewrite test infrastructure


gingerwizard:
We have quite a comprehensive testing, especially when we consider the testing in ch-go.

Our current tests could be improved, however. Specifically:

- [x] Tests rely on a ClickHouse instance being available. Tests should orchestrate ClickHouse themselves e.g. using https://golang.testcontainers.org/. This would allow us to set settings in the containers and allow people to easily execute tests without needing to run ClickHouse themselves. https://github.com/ClickHouse/clickhouse-go/issues/653
- [x] Our tests are currently very nested with conditionals e.g. [here](https://github.com/ClickHouse/clickhouse-go/blob/v2/tests/std/array_test.go). These are painful to read. Using `require` we can flatten these.  https://github.com/ClickHouse/clickhouse-go/pull/729
- [x] Tests have a lot of boiler plate code. They are repetitive. General functions would allow us to write these more succiently.
- [x] Testing across different versions relies on github actions. Ideally we would be able to run tests, targeting any version of ClickHouse with a single command. https://github.com/ClickHouse/clickhouse-go/pull/729
- [x] Test against ClickHouse Cloud https://github.com/ClickHouse/clickhouse-go/issues/709

PR #729, opened 2022-08-22
Use server time  + rework tests to use orchestrated docker

gingerwizard:
- closes https://github.com/ClickHouse/clickhouse-go/issues/698. This is a breaking change. We will need to document @Felixoid 
- Introduces docker to orchestrate tests - this will later allow us to specify non docker endpoints e.g. ClickHouse Cloud. For now it lets us control tz and other server settings easily. part of https://github.com/ClickHouse/clickhouse-go/issues/653
- Flattens tests per https://github.com/ClickHouse/clickhouse-go/issues/653
- Per https://github.com/ClickHouse/clickhouse-go/issues/653 we can now test a specific version with `CLICKHOUSE_VERSION=x make test`


genzgd (approved):
I might expose a GetTestConnection somewhere that is just something like 
```
GetNativeConnection(clickhouse.Settings{
		"allow_experimental_bigint_types": 1,
	}, nil, &clickhouse.Compression{
	})
```

but this is much nicer without all the nesting

gingerwizard:
> I might expose a GetTestConnection somewhere that is just something like
> 
> ```
> GetNativeConnection(clickhouse.Settings{
> 		"allow_experimental_bigint_types": 1,
> 	}, nil, &clickhouse.Compression{
> 		Method: clickhouse.CompressionLZ4,
> 	})
> ```
> 
> but this is much nicer without all the nesting

This is available [here](https://github.com/ClickHouse/clickhouse-go/pull/729/files#diff-847cd06c1bf6ef597ea96894f8b32ecb4d074ad1bc785b39871e11fb6ab71b3dR172).

PR #732, opened 2022-08-24
Support external ClickHouse for Tests

gingerwizard:
- Allows tests to be run against an external ClickHouse instance via `CLICKHOUSE_USE_DOCKER=false`
- Environment variables allow cluster to be specified e.g. CLICKHOUSE_USERNAME, CLICKHOUSE_HOST, CLICKHOUSE_DATABASE, CLICKHOUSE_PASSWORD, CLICKHOUSE_HTTP_PORT, CLICKHOUSE_SSL_PORT, CLICKHOUSE_HTTPS_PORT, CLICKHOUSE_PORT
- Tests can now also be run to use CLICKHOUSE_USE_SSL=true 

By default, tests still use docker 

@Felixoid


gingerwizard:
There is still some factoring to do here, mainly all tests currently check if they should use SSL. I'd like to refactor this away.

PR #735, opened 2022-08-31
Fix error in nil check for *Decimal column Append

maxjustus:
Previously this line was checking that the decimal slice is not nil to determine whether to append the element which will always be true, even if the element being appended is nil. If the element is nil the code will panic. This changes it to check if the element in the slice is not nil before appending the element.


gingerwizard:
Thanks @maxjustus appreciate this, will merge once tests pass

PR #774, opened 2022-10-04
Handle double pointers for Nullable columns when batch inserting

mdonkers:
Resolves https://github.com/ClickHouse/clickhouse-go/issues/777

When batch inserting `Nullable` type, check if the type is a double pointer and if so, 'unwrap' to just a single pointer.

Made a small proposal to force IPv4 usage when using Docker for tests, as at least locally I'm running into irregular behaviour. Happy to revert if we don't think this is a good idea...

### DIFF
--- a/tests/resources/custom.xml
+++ b/tests/resources/custom.xml
@@ -1,8 +1,9 @@
 <?xml version="1.0" ?>
 <clickhouse>
     <listen_host>::</listen_host>
-    <listen_host>0.0.0.0</listen_host>
     <listen_try>1</listen_try>
+    <https_port>8443</https_port>
+    <tcp_port_secure>9440</tcp_port_secure>
     <logger>
         <console>1</console>
     </logger>
@@ -30,7 +31,4 @@
         </client>
     </openSSL>
     <display_name>clickhouse</display_name>
-    <listen_host>0.0.0.0</listen_host>
-    <https_port>8443</https_port>
-    <tcp_port_secure>9440</tcp_port_secure>
 </clickhouse>
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -155,9 +155,12 @@ func CreateClickHouseTestEnvironment(testSet string) (ClickHouseTestEnvironment,
 		Image:        fmt.Sprintf("clickhouse/clickhouse-server:%s", GetClickHouseTestVersion()),
 		Name:         fmt.Sprintf("clickhouse-go-%s-%d", strings.ToLower(testSet), time.Now().UnixNano()),
 		ExposedPorts: []string{"9000/tcp", "8123/tcp", "9440/tcp", "8443/tcp"},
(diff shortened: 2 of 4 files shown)


tests/resources/custom.xml
```diff
@@ -1,8 +1,9 @@
 <?xml version="1.0" ?>
 <clickhouse>
     <listen_host>::</listen_host>
-    <listen_host>0.0.0.0</listen_host>
```
mdonkers, line 4: https://clickhouse.com/docs/en/operations/server-configuration-parameters/settings/#server_configuration_parameters-listen_host

`<listen_host>::</listen_host>` already means to listen on all hosts. No need to specify `0.0.0.0` as well (and is double specified).
Moved port definitions closer to listen host for easier spotting.

tests/utils.go
```diff
@@ -155,9 +155,12 @@ func CreateClickHouseTestEnvironment(testSet string) (ClickHouseTestEnvironment,
 		Image:        fmt.Sprintf("clickhouse/clickhouse-server:%s", GetClickHouseTestVersion()),
 		Name:         fmt.Sprintf("clickhouse-go-%s-%d", strings.ToLower(testSet), time.Now().UnixNano()),
 		ExposedPorts: []string{"9000/tcp", "8123/tcp", "9440/tcp", "8443/tcp"},
-		WaitingFor: wait.ForAll(wait.ForLog("Ready for connections").WithStartupTimeout(time.Second*time.Duration(120)), wait.ForSQL("9000/tcp", "clickhouse", func(port nat.Port) string {
-			return fmt.Sprintf("clickhouse://default:ClickHouse@localhost:%s", port.Port())
-		})).WithStartupTimeout(time.Second * time.Duration(120)),
+		WaitingFor: wait.ForAll(
+			wait.ForLog("Ready for connections").WithStartupTimeout(time.Second*time.Duration(120)),
+			wait.ForSQL("9000/tcp", "clickhouse", func(port nat.Port) string {
+				return fmt.Sprintf("clickhouse://default:ClickHouse@127.0.0.1:%s", port.Port())
```
mdonkers, line 161: Changed `localhost` to `127.0.0.1` explicitly, because on IPv6 enabled systems this might be pointing to the IPv6 loopback address `::1`:

```
$ cat /etc/hosts
127.0.0.1        localhost
127.0.1.1        housepaper
::1              localhost ip6-localhost ip6-loopback
ff02::1          ip6-allnodes
ff02::2          ip6-allrouters
```

In that scenario, startup of the container might timeout and ClickHouse only shows very strange error messages.

When debugging, it can be seen that the golang `net.Dialer` prefers the IPv6 address above the IPv4 one, though with the incorrect port. Using `127.0.0.1` makes it explicit IPv4.

image

image

mdonkers: And I think it's fine for now to use IPv4 explicitly.

I'll see with the Testcontainers project if its possible to create an issue / PR to solve this, specifying whether you want the IPv4 or IPv6 mapped port.

mdonkers: https://github.com/testcontainers/testcontainers-go/issues/551

Issue #733, opened 2022-08-25
clickhouse.stdDriver implement the sql/driver/ConnBeginTx interface


albertlockett:
**Is your feature request related to a problem? Please describe.**
I'm trying to use the otelsql instrumentation and running into a problem when I begin a transaction.

```golang
// main.go
conn = otelsql.OpenDB(clickhouse.Connector(opts))
tx, err := conn.Begin()
```

otelsql's driver.Conn implementation does implement ConnBeginTx, so golang will call it's BeginTx method
https://github.com/golang/go/blob/master/src/database/sql/ctxutil.go#L98

otelsql expects it's connection to also implement the ConnBeginTx interface, so this will return an err
https://github.com/XSAM/otelsql/blob/main/conn.go#L170

**Describe the solution you'd like**
Currently stdDriver only implements driver.Conn https://pkg.go.dev/database/sql/driver#Conn e.g. It exposes a BeginTx method, which is allegedly deprecated:
```golang
type Conn interface {
       ...
        // Begin starts and returns a new transaction.
	//
	// Deprecated: Drivers should implement ConnBeginTx instead (or additionally).
	Begin() ([Tx](https://pkg.go.dev/database/sql/driver#Tx), [error](https://pkg.go.dev/builtin#error))
```

We could add an additional BeginTx method to the struct so it will implement ConnBeginTx
https://pkg.go.dev/database/sql/driver#ConnBeginTx

According to the interface definition, we'll need to check the txn options argument and return some errors if ReadOnly and some Isolation levels aren't supported. Here's how the standard library does these checks, so we can do something similar.
https://github.com/golang/go/blob/master/src/database/sql/ctxutil.go#L110

i.e. add something like this:
```golang
func (std *stdDriver) BeginTx(ctx context.Context, opts TxOptions) (Tx, error) {
  	if opts != nil {
		// Check the transaction level. If the transaction level is non-default
		// then return an error here as the BeginTx driver value is not supported.
		if opts.Isolation != LevelDefault {
			return nil, errors.New("sql: driver does not support non-default isolation level")
		}

		// If a read-only transaction is requested return an error as the
		// BeginTx driver value is not supported.
		if opts.ReadOnly {
			return nil, errors.New("sql: driver does not support read-only transactions")
		}
	}

	if ctx.Done() == nil {
		return std.Begin()
	}
}
```

**Describe alternatives you've considered**
[...]

PR #779, opened 2022-10-07
#733 driver implemets the sql/driver/ConnBeginTx interface

albertlockett:
https://github.com/ClickHouse/clickhouse-go/issues/733

### DIFF
--- a/clickhouse_std.go
+++ b/clickhouse_std.go
@@ -160,6 +160,22 @@ func (std *stdDriver) Ping(ctx context.Context) error { return std.conn.ping(ctx
 
 func (std *stdDriver) Begin() (driver.Tx, error) { return std, nil }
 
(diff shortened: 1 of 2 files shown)


jkaflik:
Hi @albertlockett 

Thanks for submitting this PR.
Did you check if `otelsql` is able to support SQL drivers that do not implement `ConnBeginTx` interface? From our point of view, it makes more sense to keep our API away from fake doors.

albertlockett:
We could experiment by removing the `Begin` function. I'm not sure if there are other reasons to keep it around for backwards compatibility? 

Also, it looks like `otelsql` will begin supporting the `driver.Conn` interface if/when this PR merges:
https://github.com/XSAM/otelsql/pull/153

PR #781, opened 2022-10-12
fix timeZone() not working on versions below 21.4

### LINKED ISSUE #780: use function timezone instead of timeZone to readTimeZone in http-conn dialHttp func
### Issue description
for clickhouse with version lower than 21.4, the function timeZone does not supported, while for all clickhouse versions, function timezone could be recognized 

### Example code
```go
[If possible, please enter some example code here to reproduce the issue.](https://github.com/ClickHouse/clickhouse-go/blob/main/conn_http.go#L242)

replace SELECT timeZone() with SELECT timezone()
```

### Error log
image

### Configuration
*OS:*

*Interface:* E.g. `native`, `database/sql`
database/sql

*Driver version:*
2.3.0

*Go version:* run `go version` in your console
1.19

*ClickHouse Server version:*
21.3.10.1

derN3rd:
CH Server below 21.4 does not understand the timeZone() function.
timezone() works fine through all versions therefore it should be used instead.

Fixes #780

### DIFF
--- a/conn_http.go
+++ b/conn_http.go
@@ -239,7 +239,7 @@ func (h *httpConnect) isBad() bool {
 }
 
 func (h *httpConnect) readTimeZone(ctx context.Context) (*time.Location, error) {
-	rows, err := h.query(ctx, func(*connect, error) {}, "SELECT timeZone()")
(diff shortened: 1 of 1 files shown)


derN3rd:
@gingerwizard I would love to sign the CLA, but the website shows no Version to sign

PR #794, opened 2022-10-22
feat: support for map[string]interface{} / []map[string]interface{} and some types can be converted to int64/uint64 with no loss of precision

DGuang21:
To some extent, it is supported that some types can be converted to int64/uint64 with no loss of precision, and that reads and writes to incoming map[string]interface / []map[string]interface are supported.

The code may still need to be optimized, and this is only the first version


gingerwizard:
@DGuang21 not comfortable with the warnings although I can potentially see they might not occur.

I think you only want to support this for the uint64 and in64 types not all of them - for this you'll need to have conditional check in the gen code.

DGuang21:
> @DGuang21 not comfortable with the warnings although I can potentially see they might not occur.
> 
> I think you only want to support this for the uint64 and in64 types not all of them - for this you'll need to have conditional check in the gen code.

Regarding the code warning, I will fix it after the discussion, don't worry

Because there are only simple modifications here, my purpose is to support the writing and reading of map[string]interface. At present, the uint/int/string conversion is modified so that, for example, when the database type is map[string]int64, I can define map[string]interface in Golang to write instead of having to specify map[string]int64. This seems a little unfriendly.

Of course, the commonly used types are more than this. Whether it can be modified in this way, and which types need to be compatible, can only be developed after your approval. This can currently be understood as a demo.

The reason for this modification is to define and generate related model / dao files more conveniently in the ORM scaffolding without having to generate a specific type.

gingerwizard:
Ok @DGuang21 i think i see what you're trying to achieve. Currently maps and slices must be strongly typed on insert and you want flexibility. Makes sense. If you want to rework this to avoid errors i'll review properly and we can move forward with the proposal.

jkaflik:
@DGuang21 can you have a look at the failing tests?

DGuang21:
I'm very sorry, it's the end of the year now, and many employees of Chinese companies need to deal with their own promotions at the end of the year, and I am no exception.

I expect that I will be able to complete the remaining Fail CI Fix and logical supplements of this PR in January.

But the general logic and processing method should not change much, so if you have time, you can do a review first, then I can complete the fixes and code changes more quickly. grateful.

@jkaflik @gingerwizard

lib/column/column_gen.go
```diff
@@ -598,37 +598,15 @@ func (col *Int8) Append(v interface{}) (nulls []uint8, err error) {
 }
 
 func (col *Int8) AppendRow(v interface{}) error {
-	switch v := v.(type) {
```
jkaflik, line 601: This file is code generated:
```
// Code generated by make codegen DO NOT EDIT.
// source: lib/column/codegen/column.tpl
```

lib/column/map.go
```diff
@@ -54,6 +54,7 @@ func (col *Map) Name() string {
 func (col *Map) parse(t Type, tz *time.Location) (_ Interface, err error) {
 	col.chType = t
 	if types := strings.SplitN(t.params(), ",", 2); len(types) == 2 {
+		fmt.Println(types)
```
jkaflik, line 57: please remove unnecessary logging

lib/column/map.go
```diff
@@ -93,6 +94,15 @@ func (col *Map) ScanRow(dest interface{}, i int) error {
 		value.Set(col.row(i))
 		return nil
 	}
+	if value.Kind() == reflect.Map && col.row(i).Kind() == reflect.Map {
```
jkaflik, line 97: can we have a function extract with name explaining the logic?

lib/column/tuple.go
```diff
@@ -203,6 +203,15 @@ func setJSONFieldValue(field reflect.Value, value reflect.Value) error {
 		field.Set(value.Convert(field.Type()))
 		return nil
 	}
+	if value.Kind() == reflect.Map && field.Kind() == reflect.Map {
```
jkaflik, line 206: this code seems to be duplicated

jkaflik (changes requested):

PR #802, opened 2022-10-28
Support inserting Enums as int8/int16/int values

mdonkers:
`Enum` values in ClickHouse are always stored as numeric (`Int8` or `Int16`) values. While normally they will be retrieved as `String` values in Golang, in queries its perfectly fine to insert them as the numeric value.

Customers might have their reasons to insert as numeric value instead of the string value. Extend the API so batch inserts support this, as currently you would get the error:
```
        	Error:      	Received unexpected error:
        	            	clickhouse [AppendRow]: auth_type clickhouse [AppendRow]: converting int8 to Enum8 is unsupported
```

### DIFF
--- a/lib/column/enum16.go
+++ b/lib/column/enum16.go
@@ -79,6 +79,46 @@ func (col *Enum16) ScanRow(dest interface{}, row int) error {
 
 func (col *Enum16) Append(v interface{}) (nulls []uint8, err error) {
 	switch v := v.(type) {
+	case []int16:
+		nulls = make([]uint8, len(v))
+		for _, elem := range v {
+			if err = col.AppendRow(elem); err != nil {
+				return nil, err
+			}
+		}
+	case []*int16:
+		nulls = make([]uint8, len(v))
+		for i, elem := range v {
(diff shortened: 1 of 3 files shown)


mdonkers:
Next to the type-specific `int8` and `int16` I've also decided to use the generic Golang `int` for convenience.
Because internally the library holds a mapping between integer and string values, and this is checked, even if the `int` value would be out of range the mapping would not exist and still ('gracefully') fail.

gingerwizard (approved):
LGTM thanks @mdonkers

gingerwizard:
Just a thought, wouldn't uint have made more sense here? afaik, enums cant be signed

mdonkers:
> Just a thought, wouldn't uint have made more sense here? afaik, enums cant be signed

@gingerwizard strangely, enums are signed and thus can be negative: https://clickhouse.com/docs/en/sql-reference/data-types/enum/

Thanks for merging!

gingerwizard:
ha surprising and i should of checked docs. Cheers @mdonkers

Issue #755, opened 2022-09-20
Compress insertion data column by column to reduce memory usage


qixiaogang:
**Is your feature request related to a problem? Please describe.**
After upgrading clickhouse-go to v2, find client memory usage increased when inserting data with compression into tables of 150 columns, almost doubled. 
We don't see memory increase when inserting data with compression into tables of 20 columns.

After investigating, we find v2 encode the whole block then compress the encode result, v1 encode and compress column by column, so v2 will cost more memory to store the encode result of whole block.  

v1 compress https://github.com/ClickHouse/clickhouse-go/blob/v1/lib/data/block.go#L220-L235
v2 compress https://github.com/ClickHouse/clickhouse-go/blob/main/conn.go#L169-L185

**Describe the solution you'd like**
v2 implement compression column by column just as v1 do

**Describe alternatives you've considered**

**Additional context**

gingerwizard:
Good spot, agree this needs to be done.

gingerwizard:
should be doable by simply passing a compressor into the Encode method. We can then compress each column as we encode it.

qixiaogang:
In our case, each batch inserts 5K rows and each row has 150 columns, the total uncompressed size is about 7MB. Each client server maintains about 130+ connections to backend clickhouse servers. 
- V2's Encode reserved 7MB: https://github.com/ClickHouse/clickhouse-go/blob/9d4be497ed788fde5677bb39f1ad0a37e9fca7ba/conn.go#L99
- Compress reserved 7MB: https://github.com/ClickHouse/ch-go/blob/main/compress/writer.go#L22-L23
```
func (w *Writer) Compress(m Method, buf []byte) error {
	maxSize := lz4.CompressBlockBound(len(buf))
	w.Data = append(w.Data[:0], make([]byte, maxSize+headerSize)...)
``` 

The memory reserved for Encode and Compress will be (7MB + 7MB)*130 = **1820MB**, which is not a small number.

qixiaogang:
After checking v1 code, v1 has a limitation ([1M](https://github.com/ClickHouse/clickhouse-go/blob/v1/lib/binary/compress_writer_clz4.go#L24-L31)) for raw buffer to be compressed, so it seems don't need to compress column by column. 

And based on the limitation of raw buffer for compress. A quick change of v2's `sendData` implementation, memory usage reduced to 1/3 v.s. original v2 for our case.
```
func (c *connect) sendData(block *proto.Block, name string) error {
[...]

PR #808, opened 2022-11-02
Buffered compression column by column for native protocol

gingerwizard:
closes https://github.com/ClickHouse/clickhouse-go/issues/755

This is probably sufficient for now. Ultimately we should handle this better as part of https://github.com/ClickHouse/clickhouse-go/issues/809


gingerwizard:
Few pending things:

1. We need to decide what size we should flush on. The original issue https://github.com/ClickHouse/clickhouse-go/issues/755 used an MB. This feels too low IMO @genzgd thoughts?
2. Needs docs inc. recommendation on how to reduce memory usage (we should document block buffer wrt to this as well)
3. We need to benchmark and confirm reduced memory + impact on performance. Initial tests suggest this is actually faster.

jkaflik (approved):

PR #811, opened 2022-11-04
Allow optional HTTP headers; For HTTPS connections, pass auth as headers

tonickkozlov:
Some proxies require additional headers to be passed with HTTP requests in order to work correctly.
Additionally, in HTTPS mode IMO it makes sense to pass username and password as headers by default, since the URLs are not incrypted.

### DIFF
--- a/clickhouse_options.go
+++ b/clickhouse_options.go
@@ -129,7 +129,8 @@ type Options struct {
 	MaxIdleConns     int           // default 5
 	ConnMaxLifetime  time.Duration // default 1 hour
 	ConnOpenStrategy ConnOpenStrategy
-	BlockBufferSize  uint8 // default 2 - can be overwritten on query
+	HttpHeaders      map[string]string // set additional headers on HTTP requests
+	BlockBufferSize  uint8             // default 2 - can be overwritten on query
 
 	scheme      string
 	ReadTimeout time.Duration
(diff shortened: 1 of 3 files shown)


gingerwizard:
@tonickkozlov great catch url creds in url. Appreciate. If tests pass I'll merge.

tonickkozlov:
Thanks for taking a look @gingerwizard . 
My comment above is not accurate - URLs are encrypted of course but can be logged by servers etc.
I'll work on getting the tests fixed.

gingerwizard:
@tonickkozlov no need. cloud tests currently wont pass for external contributors. All tests pass.

tonickkozlov:
Thanks @gingerwizard .
I see the tests are failing on master :( That's not expected right? Could it be that Clickhouse Cloud instance doesn't actually respect X-Clickhouse-User and X-Clickhouse-Key headers?

gingerwizard:
So it should work...but yes we seem to have a problem. Sorry been OOTO and may have merged pre-emptively. taking a look now @tonickkozlov

gingerwizard:
This works locally over SSL secured cluster - can also also auth against a cluster with curl i.e.

```
curl -H "X-Clickhouse-User: default" -H "X-Clickhouse-Key: blurb" --data-binary 'SELECT 1' https://awi597fyap.eu-west-1.aws.clickhouse.cloud:8443
```

gingerwizard:
I see the issue. Headers aren't being persisted across calls. Will PR.

gingerwizard:
https://github.com/ClickHouse/clickhouse-go/pull/820 @tonickkozlov

Issue #813, opened 2022-11-13
Type mismatch with Array(Map(LowCardinality(String), String))


Frapschen:
### Issue description
I get an error when insert data into a table which have nest column. it works fine to me if use a prepared statement to insert data, but it will occur an error when I use `Exec`.

### Example code
```go
import (
	"database/sql"
	"log"
	"testing"

	_ "github.com/ClickHouse/clickhouse-go/v2"
	"github.com/stretchr/testify/require"
)

var (
	driverName = "clickhouse"

	createSQL = `
CREATE TABLE IF NOT EXISTS foo (
    IntValue Int64,
    Exemplars Nested (
		Attributes Map(LowCardinality(String), String)
    ) CODEC(ZSTD(1))          
) ENGINE MergeTree()
PARTITION BY IntValue
ORDER BY IntValue
SETTINGS index_granularity=8192, ttl_only_drop_parts = 1;`

	insertSQL = "INSERT INTO foo (IntValue,Exemplars.Attributes) VALUES (?,?)"

	valueArgs = []interface{}{
		int64(14),
		[]map[string]string{
			{"array1_key1": "array1_value2", "array1_key2": "array1_value2"},
		},
	}
)
// work fine
func Test_tx(t *testing.T) {
	client, err := sql.Open(driverName, "tcp://127.0.0.1:9000/default")

	_, err = client.Exec("drop table if exists foo")
	require.NoError(t, err)

	_, err = client.Exec(createSQL)
	require.NoError(t, err)

	tx, err := client.Begin()
	if err != nil {
		log.Fatal(err)
	}
	defer tx.Rollback()
	stmt, err := tx.Prepare(insertSQL)
	if err != nil {
		log.Fatal(err)
	}
	_, err = stmt.Exec(valueArgs...)
	if err != nil {
		log.Fatal(err)
	}
	err = tx.Commit()
	if err != nil {
		log.Fatal(err)
	}
	stmt.Close()

}
// error: Type mismatch in IN or VALUES section. Expected: Array(Map(LowCardinality(String), String)). Got: Map: While executing ValuesBlockInputFormat
func Test_exec(t *testing.T) {
	client, err := sql.Open(driverName, "tcp://127.0.0.1:9000/default")
	_, err = client.Exec("drop table if exists foo")
	require.NoError(t, err)

	_, err = client.Exec(createSQL)
	require.NoError(t, err)

	_, err = client.Exec(insertSQL, valueArgs...)
[...]

PR #814, opened 2022-11-13
Fix format array nest map scene

Frapschen:
resoved: #813

### DIFF
--- a/bind.go
+++ b/bind.go
@@ -272,7 +272,7 @@ func format(tz *time.Location, scale TimeUnit, v interface{}) (string, error) {
 			}
 			values = append(values, val)
 		}
-		return strings.Join(values, ", "), nil
+		return fmt.Sprintf("[%s]", strings.Join(values, ", ")), nil
 	case reflect.Map: // map
 		values := make([]string, 0, len(v.MapKeys()))
 		for _, key := range v.MapKeys() {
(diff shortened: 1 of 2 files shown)


gingerwizard:
I suspect other tests will fail here - see https://clickhouse.com/docs/en/integrations/go/clickhouse-go/clickhouse-api#special-cases

gingerwizard:
@Frapschen this is the purpose of ArraySet, I.e.. 
```
func Test813(t *testing.T) {
	useSSL, err := strconv.ParseBool(clickhouse_tests.GetEnv("CLICKHOUSE_USE_SSL", "false"))
	require.NoError(t, err)
	conn, err := clickhouse_std_tests.GetDSNConnection("issues", clickhouse.Native, useSSL, "false")
	const ddl = `
		CREATE TABLE test_813 (
		  	IntValue Int64,
			Exemplars Nested (
				Attributes Map(LowCardinality(String), String)
			) CODEC(ZSTD(1)) 
		) Engine MergeTree() ORDER BY tuple()
		`
	conn.Exec("DROP TABLE test_813")
	defer func() {
		conn.Exec("DROP TABLE test_813")
	}()
	_, err = conn.Exec(ddl)
	require.NoError(t, err)

	valueArgs := []interface{}{
		int64(14),
		clickhouse.ArraySet{map[string]string{"array1_key1": "array1_value2", "array1_key2": "array1_value2"}},
	}
	_, err = conn.Exec("INSERT INTO test_813 (IntValue, Exemplars.Attributes) VALUES (?,?)", valueArgs...)
	require.NoError(t, err)
}
```

We unfold arrays comma-separated as this is the better default case.

Frapschen:
Ok, I get it now, thanks!

Issue #823, opened 2022-11-21
An element with a value of nil in the map will have an error


ACV-er:
### Issue description
when I insert multiple rows, and have a field is nil in json, maybe get error

### Example code
```go
If possible, please enter some example code here to reproduce the issue.
```

### Error log
```
clickhouse [Encode]:  mismatched len of columns and  various random errors
```

### Configuration
*OS:* linux & macos

*Driver version: v2.20

*Go version: go1.19.1

*ClickHouse Server version: 22.9.3.18

gingerwizard:
Fixed in 2.4.2

PR #824, opened 2022-11-21
Fix json with nil will get error

ACV-er:
resoved: #823


gingerwizard:
Rather than test on the function level - we don't have any other tests like this. Can you do a high level set of inserts/queries which reproduce the issue please? Add to https://github.com/ClickHouse/clickhouse-go/blob/main/tests/json_test.go

ACV-er:
> Rather than test on the function level - we don't have any other tests like this. Can you do a high level set of inserts/queries which reproduce the issue please? Add to https://github.com/ClickHouse/clickhouse-go/blob/main/tests/json_test.go

I don't know all the implementations very well, it's a bit of a hassle, I'll give it a try

ACV-er:
> 


> Rather than test on the function level - we don't have any other tests like this. Can you do a high level set of inserts/queries which reproduce the issue please? Add to https://github.com/ClickHouse/clickhouse-go/blob/main/tests/json_test.go

Is it okay to submit a commit like this?

gingerwizard:
OK post some JSON which recreates your issue @ACV-er - I'll write a test.

gingerwizard:
@ACV-er our cloud CI tests wont pass for external contributors as you don't have access to cluster secrets. Needs to be fixed. I'll test locally today and merge. thanks again. Will make 2.4.2 - likely end of week.

PR #833, opened 2022-11-29
concurrency on batch fix

### LINKED ISSUE #798: Flush batch cause panic
### Issue description
Using Flush() method in clickhouse.Batch causes panic

### Error log
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x895d0b]

goroutine 12 [running]:
github.com/ClickHouse/ch-go/proto.(*Buffer).PutUInt8(...)
	/go/pkg/mod/github.com/!click!house/ch-go@v0.47.3/proto/buffer.go:97
github.com/ClickHouse/ch-go/proto.(*Buffer).PutByte(...)
	/go/pkg/mod/github.com/!click!house/ch-go@v0.47.3/proto/buffer.go:82
github.com/ClickHouse/clickhouse-go/v2.(*connect).sendData(0xc000146270, 0xc00011e1c0?, {0x0, 0x0})
	/go/pkg/mod/github.com/!click!house/clickhouse-go/v2@v2.3.0/conn.go:171 +0xcb
github.com/ClickHouse/clickhouse-go/v2.(*batch).Flush(0xc00014e280)
	/go/pkg/mod/github.com/!click!house/clickhouse-go/v2@v2.3.0/conn_batch.go:165 +0x79
```

### Configuration
*OS:* `Alpine Linux 3.16`

*Interface:* `native`

*Go version:* 1.19

*ClickHouse Server version:* 22.9.3.18

gingerwizard:
Close #798 specifically https://github.com/ClickHouse/clickhouse-go/issues/798#issuecomment-1330584467

The issue was originally detected due to the use of a batch in a go routine and `Abort` being called after the connection was released on the batch. This would invalidate the connection which had been subsequently reassigned.

This issue could occur as soon as the conn is released (this can happen in a number of places e.g. after `Send` or an `Append` error), and it potentially returns to the pool for use in another go routine. Subsequent releases could then occur e.g., the user calls `Abort` mainly but also `Send` would do it. The result is the connection being closed in the `release` function while another batch or query potentially used it.

This adds a guard - it prevents release being called more than once on a batch. It assumes that batches are not thread safe - they aren't (only connections are).

### DIFF
--- a/conn_batch.go
+++ b/conn_batch.go
@@ -35,6 +35,11 @@ var splitInsertRe = regexp.MustCompile(`(?i)\sVALUES\s*\(`)
 var columnMatch = regexp.MustCompile(`.*\((?P<Columns>.+)\)$`)
 
 func (c *connect) prepareBatch(ctx context.Context, query string, release func(*connect, error)) (driver.Batch, error) {
+	//defer func() {
+	//	if err := recover(); err != nil {
+	//		fmt.Printf("panic occurred on %d:\n", c.num)
+	//	}
+	//}()
 	query = splitInsertRe.Split(query, -1)[0]
 	colMatch := columnMatch.FindStringSubmatch(query)
 	var columns []string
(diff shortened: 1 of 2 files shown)


moming00 (approved):
good job

PR #835, opened 2022-11-29
readme: document `read_timeout` option

rolinh:
The DSN accepts a `read_timeout` option; ensure it is properly documented.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,7 @@ conn.SetConnMaxLifetime(time.Hour)
   - `br` - `0` (Best Speed) to `11` (Best Compression)
   - `zstd`, `lz4` - ignored
 * block_buffer_size - size of block buffer (default 2)
+* read_timeout - a duration string is a possibly signed sequence of decimal numbers, each with optional fraction and a unit suffix such as "300ms", "1s". Valid time units are "ms", "s", "m" (default 5m).
 
 SSL/TLS parameters:
 


jkaflik:
@rolinh thanks for your contribution. please make sure you sign for a CLA.

I'd suggest adding a reference to https://pkg.go.dev/time#ParseDuration `time.ParseDuration` function, so it gives a user a better understanding of what is expected.

README.md
```diff
@@ -140,6 +140,7 @@ conn.SetConnMaxLifetime(time.Hour)
   - `br` - `0` (Best Speed) to `11` (Best Compression)
   - `zstd`, `lz4` - ignored
 * block_buffer_size - size of block buffer (default 2)
+* read_timeout - a duration string is a possibly signed sequence of decimal numbers, each with optional fraction and a unit suffix such as "300ms", "1s". Valid time units are "ms", "s", "m" (default 5m).
```
jkaflik, line 143: I'd suggest adding a reference to `time.ParseDuration` function (https://pkg.go.dev/time#ParseDuration) so the user is aware of how we parse the string.

rolinh: I have simply adapted the text from `dial_timeout` but I'm happy to add a reference to ParseDuration.

jkaflik: @rolinh let's leave it as it is. I will follow up on this later. Sorry for the confusion.

PR #837, opened 2022-12-01
Read only user integration tests

### LINKED ISSUE #736: Test client works for readonly users

jkaflik:
Draft a proposal to make it possible to establish a read-only user connection in tests. It can be used to cover further use cases.

Connection initialization may require additional attention in the future if it grows more.

Resolves #736

### DIFF
--- a/tests/conn_test.go
+++ b/tests/conn_test.go
@@ -242,33 +242,3 @@ func TestBlockBufferSize(t *testing.T) {
 	}
 	require.Equal(t, 10000000, i)
 }
-
-func TestReadOnlyUserConnInsert(t *testing.T) {
-	conn, err := GetNativeConnection(nil, nil, &clickhouse.Compression{
-		Method: clickhouse.CompressionLZ4,
-	})
-	ctx := context.Background()
-	require.NoError(t, err)
-	if !CheckMinServerServerVersion(conn, 21, 9, 0) {
-		t.Skip(fmt.Errorf("unsupported clickhouse version"))
-		return
-	}
-	const ddl = `
-		CREATE TABLE test_readonly_user (
-			  Col1 UInt8
-		) Engine MergeTree() ORDER BY tuple()
-	`
-	defer func() {
-		conn.Exec(ctx, "DROP TABLE test_readonly_user")
-	}()
-
-	require.NoError(t, conn.Exec(ctx, ddl))
-
-	roConn, err := GetReadOnlyNativeConnection(nil, nil, &clickhouse.Compression{
-		Method: clickhouse.CompressionLZ4,
-	})
-
-	actualErr := roConn.Exec(ctx, "INSERT INTO test_readonly_user VALUES (0)")
-
-	assert.EqualError(t, actualErr, "code: 164, message: readonly: Cannot execute query in readonly mode")
-}
--- a/tests/conn_test.go
+++ b/tests/conn_test.go
@@ -243,7 +243,7 @@ func TestBlockBufferSize(t *testing.T) {
 	require.Equal(t, 10000000, i)
 }
 
-func TestReadOnlyUser(t *testing.T) {
+func TestReadOnlyUserConnInsert(t *testing.T) {
 	conn, err := GetNativeConnection(nil, nil, &clickhouse.Compression{
 		Method: clickhouse.CompressionLZ4,
 	})
--- a/tests/conn_test.go
+++ b/tests/conn_test.go
@@ -242,3 +242,33 @@ func TestBlockBufferSize(t *testing.T) {
 	}
 	require.Equal(t, 10000000, i)
 }
+
+func TestReadOnlyUser(t *testing.T) {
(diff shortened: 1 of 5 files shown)


tests/conn_test.go
```diff
@@ -242,3 +242,33 @@ func TestBlockBufferSize(t *testing.T) {
 	}
 	require.Equal(t, 10000000, i)
 }
+
+func TestReadOnlyUserConnInsert(t *testing.T) {
```
mshustov, line 246: We can add other testing scenarios for read-only users:
- should select data from a table
- shouldn't select data from a table 
- shouldn't create a table
- shouldn't delete a table

see https://github.com/ClickHouse/clickhouse-js/blob/01eab4c819992d9f69d94d0d38b952723d1f203c/__tests__/integration/read_only_user.test.ts

tests/utils.go
```diff
@@ -273,7 +281,7 @@ func GetConnectionWithOptions(options *clickhouse.Options) (driver.Conn, error)
 	return clickhouse.Open(options)
 }
 
-func getConnection(env ClickHouseTestEnvironment, database string, settings clickhouse.Settings, tlsConfig *tls.Config, compression *clickhouse.Compression) (driver.Conn, error) {
+func getConnection(env ClickHouseTestEnvironment, database string, roUser bool, settings clickhouse.Settings, tlsConfig *tls.Config, compression *clickhouse.Compression) (driver.Conn, error) {
```
mshustov, line 284: Nit: we need something more powerful than `bool`. There are 3 modes for `readonly` users https://clickhouse.com/docs/en/operations/settings/permissions-for-queries/#readonly

tests/resources/admin.xml
```diff
@@ -6,11 +6,20 @@
...
+            <readonly>1</readonly>
+        </readonly>
     </profiles>
     <users>
         <default>
             <password>ClickHouse</password>
             <access_management>1</access_management>
         </default>
+        <readonly>
+            <password>ClickHouse</password>
```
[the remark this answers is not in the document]
jkaflik: Yes, totally make sense. Will have a look on that.

tests/conn_test.go
mshustov: @gingerwizard do you have other ideas?

jkaflik: I've added mentioned scenarios with one-two extra.

tests/utils.go
jkaflik: @mshustov definitely. However, I would start with having a read-only user covered. Please take a look at the recent changes.

tests/utils.go
```diff
@@ -245,6 +245,61 @@ func GetExternalTestEnvironment(testSet string) (ClickHouseTestEnvironment, erro
...
+	settings := clickhouse.Settings{}
+
+	if proto.CheckMinVersion(proto.Version{
+		Major: 22,
+		Minor: 8,
+		Patch: 0,
+	}, env.Version) {
+		settings["database_replicated_enforce_synchronous_settings"] = "1"
+	}
+	settings["insert_quorum"], _ = strconv.Atoi(GetEnv("CLICKHOUSE_QUORUM_INSERT", "1"))
```
mshustov, line 296: Nit: can we extract env var parsing to a function to avoid repeating this logic?

tests/read_only_user_test.go
```diff
@@ -0,0 +1,138 @@
...
+
+				assert.Equal(t, expectedValue, uint64(5))
+			},
+		},
+	}
+
+	writeQueryCases := []struct {
+		name             string
+		query            string
+		supportedVersion struct{ major, minor, patch uint64 }
```
mshustov, line 51: Can it be removed?

tests/read_only_user_test.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestReadOnlyUser(t *testing.T) {
+	readQueryCases := []struct {
+		name        string
+		query       string
+		assertRowFn func(row driver.Row)
```
mshustov, line 33: Could you add a test for users who can't select from a table without access granted?

mshustov (approved):

tests/read_only_user_test.go
```diff
@@ -0,0 +1,138 @@
...
+	}{
+		{
+			name:  "select from table",
+			query: "SELECT sum(Col1) FROM test_readonly_user",
```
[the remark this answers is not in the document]
jkaflik: I think asserting the exact value is not a subject of this test, so I would leave it there. We might even assume the table is empty as we want to check only if we have permission to query.

tests/read_only_user_test.go
jkaflik: I will add a dedicated test case for that in the future.

tests/utils.go
jkaflik: I want to refactor `utils.go` in close future.

Issue #845, opened 2022-12-13
Support sql.Scanner for custom string/int types


DarkDrim:
### Issue description
I know that clickhouse client supports sql.Scanner for some types: nullable, uuid, date
Example: https://github.com/ClickHouse/clickhouse-go/blob/main/lib/column/date.go#L71-L84

Can you add the same check to string / int types?
It could be useful to have custom string/int types

### Example code
```go
type CustomStr string

func (s *CustomStr) Scan(src any) error {
	if t, ok := src.(string); ok {
		*s = CustomStr(t)
		return nil
	}
	return fmt.Errorf("cannot scan %T into CustomStr", src)
}
```

### Error log
```
clickhouse [ScanRow]: (custom_str) converting Enum8 to *CustomStr is unsupported
```

For example string column type doesn't support sql.Scanner check:
https://github.com/ClickHouse/clickhouse-go/blob/main/lib/column/string.go#L63-L82

---

I can open PR quickly if this is a reasonable and there are no restrictions

Issue #753, opened 2022-09-15
int type cast A support need


Laotree:
**Is your feature request related to a problem? Please describe.**
```
type A int8

const (
	Zero A = iota
	One
)

type Ch struct {
	A A `ch:"a"`
}

```
when INSERT INTO ch, it will come with 
```
clickhouse [AppendRow]: converting A to Int8 is unsupported
```
**Describe the solution you'd like**

in my perspective, A is indeed int type, it's not a unsupported convert, would u guys support it?

Laotree:
similar to #742

gingerwizard:
converting int to int8 is possible precision loss - which we don't want to conceal. #742 would address though i think and allow you to deal with the explicit conversion

gingerwizard:
We could allow any `sql.Scanner` type to be passed.

Laotree:
> We could allow any `sql.Scanner` type to be passed.

that's good, since type checking has been done both in ck and go compiler.

gingerwizard:
hmm, sql.Scanner is good for read time. For Append, we could:

1.  probably achieve this with reflection.  I'd prefer to avoid since it would avoid potentially be expensive - probably using `reflect.ConvertableTo()`
2. require an interface to be implemented e.g. IntValue() int8. My preferred but a little work for the user.
3. Support `Stringify` and convert to string and back to int - seems suboptimal to me.

Open to other ideas @Laotree

Laotree:
> hmm, sql.Scanner is good for read time. For Append, we could:
> 
> 1. probably achieve this with reflection.  I'd prefer to avoid since it would avoid potentially be expensive - probably using `reflect.ConvertableTo()`
> 2. require an interface to be implemented e.g. IntValue() int8. My preferred but a little work for the user.
> 3. Support `Stringify` and convert to string and back to int - seems suboptimal to me.
> 
> Open to other ideas @Laotree

how about allowing users manually define type in tags? 
as following:
```
type A int8

const (
	Zero A = iota
	One
)

type Ch struct {
	A A `ch:"a, type:Int8"`
}
```
if the type in tags is wrong, ck would report error

Laotree:
> that needs reflection again, costly and no better than (1), which would be a similar cost I suspect (need to test) at no real gain.

sometimes, it's not same name in struct or table
```
type Ch struct {
	A A `ch:"b, type:Int8"`
}
```
in this case, since `ch` needs a reflection, I think u could do type check at same time.

[...]

PR #850, opened 2022-12-13
Support custom types (eg type customString string)

DarkDrim:
Proposal: In each ScanRow() check if destination type is implementing sql.Scanner
This PR will add support of custom types, eg:
```
type myStr string
type myInt int32
type myFloat float64
```
One requirement: implement sql.Scanner interface. 

Resolves https://github.com/ClickHouse/clickhouse-go/issues/845

---

Also I found unreachable code:
[lib/column/array](https://github.com/ClickHouse/clickhouse-go/blob/main/lib/column/array.go#L243-L246)
Contains unreachable code

### DIFF
--- a/examples/clickhouse_api/custom_types.go
+++ b/examples/clickhouse_api/custom_types.go
@@ -51,7 +51,7 @@ func CustomTypes() error {
 	err = conn.Exec(ctx, `
 		CREATE TABLE IF NOT EXISTS example (
 			  Col1 String,
-			  Col2 Enum ('hello'   = 1,  'world' = 2),
+			  Col2 Enum ('hello'   = 1,  'world' = 2)
 		) Engine = Memory
 	`)
 	if err != nil {
--- a/examples/clickhouse_api/custom_types.go
+++ b/examples/clickhouse_api/custom_types.go
@@ -1,0 +1,88 @@
+// Licensed to ClickHouse, Inc. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. ClickHouse, Inc. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package clickhouse_api
+
+import (
+	"context"
+	"fmt"
+)
+
+type customStr string
+
+func (s *customStr) Scan(src any) error {
+	if t, ok := src.(string); ok {
+		*s = customStr(t)
+		return nil
+	}
+	return fmt.Errorf("cannot scan %T into customStr", src)
+}
+
+func (s customStr) String() string {
+	return string(s)
+}
+
+func CustomTypes() error {
+	conn, err := GetNativeConnection(nil, nil, nil)
+	if err != nil {
+		return err
+	}
+	ctx := context.Background()
+	defer func() {
+		conn.Exec(context.Background(), "DROP TABLE example")
+	}()
+	if err := conn.Exec(ctx, `DROP TABLE IF EXISTS example`); err != nil {
+		return err
+	}
+	err = conn.Exec(ctx, `
+		CREATE TABLE IF NOT EXISTS example (
+			  Col1 String,
+			  Col2 Enum ('hello'   = 1,  'world' = 2),
+		) Engine = Memory
+	`)
+	if err != nil {
+		return err
+	}
+
+	batch, err := conn.PrepareBatch(ctx, "INSERT INTO example")
+	if err != nil {
+		return err
+	}
+
+	err = batch.Append(
+		customStr("A"), customStr("hello"),
+	)
+	if err != nil {
+		return err
+	}
+
+	err = batch.Send()
+	if err != nil {
+		return err
+	}
+
+	var (
+		col1 customStr
+		col2 customStr
+	)
+
+	if err = conn.QueryRow(ctx, "SELECT * FROM example").Scan(&col1, &col2); err != nil {
+		return err
+	}
+	fmt.Printf("col1=%v (T=%T), col2=%v (T=%T)\n", col1, col1, col2, col2)
+	return nil
+}
--- a/lib/column/array.go
+++ b/lib/column/array.go
@@ -240,10 +240,6 @@ func (col *Array) scan(sliceType reflect.Type, row int) (reflect.Value, error) {
 		}
 		return subSlice, nil
 	}
-	return reflect.Value{}, &Error{
-		ColumnType: fmt.Sprint(sliceType.Kind()),
-		Err:        fmt.Errorf("column %s - needs a slice or interface{}", col.Name()),
-	}
 }
 
 func (col *Array) scanSlice(sliceType reflect.Type, row int, level int) (reflect.Value, error) {
@@ -352,7 +348,6 @@ func (col *Array) scanSliceOfObjects(sliceType reflect.Type, row int) (reflect.V
 				Err:        fmt.Errorf("column %s - needs a slice of objects or an interface{}", col.Name()),
 			}
 		}
-		return reflect.Value{}, nil
 	}
 	return reflect.Value{}, &Error{
 		ColumnType: fmt.Sprint(sliceType.Kind()),
--- a/lib/column/codegen/column.tpl
+++ b/lib/column/codegen/column.tpl
@@ -381,10 +381,14 @@ func (col *{{ .ChType }}) AppendRow(v interface{}) error {
         col.col.Append(val)
 	{{- end }}
 	default:
-		return &ColumnConverterError{
-			Op:   "AppendRow",
-			To:   "{{ .ChType }}",
-			From: fmt.Sprintf("%T", v),
+		if rv := reflect.ValueOf(v); rv.Kind() == col.ScanType().Kind() && rv.CanConvert(col.ScanType()) {
+			col.col.Append(rv.Convert(col.ScanType()).Interface().({{ .GoType }}))
+		} else {
+			return &ColumnConverterError{
+				Op:   "AppendRow",
+				To:   "{{ .ChType }}",
+				From: fmt.Sprintf("%T", v),
+			}
 		}
 	}
 	return nil
--- a/lib/column/codegen/column.tpl
+++ b/lib/column/codegen/column.tpl
@@ -227,7 +227,7 @@ func (col *{{ .ChType }}) ScanRow(dest interface{}, row int) error {
 	{{- end }}
 	{{- if or (eq .ChType "Int64") (eq .ChType "Int32") (eq .ChType "Int16") (eq .ChType "Float64") }}
 	case *sql.Null{{ .ChType }}:
-		d.Scan(value)
+		return d.Scan(value)
 	{{- end }}
     {{- if eq .ChType "Int8" }}
 	case *bool:
@@ -239,6 +239,9 @@ func (col *{{ .ChType }}) ScanRow(dest interface{}, row int) error {
 		}
     {{- end }}
 	default:
+		if scan, ok := dest.(sql.Scanner); ok {
+			return scan.Scan(value)
+		}
 		return &ColumnConverterError{
 			Op:   "ScanRow",
 			To:   fmt.Sprintf("%T", dest),
--- a/lib/column/enum16.go
+++ b/lib/column/enum16.go
@@ -18,6 +18,7 @@
 package column
 
 import (
+	"database/sql"
 	"fmt"
 	"github.com/ClickHouse/ch-go/proto"
 	"reflect"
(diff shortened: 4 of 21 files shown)


jkaflik:
Hi @DarkDrim 
Thanks for your contribution. Can you provide a set of tests? It's also helps to understand the use case better.

DarkDrim:
> Hi @DarkDrim Thanks for your contribution. Can you provide a set of tests? It's also helps to understand the use case better.

sure, do I need to remove unreachable code in array.go?

jkaflik:
@DarkDrim that's not necessary. Thanks for letting us know

tests/string_test.go
```diff
@@ -66,6 +66,61 @@ func TestSimpleString(t *testing.T) {
...
+	require.NoError(t, batch.AppendStruct(&data{
+		Col1: "A",
+		Col2: "B",
+	}))
+	require.NoError(t, batch.Send())
+
+	var dest data
+	require.NoError(t, conn.QueryRow(ctx, "SELECT * FROM test_string").ScanStruct(&dest))
+	assert.Equal(t, "A", dest.Col1)
+	assert.Equal(t, customStr("B"), dest.Col2)
```
DarkDrim, line 121: Added simple example: now it's possible to use custom string types (as in this example - customStr)
It could be useful when you have some struct with custom string/or other types
To not add some kind of converter that will do cast types.

But it should works with array-type too.
For example when you have:
Array(String)

```
type CustomStrArr []string
```

tests/enum_test.go
```diff
@@ -57,6 +57,47 @@ func TestSimpleEnum(t *testing.T) {
...
+		col2Data,
+		col3Data,
+	))
+	require.NoError(t, batch.Send())
+	var (
+		col1 customStr
+		col2 customStr
+		col3 customStr
+	)
+	require.NoError(t, conn.QueryRow(ctx, "SELECT * FROM test_enum").Scan(&col1, &col2, &col3))
```
DarkDrim, line 95: Example #2
Enum can be used with custom string types

DarkDrim:
I want to add one more example with arrays
Will try to do it soon

lib/column/enum16.go
```diff
@@ -210,10 +214,14 @@ func (col *Enum16) AppendRow(elem interface{}) error {
 	case nil:
 		col.col.Append(0)
 	default:
-		return &ColumnConverterError{
-			Op:   "AppendRow",
-			To:   "Enum16",
-			From: fmt.Sprintf("%T", elem),
+		if s, ok := elem.(fmt.Stringer); ok {
+			return col.AppendRow(s.String())
+		} else {
```
DarkDrim, line 219: Additional check if element implement Stringer - we can use it too

lib/column/enum8.go
```diff
@@ -216,10 +220,14 @@ func (col *Enum8) AppendRow(elem interface{}) error {
 	case nil:
 		col.col.Append(0)
 	default:
-		return &ColumnConverterError{
-			Op:   "AppendRow",
-			To:   "Enum8",
-			From: fmt.Sprintf("%T", elem),
+		if s, ok := elem.(fmt.Stringer); ok {
+			return col.AppendRow(s.String())
+		} else {
```
DarkDrim, line 225: Additional check if element implement Stringer - we can use it too

lib/column/fixed_string.go
```diff
@@ -149,10 +153,14 @@ func (col *FixedString) AppendRow(v interface{}) (err error) {
 			return err
 		}
 	default:
-		return &ColumnConverterError{
-			Op:   "AppendRow",
-			To:   "FixedString",
-			From: fmt.Sprintf("%T", v),
+		if s, ok := v.(fmt.Stringer); ok {
+			return col.AppendRow(s.String())
+		} else {
```
DarkDrim, line 158: Additional check if element implement Stringer - we can use it too

lib/column/tuple.go
```diff
@@ -196,8 +196,10 @@ func setJSONFieldValue(field reflect.Value, value reflect.Value) error {
 
 	// check if our target is a string
 	if field.Kind() == reflect.String {
-		field.Set(reflect.ValueOf(fmt.Sprint(value.Interface())))
-		return nil
+		if v := reflect.ValueOf(fmt.Sprint(value.Interface())); v.Type().AssignableTo(field.Type()) {
+			field.Set(v)
+			return nil
+		}
```
DarkDrim, line 202: Super important:
see test `TestCustomArray`
in case we have custom type instead of string -> field.Set(v) , where field is a custom string type and v is string type - reflect package will panic
So to fix this possible issue firstly we're checking if value can be assignable to field (if not - just go to next condition and trying to convert val to field).

tests/array_test.go
```diff
@@ -66,6 +66,46 @@ func TestSimpleArray(t *testing.T) {
 	require.NoError(t, rows.Err())
 }
 
+type customArr []customStr
+
+func TestCustomArray(t *testing.T) {
+	conn, err := GetNativeConnection(nil, nil, &clickhouse.Compression{
```
DarkDrim, line 72: example of usage # 3 - for column Array(Enum) we can use custom string type
It could be useful

lib/column/tuple.go
DarkDrim: Will prepare go/play in a moment to show this issue

DarkDrim: Additional explanation: https://go.dev/play/p/UrtEjsnZEG0

lib/column/array.go
```diff
@@ -240,10 +240,6 @@ func (col *Array) scan(sliceType reflect.Type, row int) (reflect.Value, error) {
 		}
 		return subSlice, nil
 	}
-	return reflect.Value{}, &Error{
-		ColumnType: fmt.Sprint(sliceType.Kind()),
-		Err:        fmt.Errorf("column %s - needs a slice or interface{}", col.Name()),
-	}
```
DarkDrim, line 246: Removed unreachable code

lib/column/array.go
```diff
@@ -352,7 +348,6 @@ func (col *Array) scanSliceOfObjects(sliceType reflect.Type, row int) (reflect.V
 				Err:        fmt.Errorf("column %s - needs a slice of objects or an interface{}", col.Name()),
 			}
 		}
-		return reflect.Value{}, nil
```
DarkDrim, line 355: Removed unreachable code

jkaflik (approved):
I like the changes. If possible, I'd add one or two test cases for generated columns.

tests/string_test.go
DarkDrim: > yes. would you mind adding a custom array string test case?

Sure, added

DarkDrim:
> I like the changes. If possible, I'd add one or two test cases for generated columns.

Added support for base types
It took a little fiddling, but everything works well
Due to the fact that, for example, int to int64 conversion should fail - I had to add one more extra check

Also added a few tests, will comment them in a few moments

lib/column/codegen/column.tpl
```diff
@@ -378,10 +381,14 @@ func (col *{{ .ChType }}) AppendRow(v interface{}) error {
...
-			To:   "{{ .ChType }}",
-			From: fmt.Sprintf("%T", v),
+		if rv := reflect.ValueOf(v); rv.Kind() == col.ScanType().Kind() && rv.CanConvert(col.ScanType()) {
+			col.col.Append(rv.Convert(col.ScanType()).Interface().({{ .GoType }}))
+		} else {
+			return &ColumnConverterError{
+				Op:   "AppendRow",
+				To:   "{{ .ChType }}",
+				From: fmt.Sprintf("%T", v),
+			}
```
DarkDrim, line 391: Supporting base types:
I had to add additional check
```
rv.Kind() == col.ScanType().Kind()
```

because you have one test (SimpleInt) that should fail (int should not be converted into int64)

so how it works:

when we have type int64 and custom int64 type:
```
type customInt64 int64
```

var c customInt64
reflect.TypeOf(c).Kind() == reflect.Int64

that's why it works well

from documentation:

```
The second property is that the Kind of a reflection object describes the underlying type, not the static type. If a reflection object contains a value of a user-defined integer type, as in

type MyInt int
var x MyInt = 7
v := reflect.ValueOf(x)
the Kind of v is still reflect.Int, even though the static type of x is MyInt, not int. In other words, the Kind cannot discriminate an int from a MyInt even though the Type can.
```

https://go.dev/blog/laws-of-reflection

tests/array_test.go
```diff
@@ -66,6 +66,50 @@ func TestSimpleArray(t *testing.T) {
...
+		Method: clickhouse.CompressionLZ4,
+	})
+	ctx := context.Background()
+	require.NoError(t, err)
+	const ddl = `
+		CREATE TABLE test_array (
+			  Col1 Array(Enum ('hello'   = 1,  'world' = 2)),
+			  Col2 Array(String)
+		) Engine MergeTree() ORDER BY tuple()
```
DarkDrim, line 81: Updated test: added Array(Enum) and Array(String)
to show simple []customStrArray not only enum

tests/base_types_test.go
```diff
@@ -40,6 +50,7 @@ func TestUInt8(t *testing.T) {
 				, Col2 Nullable(UInt8)
 				, Col3 Array(UInt8)
 				, Col4 Array(Nullable(UInt8))
+				, Col5 UInt8
```
DarkDrim, line 53: Added new test with custom uint8 type

tests/float_test.go
```diff
@@ -60,6 +60,59 @@ func TestSimpleFloat(t *testing.T) {
...
+	require.NoError(t, err)
+	if !CheckMinServerServerVersion(conn, 21, 9, 0) {
+		t.Skip(fmt.Errorf("unsupported clickhouse version"))
+		return
+	}
+	const ddl = `
+		CREATE TABLE test_float (
+			  Col1 Float32,
+			  Col2 Float64
+		) Engine MergeTree() ORDER BY tuple()
```
DarkDrim, line 97: Added new test with custom float32 and float64

examples/clickhouse_api/custom_types.go
```diff
@@ -0,0 +1,88 @@
...
+	}()
+	if err := conn.Exec(ctx, `DROP TABLE IF EXISTS example`); err != nil {
+		return err
+	}
+	err = conn.Exec(ctx, `
+		CREATE TABLE IF NOT EXISTS example (
+			  Col1 String,
+			  Col2 Enum ('hello'   = 1,  'world' = 2),
+		) Engine = Memory
+	`)
```
DarkDrim, line 56: Besides tests I added example how to works with custom types (in this example String and Enum)

DarkDrim (commented):
With strings, it turned out much easier, because we can simply check for the fmt.Stringer interface and, if it is defined, use it
For other types, this did not work out, so reflection is used, but this is not the only place with reflection in this CH-client))

so should not be a problem

---

Plus, reflection is used at the very end, when the client would simply return an error
Therefore, there is no problem with performance degradation of base types

examples/clickhouse_api/custom_types.go
jkaflik: Thank you. This is something I wanted to ask for. Will review this today. :)

lib/column/codegen/column.tpl
gingerwizard: nice

jkaflik:
@DarkDrim can you have a look at the failed test? don't care about `integration-tests-cloud`. it fails as run on your fork (you don't have access to secrets)

DarkDrim:
> @DarkDrim can you have a look at the failed test? don't care about `integration-tests-cloud`. it fails as run on your fork (you don't have access to secrets)

Can you re-run CI please

jkaflik:
This solves #753

arianitu:
@jkaflik I'm currently still getting this on an Insert and I am using version 


```
{"level":"fatal","error":"clickhouse [AppendRow]: rating clickhouse [AppendRow]: converting model.Rating to String is unsupported"
```

type Rating string

type Organization struct {
    Rating Rating
}

I am also using https://github.com/go-gorm/clickhouse but I think this error comes from this driver?

Am I doing something wrong?

arianitu:
I've fixed this in a weird way, is this really necessary to make custom types work, what's going on here? 

```
type Rating string

func (p Rating) String() string {
	return string(p)
}
```


since I read the `func (col *String) AppendRow(v any) error {` code and it has this line:

```
default:
		if valuer, ok := v.(driver.Valuer); ok {
			val, err := valuer.Value()
			if err != nil {
				return &ColumnConverterError{
					Op:   "AppendRow",
					To:   "String",
					From: fmt.Sprintf("%T", v),
					Hint: "could not get driver.Valuer value",
				}
			}
			return col.AppendRow(val)
		}

		if s, ok := v.(fmt.Stringer); ok {
			return col.AppendRow(s.String())
		}

		return &ColumnConverterError{
			Op:   "AppendRow",
			To:   "String",
			From: fmt.Sprintf("%T", v),
		}
	}
```

SpencerTorres:
Hello!

It looks like this PR was mainly focused on the scanning, not appending. Your fix appears to be the correct approach based on the presence of this block:
```go
if s, ok := v.(fmt.Stringer); ok {
    return col.AppendRow(s.String())
}
```

This has to do with Go's [type identity](https://go.dev/ref/spec#Type_identity). In the internal `Append` functions the type is switched on the interface as in the example below. In this case the type is not considered a `string`. Since you can't add your own type to this `switch`, the next best option is to use the `fmt.Stringer` interface.

Example ([go playground](https://go.dev/play/p/_xJm843K7l7)):
```go

type Rating string

type Org struct {
	Rating Rating
}

func main() {
	org := Org{
		Rating: "rating!",
	}

	append(org.Rating)
}

func append(v any) {
	switch v.(type) {
	case string:
		fmt.Println("custom type is string")
	case Rating:
		fmt.Println("custom type is Rating")
	default:
		fmt.Println("custom type is not string")
	}
}
```

prints
```
custom type is Rating
```

PR #854, opened 2022-12-16
Query parameters support

### LINKED ISSUE #625: Rewrite parameter substitution
Current parameter substitution in `bind.go` isn't robust and a source of pain. Ideally, we would move to ClickHouse and pass the parameters with the query.

Depends on https://github.com/ClickHouse/ClickHouse/issues/38235

jkaflik:
Add a basic query parameters support as implemented in https://github.com/ClickHouse/ClickHouse/pull/39906
This implementation is basic and does not support type conversion. The user is forced to pass parameter values as a string compatible with [CH syntax literals](https://clickhouse.com/docs/en/sql-reference/syntax#literals). 

The native protocol implementation is not documented (yet). For an idea of how it works you can check:
https://clickhouse.com/docs/en/interfaces/http/#cli-queries-with-parameters
https://clickhouse.com/docs/en/interfaces/cli/#cli-queries-with-parameters

ClickHouse supports the following parameter syntax:
```
{<name>:<data type>}
```

Example SQL:
```sql
SELECT {str:String}, {arr:Array(String)}
```

There are two ways to provide parameters. With query context:
```go
chCtx := clickhouse.Context(ctx, clickhouse.WithParameters(clickhouse.Parameters{
	"num":   "42",
	"str":   "hello",
	"array": "['a', 'b', 'c']",
}))

row := client.QueryRow(chCtx, "SELECT {num:UInt64} v, {str:String} s, {array:Array(String)} a")
```

or as compatible with named param:

```go
row := client.QueryRow(
	ctx,
	"SELECT {num:UInt64}, {str:String}",
	clickhouse.Named("num", "42"),
	clickhouse.Named("str", "hello"),
)
```

resolves #625


conn_query.go
```diff
@@ -99,3 +102,31 @@ func (c *connect) queryRow(ctx context.Context, release func(*connect, error), q
...
+		return query, nil
+	}
+
+	// validate if query contains a {<name>:<data type>} syntax, so it's intentional use of query parameters
+	// parameter values will be loaded from `args ...interface{}` for compatibility
+	if paramsProtocolSupport &&
+		len(args) > 0 &&
+		hasQueryParamsRe.MatchString(query) {
+		options.parameters = make(Parameters, len(args))
+		for _, a := range args {
```
genzgd, line 120: What happens if there is a mismatch here?  Should we raise an err?

jkaflik: what mismatch exactly do you mean?

genzgd: Maybe I'm misunderstanding this, but if the arguments passed aren't a list of driver.NamedValues it just silently fails to add the parameters.  It's a minor point.

jkaflik: I think it's a valid point. I was confused since your comment is referred by GH to another line, not to the type-casting one.

Since the user uses a `{xx:xx}` syntax explicitly we might let them know other arg types are not supported.

PR #855, opened 2022-12-16
Expose DialStrategy function to user for custom connection routing

### LINKED ISSUE #772: Expose acquire function to user for custom connection routing
As I see in the current V2 code, every time before there is an operation, this library runs "ch.acquire" to get an active clickhouse-server. I believe we can gain some performance in latency if we move this process to outside of the queries.

So basically a dedicated goroutine, first adds all servers into a pool, and then monitors the every endpoints and if there is an issue with a server, so it can remove from the pool.

In this case, ch.acquire only gets one of the active servers from the pool and executes the operation to that machine. We are doing a similar operation in our load balanced API gateway and it handles thousands of requests per seconds.

jkaflik:
Expose a way to have a customized way how connections created: (the result of dial function)

```go
opts.DialStrategy = func(ctx context.Context, connID int, opts *clickhouse.Options, dial clickhouse.Dial) (clickhouse.DialResult, error) {
	return dial(ctx, "server.host:5678", opts)
}
```

Default `DialStrategy` has following implementation `clickhouse.DefaultDialStrategy`:
```go
func DefaultDialStrategy(ctx context.Context, connID int, opt *Options, dial Dial) (r DialResult, err error) {
	for i := range opt.Addr {
		var num int
		switch opt.ConnOpenStrategy {
		case ConnOpenInOrder:
			num = i
		case ConnOpenRoundRobin:
			num = (int(connID) + i) % len(opt.Addr)
		}

		if r, err = dial(ctx, opt.Addr[num], opt); err == nil {
			return r, nil
		}
	}

	if err == nil {
		err = ErrAcquireConnNoAddress
	}

	return r, err
}
```

`clickhouse.DialStrategy` function receives a `clickhouse.Dial` function that needs to be triggered by the user if a new connection is needed. It returns a dedicated `DialResult` that wraps over internal and coupled `connect` struct.
It's a hacky solution, as connection and pool management are highly coupled in the library. Until we find a better design in v3, we need to have a trade-off. 

Resolves #772


jkaflik:
Closing since it does not meet any of expectations for now.

PR #860, opened 2022-12-22
Append query options to context instead of overwriting.

aaron276h:
Previously every time `clickhouse.Context(context.Context, QueryOption)` is called multiple times on the same context object, only the query options in the last call would used. For example:

```
ctx := Context(context.Background(), WithQueryID("a"))
ctx = Context(ctx, WithQuotaKey("b"))
```
would not persist the the query id. 

This PR changes the `clickhouse.Context()` logic to first attempt to load the existing `QueryOptions` when called so that query options append instead of overwrite existing ones.

It is possible that this is a breaking change for users that have become dependent on this behavior, however it is unlikely that someone is purposefully depending on it today.

### DIFF
--- a/context_test.go
+++ b/context_test.go
@@ -1,0 +1,28 @@
+package clickhouse
+
+import (
+	"context"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestContext(t *testing.T) {
+	// Call context multiple times making sure query options are persisted across calls.
+	ctx := Context(context.Background(), WithQueryID("a"))
+	ctx = Context(ctx, WithQuotaKey("b"))
+	ctx = Context(ctx, WithSettings(Settings{
+		"c": "d",
+	}))
+
+	opts := queryOptions(ctx)
+	require.Equal(t, "a", opts.queryID)
+	require.Equal(t, "b", opts.quotaKey)
+	require.Equal(t, "d", opts.settings["c"])
+
+	// Call context multiple times with the same query options, making sure the latest is persisted.
+	ctx = Context(context.Background(), WithQueryID("a"))
+	ctx = Context(ctx, WithQueryID("b"))
+
+	opts = queryOptions(ctx)
+	require.Equal(t, "b", opts.queryID)
+}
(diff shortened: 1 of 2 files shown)


context_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestContext(t *testing.T) {
+	// Call context multiple times making sure query options are persisted across calls.
```
jkaflik, line 11: minor point, but we could achieve it in a way:
```go
t.Run("call context multiple times making sure query options are persisted across calls", func(t *testing.T) {
```
and so on... It gives better visibility on test analysis.

jkaflik (approved):
Looks good. There is one minor we could have done.
Please let me know.

context_test.go
aaron276h: Updated, appreciate the quick review.

context_test.go
```diff
@@ -0,0 +1,35 @@
...
+			ctx := Context(context.Background(), WithQueryID("a"))
+			ctx = Context(ctx, WithQuotaKey("b"))
+			ctx = Context(ctx, WithSettings(Settings{
+				"c": "d",
+			}))
+
+			opts := queryOptions(ctx)
+			require.Equal(t, "a", opts.queryID)
+			require.Equal(t, "b", opts.quotaKey)
+			require.Equal(t, "d", opts.settings["c"])
```
jkaflik, line 22: `require` is more powerful than `assert` in testify lib. It not only does an assertion but also fails immediately. 

It's more helpful if you want to bootstrap some test dependencies and fail fast. For pure assertions, I recommend switching to `assert` package as it gives more transparency on all failing assertions.

context_test.go
jkaflik: @aaron276h looks good. Added another minor comment.

context_test.go
aaron276h: Switched over to asserts.

jkaflik:
@aaron276h thanks! it will be included in upcoming v2.5 release

PR #862, opened 2022-12-22
Fully parse secure and skip_verify DSN params

n-oden:
The current code will interpret a DSN with `?secure=false` in its options as a directive to turn TSL _on!_

Obviously re-setting what is the default value is redundant, but deployment systems that template the DSN from e.g. environment variables may need to do this.  (Also, the v1 clickhouse-go library supported this pattern, so this is technically a regression.)

The previous behavior (`?secure` without a value set turns secure on) is preserved to minimize surprise.

Signed-off-by: Nathan J. Mehl <n@oden.io>

### DIFF
--- a/clickhouse_options.go
+++ b/clickhouse_options.go
@@ -21,13 +21,14 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
-	"github.com/ClickHouse/ch-go/compress"
-	"github.com/pkg/errors"
 	"net"
 	"net/url"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


n-oden:
AFAICT the integration test failures are unrelated to this PR?

jkaflik:
> AFAICT the integration test failures are unrelated to this PR?

@n-oden yes. your fork is not able to access secrets. it's fine.

Thanks for this contribution. Really appreciated. I'm going to include it in an upcoming release.

Issue #844, opened 2022-12-12
broken pipe driver issue


piyushdatazip:
### Issue description
```
cause: doRequest: transport failed to send a request to ClickHouse: write tcp 19.17.2.250:49816->10.100.164.134:80: write: broken pipe. 
```
Getting this error while making an insert query to clickhouse, Batch Insert, 10000 object in single batch

*Interface:* E.g.  `database/sql`

*Driver version:* v2.4.3

*Go version:* go version go1.19.1 darwin/arm64

*ClickHouse Server version:* Tried latest as well as 4 month old release

piyushdatazip:
Mentioning maintainers here @jkaflik . Similar issue was reported in https://github.com/ClickHouse/ClickHouse/issues/17446
for clickhouse-jdbc. Facing errors randomly when rows are heavy in size, sometime happens, sometimes not.

piyushdatazip:
also getting similar issues such as `EOF` and `Use of closed network connection`
```
cause: Post "CLICKHOUSEDSN&date_time_input_format=best_effort&default_format=Native&wait_end_of_query=1": EOF.
```

jkaflik:
Hi @piyushdatazip 

What's your estimated block size? You are using the HTTP protocol, right? Do you know if that's reproducible on native?

vnazarenko:
I'm also getting `write: write tcp 127.0.0.1:37686-\u003e127.0.0.1:9000: write: broken pipe`  during batch inserts. I have script to insert into clickhouse in batches, a pretty big amount of data ( hundreds millions ) so this script run for hours, and this error start appear only after several hours.

rtkaratekid:
Just wanted to chime in and say that I also ran across this same error today. Before today I've not had this error before. My insert code looks like this
```go
if err := batch.Send(); err != nil && err != clickhouse.ErrBatchAlreadySent {
	log.Println("Failed inserting batch of flows, retrying: ", err)

	if err := batch.Send(); err != nil {
			lostFlowCount += flowCount
			log.Println("Retry insert failed, continuing to next batch: ", err)
			log.Println("Number of lost flows: ", lostFlowCount)
	}
}
```
The first insert attempt fails with:
`Failed inserting batch of flows, retrying: write tcp <ip>:<send port> -><ip>:9000: write: broken pipe`

The retry fails with:
`Retry insert failed, continuing to next batch: clickhouse: batch has already been sent`

[...]

PR #871, opened 2023-01-13
Handle epipe errors

n-oden:
Presently, if the tcp session to the server closes remotely (e.g. if the clickhouse-server process restarts), we will emit broken pipe errors potentially until we reach ConnMaxLifetime.

Since EPIPE means that the tcp session is dead (ie we received a RST packet from the server), there is no point in attempting to proceed past that point: set the connection as closed.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -26,6 +26,7 @@ import (
 	"log"
 	"net"
 	"os"
+	"syscall"
 	"time"
 
 	"github.com/ClickHouse/ch-go/compress"
@@ -226,6 +227,10 @@ func (c *connect) sendData(block *proto.Block, name string) error {
 		return err
 	}
 	if err := c.flush(); err != nil {
+		if errors.Is(err, syscall.EPIPE) {
+			c.debugf("[send data] pipe is broken, closing connection")
+			c.closed = true
+		}
 		return err
 	}
 	defer func() {


rf:
#844 is most likely related!

n-oden:
@jkaflik Could I trouble you for a look at this?

n-oden:
Reproducing it is in conception simple: create a client that continuously writes to a clickhouse db, and then restart the clickhouse server while the client is writing and observe the infinite flood of broken pipe errors in the logs.

In practice with the integration test framework you have here, I'm not sure how we'd implement that while still keeping test times reasonable and results deterministic: you'd want to stop and start the container while writes were happening and I'm not sure how you'd do that with any precision.  I'll keep poking at it but am _very_ open to suggestions!

jkaflik:
@n-oden I looked briefly at the issue. Can you please let me know the version you use? I assume you are using the `database/sql` interface as your proposal does implement a `driver.ErrBadConn`.

jkaflik:
I agree we should mark the connection as closed. I will look at how we can achieve the same result for `database/sql` driver, but without remapping an error on a native interface level.

n-oden:
@jkaflik we're using v2.5.0, having recently updated from v1.  And correct, we're using the database/sql interface.

jkaflik:
@n-oden while I agree the connection should be marked as closed and the SQL driver on top should receive `driver.ErrBadConn`, I struggle to reproduce the following:

> we will emit broken pipe errors potentially until we reach ConnMaxLifetime.

having CH server restart just before writing to the socket (`conn.go:259`) causes me to get the broken pipe, but later on, I get the connection established back and everything works fine. What happens is I get a few `connection refused` during dial. This tells me that the conn is abandoned and the new one is being established until the server is back to accept conn.

It has the same behavior no matter if we introduce your change or not.

My use case is:
```
conn := clickhouse.OpenDB( ... )
conn.SetMaxOpenConns(1)

for {
  scope, err := conn.Begin()
  stmt, err := scope.Prepare("INSERT....")
  stmt.Exec(...)
  scope.Commit()
}
```

I think there is some difference in how you do that and how you are able to reproduce it.

n-oden:
@jkaflik I've been having the exact same issue trying to reproduce the error in the test harness here, which is frustrating, but we have _definitely_ seen the described behavior on our production systems.  Happy to do a screenshare or post a recording somewhere if that would help!

jkaflik:
@n-oden anything that brings me closer to this issue is welcome. Can you also let me know what/when library functions you call?

n-oden:
drop me a line -- n@oden.io

n-oden:
@jkaflik updated as we discussed!

jkaflik:
@n-oden thanks! as we agreed, merging it and merge in order to mitigate the issue on your side.

Let's continue with #844 since it might have the same root cause.

PR #876, opened 2023-01-23
Client info specification implementation

jkaflik:


clickhouse_options.go
```diff
@@ -256,19 +263,50 @@ func (o *Options) fromDSN(in string) error {
...
+
+			if len(subKey) == 0 || len(val) == 0 {
+				continue
+			}
+
+			o.ClientInfo.Products = append(o.ClientInfo.Products, struct{ Name, Version string }{
+				subKey,
+				val,
+			})
+		case "client_info_meta":
```
mshustov, line 287: Do we need to provide `meta` and `comment` config options? The client can derive them from the env. 
What use cases do you envision?

client_info_test.go
```diff
@@ -0,0 +1,108 @@
...
+import (
+	"fmt"
+	"github.com/stretchr/testify/assert"
+	"runtime"
+	"testing"
+)
+
+func TestClientInfoString(t *testing.T) {
+	expectedClientProduct := fmt.Sprintf("%s/%d.%d.%d", ClientName, ClientVersionMajor, ClientVersionMinor, ClientVersionPatch)
+	expectedDefaultMeta := fmt.Sprintf("lv:go/%s; os:%s", runtime.Version()[2:], runtime.GOOS)
```
mshustov, line 29: The test doesn't cover a case when a user specifies `meta.lv` and `meta.os`, although the runtime values take precedence.

tests/std/bool_test.go
```diff
@@ -34,7 +34,7 @@ func TestStdBool(t *testing.T) {
 	require.NoError(t, err)
 	for name, protocol := range dsns {
 		t.Run(fmt.Sprintf("%s Protocol", name), func(t *testing.T) {
-			if conn, err := GetStdDSNConnection(protocol, useSSL, "false"); assert.NoError(t, err) {
+			if conn, err := GetStdDSNConnection(protocol, useSSL, nil); assert.NoError(t, err) {
```
mshustov, line 37: Is this a relevant change? If not, let's avoid mixing several changes in a single PR.

tests/std/client_info_test.go
```diff
@@ -0,0 +1,71 @@
...
+			_, err = conn.Exec("SYSTEM FLUSH LOGS")
+			require.NoError(t, err)
+
+			var clientName string
+			row = conn.QueryRow("SELECT IF(interface = 2, http_user_agent, client_name) as client_name FROM system.query_log WHERE query_id = {queryId:String}", clickhouse.Named("queryId", queryID))
+			require.NoError(t, row.Err())
+			require.NoError(t, row.Scan(&clientName))
+
+			expectedClientName := fmt.Sprintf(
+				"tests/dev %s/%d.%d.%d (std; lv:go/%s; os:%s; protocol:%s)",
```
mshustov, line 59: Could you add a comment with an example of the resulting string?

clickhouse_options.go
jkaflik: This will be provided in docs.

tests/std/bool_test.go
jkaflik: As mentioned in a internal message, I couldn't avoid it.

mshustov:
how are you going to provide `database/sql` API usage in the client info? `database/sql clickhouse-go/2.5.1...` or `clickhouse-go/2.5.1 (api:database/sql....)` ?

clickhouse_options.go
mshustov: > This will be provided in docs.

Could you elaborate, please?

jkaflik: @mshustov not sure I understood your comment then. Do you mean, if it makes sense to expose an API to change meta or comment?

jkaflik:
> how are you going to provide `database/sql` API usage in the client info? `database/sql clickhouse-go/2.5.1...` or `clickhouse-go/2.5.1 (api:database/sql....)` ?

good idea

clickhouse_options.go
mshustov: >  Do you mean, if it makes sense to expose an API to change meta or comment?

I'm wondering in what cases it might be necessary for a developer to supply custom `meta` and `comment` fields. I would prefer not to introduce an API that is rarely used.

mshustov:
> I think providing it in bracket is a better option.

Will you add it (along with the tests) to the current PR?

jkaflik:
Yes. Please keep in mind this PR is draft.

clickhouse_options.go
jkaflik: I don't know such use cases yet. That is definitely a good point to cut it off for now until we get a clear request.

README.md
```diff
@@ -146,6 +162,7 @@ conn.SetConnMaxLifetime(time.Hour)
 * block_buffer_size - size of block buffer (default 2)
 * read_timeout - a duration string is a possibly signed sequence of decimal numbers, each with optional fraction and a unit suffix such as "300ms", "1s". Valid time units are "ms", "s", "m" (default 5m).
 * max_compression_buffer - max size (bytes) of compression buffer during column by column compression (default 10MiB)
+* client_info_product - optional map of products to be added to client info
```
mshustov, line 165: Should we mention that the order matters?

clickhouse_std.go
```diff
@@ -82,9 +84,14 @@ func (o *stdConnOpener) Connect(ctx context.Context) (_ driver.Conn, err error)
 			}, nil
 		}
 	}
+
 	return nil, err
 }
 
+func (o *stdConnOpener) setClientInfoDefaults() {
+	o.opt.ClientInfo.comment = []string{"database/sql"}
```
mshustov, line 92: `database/sql` or `api:database/sql`? Maybe it would be easier to parse with `key:value` structure?

jkaflik: I treat it more like a flag, that tells there is an additional abstraction layer used. We will have parse capability anyway.

README.md
jkaflik: yes, added.

mshustov (approved):

PR #879, opened 2023-01-25
Better handling for connection errors in the std interface

n-oden:
- check for EOF/EPIPE errors in the major functions of clickhouse_std.go and explicitly return driver.ErrBadConn when we catch them

- add debug logging of most error conditions to the std interface, following the existing pattern for error handling

Signed-off-by: Nathan J. Mehl <n@oden.io>

### DIFF
--- a/clickhouse_std.go
+++ b/clickhouse_std.go
@@ -24,9 +24,12 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"log"
+	"os"
 	"reflect"
 	"strings"
 	"sync/atomic"
+	"syscall"
 
 	"github.com/ClickHouse/clickhouse-go/v2/lib/column"
 	ldriver "github.com/ClickHouse/clickhouse-go/v2/lib/driver"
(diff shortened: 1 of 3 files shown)


n-oden:
@jkaflik I think this is the other piece of the puzzle, at least for the std interface: with this patch in place on our side and debugging enabled, resetting a server node results in one EOF error logged per active connection, no "context cancelled" errors, and an immediate recovery.

clickhouse_std.go
```diff
@@ -89,18 +100,34 @@ func init() {
 	sql.Register("clickhouse", &stdDriver{})
 }
 
+// isFatalError returns true if the error class indicates that the
+// db connection is no longer usable and should be marked bad
+func isFatalError(err error) bool {
```
jkaflik, line 105:
```suggestion
func isConnBrokenError(err error) bool {
```

what do you think?

n-oden: LGTM; updated

PR #880, opened 2023-01-26
Add conn options method

### LINKED ISSUE #877: Options method for Conn
**Is your feature request related to a problem? Please describe.**
I want the ability to get Conn's Options that were passed to `clickhouse.Open`. Sometimes i need some information like Auth.Database to build queries. I can store that Options by myself, but i think it would be nice if i could retrieve that Options from some `Conn` method. 

**Describe the solution you'd like**
A method for `Conn` for retrieving `*Options`.

mayamika:
Add Options method for Conn. Closes #877.

### DIFF
--- a/clickhouse.go
+++ b/clickhouse.go
@@ -24,14 +24,19 @@ import (
 	"sync/atomic"
 	"time"
 
+	_ "time/tzdata"
+
 	"github.com/ClickHouse/clickhouse-go/v2/contributors"
 	"github.com/ClickHouse/clickhouse-go/v2/lib/column"
 	"github.com/ClickHouse/clickhouse-go/v2/lib/driver"
 	"github.com/ClickHouse/clickhouse-go/v2/lib/proto"
-	_ "time/tzdata"
 )
 
-type Conn = driver.Conn
+type Conn interface {
+	driver.Conn
+
+	Options() *Options
+}
 
 type (
 	Progress      = proto.Progress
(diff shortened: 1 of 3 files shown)


mayamika:
@jkaflik, just started implementing #877, but i have some problems:

Firstly, i can't extend driver.Conn with Options method, because Options defined in root package, so i changed clickhouse.Conn interface to be like this: 
```
type Conn interface {
	driver.Conn

	Options() *Options
}
```
I think this breaks compatibility.

Maybe this issue just not worth it, or it can be implemented in v3?

jkaflik:
@mayamika we can extend `driver.Conn` interface

mayamika:
> @mayamika we can extend `driver.Conn` interface

To do this we need to move `clickhouse.Options` to `driver` or separate package to avoid import cycle. But i don't think it's a good solution.

jkaflik:
@mayamika you are right. The current design is not consistent. Indeed change `Open` return type still matches `driver.Conn` interface. (https://go.dev/ref/spec#Interface_types )

The problem is when someone mixes the usage of two interfaces `clickhouse.Conn`, `driver.Conn` in their code. It breaks like in one of the integration tests. (I just run it for this PR - you will see the result)

IMO it's not doable in v2.

mayamika:
So i close this PR? @jkaflik

jkaflik:
Yes.

Issue #883, opened 2023-01-29
wrong placeholder in log


fredgnr:
in https://github.com/ClickHouse/clickhouse-go/blob/b2f9409ba1c7bb239a4f6553a6da347f3f5f1330/conn_send_query.go#L27 
the placeholder for connect.compress is wrong.the type of this field has been modified from bool to  byte,but the placeholder is not changed the same time
I have pulled a request to fix this

PR #884, opened 2023-01-29
correct the wrong placeholder in log output

fredgnr:
#883 
the type of ‘compression' field changed while the placeholder in log not changed

### DIFF
--- a/conn_send_query.go
+++ b/conn_send_query.go
@@ -24,7 +24,7 @@ import (
 // Connection::sendQuery
 // https://github.com/ClickHouse/ClickHouse/blob/master/src/Client/Connection.cpp
 func (c *connect) sendQuery(body string, o *QueryOptions) error {
-	c.debugf("[send query] compression=%d %s", c.compression, body)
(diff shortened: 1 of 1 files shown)


conn_send_query.go
```diff
@@ -24,7 +24,7 @@ import (
 // Connection::sendQuery
 // https://github.com/ClickHouse/ClickHouse/blob/master/src/Client/Connection.cpp
 func (c *connect) sendQuery(body string, o *QueryOptions) error {
-	c.debugf("[send query] compression=%t %s", c.compression, body)
+	c.debugf("[send query] compression=%q %s", c.compression, body)
```
jkaflik, line 27:
```suggestion
	c.debugf("[send query] compression=0x%X %s", c.compression, body)
```
It's minor, but perhaps simple notation works better here

jkaflik (approved):

conn_send_query.go
fredgnr: I think that quoted string like "ztsd" is  clear than hex 0x7A737464, the latter may be confusing when reading log

PR #922, opened 2023-03-02
add to  batchColumn  AppendRow and fix parseDate

pikot:
## Summary
-  fix integration test fail

```
--- FAIL: TestDate32 (0.01s)
    date32_test.go:81: 
                Error Trace:    /clickhouse-go/tests/date32_test.go:81
                Error:          Received unexpected error:
                                clickhouse: dateTime overflow. Col7 must be between 1925-01-01 and 2283-11-11
                Test:           TestDate32
```
error may hapened even if time string parsed without error

-  add to  batchColumn function AppendRow, 
usecase:
I have data in rows and want to use batch column insert. Logic with creating columns arrays, and  iterate  this array in Append, looks like overkill. Enouh iterate my data rows and call AppendRow without creating column arrays.

### DIFF
--- a/lib/column/date.go
+++ b/lib/column/date.go
@@ -36,8 +36,14 @@ const (
 )
 
 type Date struct {
-	col  proto.ColDate
-	name string
+	col      proto.ColDate
+	name     string
+	location *time.Location
+}
+
+func (col *Date) parse(t Type, tz *time.Location) (_ *Date, err error) {
+	col.location = tz
+	return col, nil
 }
 
 func (col *Date) Reset() {
@@ -222,7 +228,11 @@ func (col *Date) AppendRow(v interface{}) error {
 	return nil
 }
 
-func parseDate(value string, minDate time.Time, maxDate time.Time) (tv time.Time, err error) {
+func parseDate(value string, minDate time.Time, maxDate time.Time, location *time.Location) (tv time.Time, err error) {
+	if location == nil {
+		location = time.Local
+	}
+
 	defer func() {
 		if err == nil {
 			err = dateOverflow(minDate, maxDate, tv, defaultDateFormatNoZone)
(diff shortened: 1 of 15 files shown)


lib/column/date.go
```diff
@@ -236,6 +231,10 @@ func parseDate(value string, minDate time.Time, maxDate time.Time) (tv time.Time
 			tv.Year(), tv.Month(), tv.Day(), tv.Hour(), tv.Minute(), tv.Second(), tv.Nanosecond(), time.Local,
 		), nil
 	}
+        if err == nil {
+                err = dateOverflow(minDate, maxDate, tv, defaultDateFormatNoZone)
```
jkaflik, line 235: This line will never be reached. We do return in case err nil. 

We don't have tests for this error case.

jkaflik (changes requested):

jkaflik:
@pikot, thanks for your contribution!

lib/column/date.go
pikot: ups, yes, my fault,  At first glans it looked easier, but I found! )  there is TIME ZONE Problem =)
vars
        minDate32, _ = time.Parse("2006-01-02 15:04:05", "1925-01-01 00:00:00")
        maxDate32, _ = time.Parse("2006-01-02 15:04:05", "2283-11-11 00:00:00")
has UTC time zone

but tv, according to code
```
       if tv, err = time.Parse(defaultDateFormatNoZone, value); err == nil {
                return time.Date(
                        tv.Year(), tv.Month(), tv.Day(), tv.Hour(), tv.Minute(), tv.Second(), tv.Nanosecond(), time.Local,
                ), nil
        }
```
has  time.Local, in my case MSK. 

Аnd date 1925-01-01 00:00:00 +0300 MSK is before date 1925-01-01 00:00:00 +0000 UTC, thats why tests failed


I tried to change Min/Max to
```
        minDate32, _ = time.ParseInLocation("2006-01-02 15:04:05", "1925-01-01 00:00:00", time.Local)
        maxDate32, _ = time.ParseInLocation("2006-01-02 15:04:05", "2283-11-11 00:00:00", time.Local)
```
but I have  problem, in test case - min value now has EET timezone -- 1925-01-01 00:00:00 +0200 EET 
at the same time max value, has good  timezone 2283-11-11 00:00:00 +0300 MSK

I don't know why, but I will find solution )

pikot: I see fix for timezone bug. thanks! 
As I think - ETT timezone appears due to in 1925 MSK timezone doesn't exist.
I will recreate pull request.

Issue #925, opened 2023-03-05
Possible to allow `Auth`.`Database` to be empty?


v4run:
Currently, if the Database is not set, it is set to "default" in the `setDefaults` method. I am facing a scenario where the `default` database may not exist. Is there any possibility to allow `Auth`.`Database` to be empty?

jkaflik:
Hi @v4run 
It's possibly supported by native protocol. Contributions are welcome.

v4run:
I would be happy to work on this

v4run:
Hi @jkaflik, I have created a PR for this. I tested it with using native and http interface. It was working. The tests were also passing. Could you please take a look? 
Thanks 😃

v4run:
#926

PR #926, opened 2023-03-06
Allow `Auth.Database` option to be empty

v4run:
## Summary
If we do not set the `Database` in `Option`, then it is assigned the value `default`. This PR aims to remove that and will keep the value of `Database` as it is.

Issue: #925

### DIFF
--- a/clickhouse_options.go
+++ b/clickhouse_options.go
@@ -319,9 +319,6 @@ func (o *Options) fromDSN(in string) error {
 
 // receive copy of Options, so we don't modify original - so its reusable
 func (o Options) setDefaults() *Options {
-	if len(o.Auth.Database) == 0 {
-		o.Auth.Database = "default"
-	}
 	if len(o.Auth.Username) == 0 {
 		o.Auth.Username = "default"
 	}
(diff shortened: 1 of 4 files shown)


jkaflik:
Hi @v4run 
Thanks for your contribution! Could you please provide a set of integration tests that verifies if everything works fine?
You can add this test into `tests/conn_test.go` and `tests/std/conn_test.go`.

v4run:
Hi @jkaflik, I have added the integrations tests. To test, what I did is drop the `default` database, then try to connect to clickhouse and recreate the `default` database. (Since the issue occurs only in case where we try to connect to a clickhouse with the `default` database). 

I feel like it makes more sense to use a separate test environment for this. But, that's just me 😅 . What do you think?

jkaflik:
@v4run LGTM. Let me stage this for the next release.

v4run:
Hi @jkaflik, I was wondering if there is any update on the status of this pull request. Is there an estimated timeline for when it might be merged? 🙂

jkaflik:
Hi @v4run 
I planned doing that last Friday, but will be released today

PR #927, opened 2023-03-06
Add AppendRow function to BatchColumn 

pikot:
## Summary
- add function to batch column. usecase:
I have data in rows and want to use batch column insert. Logic with creating columns arrays, and iterate this array in Append, looks like overkill. Enough iterate my data rows and call AppendRow without creating column arrays.
- smallfix for date test
## Checklist
+ Unit and integration tests covering the common scenarios were added


jkaflik:
Thanks @pikot . Great work!

Issue #937, opened 2023-03-18
Support customized "url path" in http connection


crisismaple:
We are hosting our Clickhouse clusters behind a nginx gateway and routing request by url path instead of different host, since we are supporting customize http headers via `Options.HttpHeaders` and there is no other entry we could inject the url path before sending request, it would be good to support customized url path from Options.HttpPath while constructing the URL in `dialHttp`:

```go
type Options struct {
//...
	HttpHeaders      map[string]string // set additional headers on HTTP requests
	HttpPath         string            // HTTP request url additional path if any
//...
}

func dialHttp(ctx context.Context, addr string, num int, opt *Options) (*httpConnect, error) {
	if opt.scheme == "" {
		switch opt.Protocol {
		case HTTP:
			opt.scheme = opt.Protocol.String()
			if opt.TLS != nil {
				opt.scheme = fmt.Sprintf("%ss", opt.scheme)
			}
		default:
			return nil, errors.New("invalid interface type for http")
		}
	}
	u := &url.URL{
		Scheme: opt.scheme,
		Host:   addr,
		Path:   opt.HttpPath,
	}
   
}
```

jkaflik:
Hi @crisismaple 

Would you like to submit a PR with such a change?

crisismaple:
> 
Hi @jkaflik,

Sure, I already prepared a local branch for this and just waiting for a contributor role XD

jkaflik:
@crisismaple you can fork the project and then submit PR

crisismaple:
#938

PR #938, opened 2023-03-20
Support customized "url path" in http connection

crisismaple:
## Summary

add a new `HttpUrlPath` field in `Clickhouse.Options` to support customize URL path used by HTTP connection ([#937](https://github.com/ClickHouse/clickhouse-go/issues/937))


jkaflik (commented):
This change looks ok. Let's introduce a minimal test that verifies if we can connect to CH with HTTP via some proxy.

This could be an additional nginx container that does a reverse proxy.

WDYT?

crisismaple:
@jkaflik I just pushed a new iteration, please help to review it ;)

jkaflik:
@crisismaple that's great! I will release it today

PR #942, opened 2023-03-25
support query with external tables via http connection

crisismaple:
## Summary
use mutipart http request to send both external tables and query

## Checklist

### DIFF
--- a/conn_http.go
+++ b/conn_http.go
@@ -28,9 +28,11 @@ import (
 	"github.com/ClickHouse/clickhouse-go/v2/resources"
 	"io"
 	"io/ioutil"
+	"mime/multipart"
 	"net"
 	"net/http"
 	"net/url"
+	"strings"
 	"sync"
 	"time"
 
@@ -381,7 +383,7 @@ func (h *httpConnect) readData(ctx context.Context, reader *chproto.Reader) (*pr
 	return &block, nil
 }
 
-func (h *httpConnect) sendQuery(ctx context.Context, r io.Reader, options *QueryOptions, headers map[string]string) (*http.Response, error) {
+func (h *httpConnect) sendQueryBody(ctx context.Context, r io.Reader, options *QueryOptions, headers map[string]string) (*http.Response, error) {
 	req, err := h.prepareRequest(ctx, r, options, headers)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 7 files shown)


jkaflik:
Thanks @crisismaple. Let me have a look on this today!

jkaflik:
@crisismaple I didn't manage to review it yesterday. Hopefully, I will have time today.

crisismaple:
> @crisismaple I didn't manage to review it yesterday. Hopefully, I will have time today.

Take your time please~

crisismaple:
> @crisismaple could you rebase your PR?

just updated

conn_http.go
```diff
@@ -265,7 +267,7 @@ func (h *httpConnect) isBad() bool {
 }
 
 func (h *httpConnect) readTimeZone(ctx context.Context) (*time.Location, error) {
-	rows, err := h.query(ctx, func(*connect, error) {}, "SELECT timezone()")
+	rows, err := h.query(Context(ctx, ignoreExternalTables()), func(*connect, error) {}, "SELECT timezone()")
```
jkaflik, line 270: Since it's an internal logic for HTTP proto, we should not leak it into Options struct.

Instead, can we reset `external` in context? Or can we pass implicit bool to `query` func?

conn_http.go
```diff
@@ -283,7 +285,7 @@ func (h *httpConnect) readTimeZone(ctx context.Context) (*time.Location, error)
 }
 
 func (h *httpConnect) readVersion(ctx context.Context) (proto.Version, error) {
-	rows, err := h.query(ctx, func(*connect, error) {}, "SELECT version()")
+	rows, err := h.query(Context(ctx, ignoreExternalTables()), func(*connect, error) {}, "SELECT version()")
```
jkaflik, line 288: same as above

conn_http.go
```diff
@@ -454,10 +467,48 @@ func (h *httpConnect) prepareRequest(ctx context.Context, reader io.Reader, opti
 		}
 		req.URL.RawQuery = query.Encode()
 	}
-
 	return req, nil
 }
 
+func (h *httpConnect) prepareRequest(ctx context.Context, query string, options *QueryOptions, headers map[string]string) (*http.Request, error) {
+	if options == nil || options.ignoreExternalTables || len(options.external) == 0 {
```
jkaflik, line 474:
```suggestion
	if options == nil || len(options.external) == 0 {
```

`ignoreExternalTables` looks redundant in this use case, if we properly pass the context here.

conn_http.go
```diff
@@ -454,10 +467,48 @@ func (h *httpConnect) prepareRequest(ctx context.Context, reader io.Reader, opti
...
+	h.url.RawQuery = queryValues.Encode()
+	err := w.WriteField("query", query)
+	if err != nil {
+		return nil, err
+	}
+	err = w.Close()
+	if err != nil {
+		return nil, err
+	}
+	headers["Content-Type"] = w.FormDataContentType()
```
jkaflik, line 508: Let's extract this logic into sub function.

conn_http.go
crisismaple: looks like `query` is defined in `stdConnect` interface with a generic signature, modify `context` from caller side seems a more reasonable way to optimize internal calls for timezone and ping.

conn_http.go
crisismaple: just pushed an update to remove `ignoreExternalTables` in options

context.go
```diff
@@ -148,6 +148,13 @@ func WithUserLocation(location *time.Location) QueryOption {
 	}
 }
 
+func ignoreExternalTables() QueryOption {
+	return func(o *QueryOptions) error {
+		o.external = nil
```
crisismaple, line 153: force set `external` to `nil` here

jkaflik (approved):
Thanks. Staging it to a next release

PR #949, opened 2023-03-29
chore: update GH Actions CH/Go matrix

jkaflik:
## Summary
Update ClickHouse versions run in actions. Drop Go 1.18 and use 1

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -22,12 +22,9 @@ jobs:
         go:
           - "1.19"
           - "1.20"
-        clickhouse:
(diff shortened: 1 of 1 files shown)


.github/workflows/run-tests.yml
```diff
@@ -20,15 +20,18 @@ jobs:
...
         clickhouse:
-          - 22.3
-          - 22.8
-          - 22.9
-          - '22.10'
-          - 22.11
-          - latest
+          - "22.3"
+          - "22.8"
+          - "22.9"
```
mshustov, line 28: 22.9 - 22.11 aren't [maintained](https://github.com/ClickHouse/ClickHouse/blob/master/SECURITY.md) anymore, we can skip them in the tests

jkaflik: Thanks. Added a reference.

Issue #962, opened 2023-04-06
Unable to use ternary operator with named arguments


crisismaple:
### Describe the bug
Bind function will return error `clickhouse [bind]: mixed named, numeric or positional parameters` when using ternary operator "?:" with named arguments.

Maybe we should skip the `havePositional` check when passing named arguments (since the regex expression may not cover all query scenarios using ternary operators or constraint "?").

https://github.com/ClickHouse/clickhouse-go/blob/b3c72d8fb8ed175f40c9175d988024723e77fecc/bind.go#L78-L94

### Code example
```go
func TestTernaryOperator(t *testing.T) {
	sqls := []string{
		`SELECT if(@arg1,@arg2,@arg3)`, // correct
		`SELECT @arg1?@arg2:@arg3`, // failed here
	}
	for _, sql := range sqls {
		_, err := bind(time.Local, sql,
			Named("arg1", 0),
			Named("arg2", 1),
			Named("arg3", 2))
		if err != nil {
			t.Fatal(err)
		}
	}
}
```

jkaflik:
Hi, historically, I don't know why such a check was introduced. I will have to dig into it.

On a side note: did you consider using the native query parameters mechanism? https://github.com/ClickHouse/clickhouse-go/pull/854

crisismaple:
> 

This logic seems to be introduced long ago, since we are using an archived CH version (19.x) which not supporting directly pass parameters with query, we now have switched to use `if` function to replace all ternary operators as a work around.

I will also try to check binding strategies in other language implementations for clickhouse ;)

PR #965, opened 2023-04-07
Enable to use ternary operator with named arguments

crisismaple:
## Summary
Add argument type based check before binding logic for named parameters, intercept passing mixed named, numeric or positional parameters in query #962 

## Checklist


jkaflik:
Thanks. Looks good.

jkaflik (approved):

PR #971, opened 2023-04-14
Tuple scan respects both value and pointer variable

crisismaple:
## Summary
Now scan result of `tuple` typed column is `*[]interface{}`/`*map[string]interface{}` instead of `[]interface{}`/`map[string]interface{}`, since `tuple` type is not allowed to be nullable in CH, the destination type of `tuple` columns should not be pointer.

## Checklist


jkaflik:
@crisismaple could you resolve the conflict?

crisismaple:
> @crisismaple could you resolve the conflict?

Sure, just update the PR

jkaflik (approved):
Thanks! LGTM

PR #975, opened 2023-04-19
Do not return hard error on unparsable version in HTTP proto

hexchain:
## Summary

Currently, if the ClickHouse server returns a version number that cannot be parsed, clickhouse-go returns an error and refuses to continue. This change makes clickhouse-go more relaxed when the server has a non-regular non-official version: it treats unparseable parts as 0 in the version number and still works after printing a warning.


jkaflik (approved):
Thanks. Looks good.

PR #987, opened 2023-05-05
Pass http.ProxyFromEnvironment configuration to http.Transport

slvrtrn:
## Summary
I was investigating the Grafana data source issue (Cloud cannot be accessed via proxy) that boiled down to ignored HTTP_PROXY and HTTPS_PROXY env variables in the CH data source. 

I suspect this is due to the lack of `Proxy: http.ProxyFromEnvironment` configuration passed to `http.Transport`


jkaflik (approved):
The test can be skipped for now. It's flaky in CI.

PR #991, opened 2023-05-09
support http connection reuse after external table query execution

### LINKED ISSUE #990: Http connection could not be reused after executed query with external table
The sql connection could not be reused since the connection url been modified and add query parameters with external table structure https://github.com/ClickHouse/clickhouse-go/blob/e1ca29dc057ef9701a8be52aec3fa36a8c2613e8/conn_http.go#LL490C1-L510C39

### Steps to reproduce
1. init new http connection
2. run query with external tables
3. reuse that connection to run other query not related to previous external table

crisismaple:
## Summary
Support http connection reuse after external table query execution and  fix #990 

## Checklist

### DIFF
--- a/conn_http.go
+++ b/conn_http.go
@@ -477,14 +477,10 @@ func (h *httpConnect) prepareRequest(ctx context.Context, query string, options 
 	if options == nil || len(options.external) == 0 {
 		return h.createRequest(ctx, h.url.String(), strings.NewReader(query), options, headers)
 	}
-	requestUrl, payloadBytes, err := h.createRequestPayloadWithExternalTables(ctx, query, options, headers)
-	if err != nil {
-		return nil, err
-	}
-	return h.createRequest(ctx, requestUrl, bytes.NewReader(payloadBytes), options, headers)
+	return h.createRequestWithExternalTables(ctx, query, options, headers)
 }
 
-func (h *httpConnect) createRequestPayloadWithExternalTables(ctx context.Context, query string, options *QueryOptions, headers map[string]string) (string, []byte, error) {
+func (h *httpConnect) createRequestWithExternalTables(ctx context.Context, query string, options *QueryOptions, headers map[string]string) (*http.Request, error) {
 	payload := &bytes.Buffer{}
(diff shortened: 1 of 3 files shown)


conn_http.go
```diff
@@ -475,49 +475,51 @@ func (h *httpConnect) createRequest(ctx context.Context, reader io.Reader, optio
...
+	requestUrl, payloadBytes, err := h.createRequestPayloadWithExternalTables(ctx, query, options, headers)
 	if err != nil {
 		return nil, err
 	}
-	return h.createRequest(ctx, bytes.NewReader(payloadBytes), options, headers)
+	return h.createRequest(ctx, requestUrl, bytes.NewReader(payloadBytes), options, headers)
 }
 
-func (h *httpConnect) createRequestPayloadWithExternalTables(ctx context.Context, query string, options *QueryOptions, headers map[string]string) ([]byte, error) {
+func (h *httpConnect) createRequestPayloadWithExternalTables(ctx context.Context, query string, options *QueryOptions, headers map[string]string) (string, []byte, error) {
```
jkaflik, line 487: minor: can we return the entire request instead of a few values here? or make this function mutate request struct
```suggestion
func (h *httpConnect) appendExternalTablesToRequest(ctx context.Context, r *http.Request, query string, options *QueryOptions, headers map[string]string) error {
```

jkaflik (approved):

lib/column/ipv6.go
```diff
@@ -70,6 +70,27 @@ func (col *IPv6) ScanRow(dest any, row int) error {
...
+		v := col.row(row)
+		if len(v) < 16 {
+			return &ColumnConverterError{
+				Op:   "ScanRow",
+				To:   fmt.Sprintf("%T", dest),
+				From: "IPv6",
+			}
+		}
+		*d = new([16]uint8)
+		**d = IPv6ToBytes(v)
```
jkaflik, line 93: I don't think these changes are intended in this PR?

jkaflik (changes requested):

jkaflik (approved):

Issue #976, opened 2023-04-20
IPv6 column to be mapped from []uint8


lspgn:
Hello!
Thank you for this great library!

**Is your feature request related to a problem? Please describe.**

I am trying to insert data into an `IPv6` column. The real type is a `FixedString(16)` inside ClickHouse.
The `Append` function of the library requires the underlying type to be:
* string
* netip.Addr
* net.IP

I would prefer to avoid manipulating my array and insert the data directly.

**Describe the solution you'd like**

When looking at the IPv4 column, there is a way to pass a uint32
https://github.com/ClickHouse/clickhouse-go/blob/558434c679185cbf403018ebd6c74748d846343b/lib/column/ipv4.go#L74

For IPv6, strings are currently parsed and then the bytes are inserted.
I would like to be able to pass a [16]byte or a []uint8 (byte array).

**Describe alternatives you've considered**

Ended up using `netip.AddrFrom16()` but I'd like to avoid a cast or pass a pointer to the byte array I already have.

jkaflik:
Hi @lspgn 

Thanks for this request. This will likely not be a priority for me to implement in the upcoming weeks. If someone from the community wants to work on it - I am happy to merge PR.

rogeryk:
@jkaflik this issue have been solved.

gogingersnap777:
@jkaflik Looking through `good first issues` and this looks like it can be closed.

laeg:
Thanks folks! Closing this issue out

PR #996, opened 2023-05-12
Support [16]byte/[]byte typed scan/append for IPv6 column

crisismaple:
## Summary
add support for [16]byte/[]byte IPv6 columns #976 
## Checklist

### DIFF
--- a/tests/ipv6_test.go
+++ b/tests/ipv6_test.go
@@ -306,24 +306,23 @@ func assertRowColumnEqualToIP(t *testing.T, value any, ip net.IP) {
 func TestIPv6AppendRow(t *testing.T) {
 	ip := getTestIPv6()[0]
 	strIp := ip.String()
-
+	ipBytes := netip.MustParseAddr(strIp).As16()
 	col := column.IPv6{}
 
 	rows := []any{
-		strIp,                                   // appending string
-		&ip,                                     // appending IP pointer
-		ip,                                      // appending IP
-		&strIp,                                  // appending string pointer
-		netip.MustParseAddr(ip.String()).As16(), // appending [16]byte
-		proto.IPv6(netip.MustParseAddr(ip.String()).As16()),                   // appending proto.IPv6
-		&[][16]byte{netip.MustParseAddr(ip.String()).As16()}[0],               // appending [16]byte pointer
-		&[]proto.IPv6{proto.IPv6(netip.MustParseAddr(ip.String()).As16())}[0], // appending proto.IPv6 pointer
+		strIp,                     // appending string
+		&ip,                       // appending IP pointer
+		ip,                        // appending IP
+		&strIp,                    // appending string pointer
+		ipBytes,                   // appending [16]byte
+		proto.IPv6(ipBytes),       // appending proto.IPv6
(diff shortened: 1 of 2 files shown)


tests/ipv6_test.go
```diff
@@ -291,15 +339,42 @@ func TestIPv6_AppendRow(t *testing.T) {
 	if !col.Row(3, false).(net.IP).Equal(ip) {
 		require.Failf(t, "Invalid result of AppendRow", "Added %q instead of %q", col.Row(3, false), ip)
 	}
-}
 
-func TestIPv6_Append(t *testing.T) {
-	ips := getTestIPv6()
+	// appending [16]byte
+	require.NoError(t, col.AppendRow(netip.MustParseAddr(ip.String()).As16()))
+	require.Equal(t, 5, col.Rows(), "AppendRow didn't add IP")
```
jkaflik, line 345: It's better to use `assert.Equal`, since it does not fail immediately, so you see all assertion's effect. Require is good when some prerequisites setup is necessary.

tests/ipv6_test.go
```diff
@@ -291,15 +339,42 @@ func TestIPv6_AppendRow(t *testing.T) {
...
+	if !col.Row(4, false).(net.IP).Equal(ip) {
+		require.Failf(t, "Invalid result of AppendRow", "Added %q instead of %q", col.Row(4, false), ip)
+	}
+
+	// appending proto.IPv6
+	require.NoError(t, col.AppendRow(proto.IPv6(netip.MustParseAddr(ip.String()).As16())))
+	require.Equal(t, 6, col.Rows(), "AppendRow didn't add IP")
+	if !col.Row(5, false).(net.IP).Equal(ip) {
+		require.Failf(t, "Invalid result of AppendRow", "Added %q instead of %q", col.Row(5, false), ip)
+	}
```
jkaflik, line 355:
```suggestion
        assert.Truef(t, !col.Row(5, false).(net.IP).Equal(ip), "Added %q instead of %q", col.Row(5, false), ip)
```

tests/ipv6_test.go
```diff
@@ -313,25 +388,43 @@ func TestIPv6_Append(t *testing.T) {
 			require.Failf(t, "Invalid result of Append", "Added %q instead of %q", col.Row(i, false), ip)
 		}
 	}
+}
 
-	// appending string pointers
-	var strPtrIps []*string
+func TestIPv6_Append(t *testing.T) {
```
jkaflik, line 393: let's avoid underscore symbols

tests/ipv6_test.go
jkaflik: or there might be a helper func that does following

```go
func assertRowColumnEqualToIP(t *testing.T, value any, ip net.IP) {
	require.IsType(t, net.IP{}, value, "Invalid type of column value. net.IP expected.")

	assert.Truef(t, value.(net.IP).Equal(ip), "%q is not equal to IP %q", value, ip)
}
```

```suggestion
	assertRowColumnEqualToIP(t, col.Row(5, false), IP)
```

it can be reused in duplicated code

jkaflik (approved):
Thanks for your contribution. LGTM. Please let me know if you want to reiterate on my review's comments.

PR #1002, opened 2023-05-19
Docs: update options for http(s) protocols

deevroman:
## Summary

1. Starting from https://github.com/ClickHouse/clickhouse-go/pull/650 the Interface option has been renamed to Protocol
2. Starting from https://github.com/ClickHouse/clickhouse-go/pull/664 https is used when passing a non-null pointer to tls.Config to the TLS option


jkaflik:
Already fixed in main

PR #1021, opened 2023-07-07
Git actions terraform

gingerwizard:
## Summary

Moves CI/CD off a static cluster to a terraform provisioned cluster. 

We will need a means to delete these clusters - if this works ill see if we can name the cluster with a timestamp. We can then remove clusters older than a defined period with a scheduled lambda (or eqv). Clusters will idle however so no rush on our side.


Kinzeng (approved):

.github/workflows/run-tests.yml
```diff
@@ -52,7 +55,31 @@ jobs:
...
+
+      - name: Terraform Validate
+        id: validate
+        run: terraform validate -no-color
+
+      - name: Terraform Apply
+        id: apply
+        run: terraform apply -no-color
+        env:
+          TF_VAR_organization_id: '369a3c9a-868e-437e-8bdc-e98b8bd180dd'
```
jkaflik, line 78: minor, but we could have it in GH Action env vars

main.tf
```diff
@@ -0,0 +1,59 @@
...
+  environment     = "production"
+  organization_id = var.organization_id
+  token_key       = var.token_key
+  token_secret    = var.token_secret
+}
+
+resource "clickhouse_service" "service" {
+  name           = var.cluster_name
+  cloud_provider = "aws"
+  region         = "us-east-2"
```
jkaflik, line 40: Is it a region where we run our runners?

jkaflik (approved):
Thanks @gingerwizard! 
Overall, let's try to add a cleanup action with `terraform destroy`

PR #1023, opened 2023-07-10
Git actions terraform

gingerwizard:
# Summary

Moves CI/CD off a static cluster to a terraform provisioned cluster.

We will need a means to delete these clusters - if this works ill see if we can name the cluster with a timestamp. We can then remove clusters older than a defined period with a scheduled lambda (or eqv). Clusters will idle however so no rush on our side.


gingerwizard:
All tests pass @jkaflik all ready to merge!

main.tf
```diff
@@ -0,0 +1,59 @@
...
+  name           = var.cluster_name
+  cloud_provider = "aws"
+  region         = "us-east-2"
+  tier           = "development"
+  idle_scaling   = true
+  password  = var.cluster_password
+
+  ip_access = [
+    {
+        source      = "0.0.0.0/0"
```
jkaflik, line 47: Small improvement for the future: we pass explicit runner NAT IP

main.tf
```diff
@@ -0,0 +1,59 @@
+terraform {
+  required_providers {
+    clickhouse = {
+      source = "ClickHouse/clickhouse"
+      version = "0.0.2"
```
jkaflik, line 5: Let's either allow to use newer patch version
```suggestion
      version = "~> 0.0.2"
```
or minimum accepted version
```suggestion
      version = ">= 0.0.2"
```

jkaflik (approved):
LGTM. Room for small improvements

Issue #1028, opened 2023-07-14
IP netip.Addr pointer panic


anjmao:
Types with netip.Addr pointers panics both append and scan.

jkaflik:
Fixed in #1029

PR #1029, opened 2023-07-14
Fix netip.Addr pointer panic

anjmao:
## Summary
[](https://github.com/ClickHouse/clickhouse-go/issues/1028)

## Checklist


anjmao:
Found one more issue. Array(IPv4) will not work since Row() any returns net.IP type.

anjmao:
@gingerwizard Hi, can we merge this fix for IPv4 type? Arrays and Tuples can be addressed later.

PR #1032, opened 2023-07-21
Remove incompatible server version direct stdout

### LINKED ISSUE #1031: Warning when clickhouse server version is not supported is polluting the log
### Describe the bug
When using clickhouse-go to connect to a unsupported clickhouse version, a message is printed in stdout: `WARNING: version 21.3.13 of ClickHouse is not supported by this client`. It's not a good practice to print messages directly like, especially since it's not actually an error. In my case, I have to create different connections (I can't use just one connection pool), so my application log gets bombarded by these messages.

### Steps to reproduce
1. Use clickhouse-go v2.10.1 to connect to an unsupported CH server (like 21.3.13) via HTTP or Native client.

### Expected behaviour
Give the client user the ability to turn the warning on/off or to redirect it somewhere else.

#### ClickHouse server
* ClickHouse Server version: 21.3.13

jkaflik:
## Summary
This PR removes a direct stdout with a message about an incompatible client version.

Resolves #1031

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -102,8 +102,6 @@ func dial(ctx context.Context, addr string, num int, opt *Options) (*connect, er
 
 	// warn only on the first connection in the pool
 	if num == 1 && !resources.ClientMeta.IsSupportedClickHouseVersion(connect.server.Version) {
-		// send to debugger and console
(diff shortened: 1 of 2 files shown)


Slach:
thanks god, i hope this PR will merges, it was so annoying message

PR #1045, opened 2023-07-27
Fix send query on batch retry

EpicStep:
Hi, now `retry()` create empty connection, without `INSERT INTO` statement, because of this, it cannot be used.


EpicStep:
Hi @jkaflik, I think some tests are flaky, this is true?

jkaflik:
Hi @EpicStep 
Not really. This might be a regression with a recent ClickHouse version. I will be checking this.

EpicStep:
> Hi @EpicStep
> Not really. This might be a regression with a recent ClickHouse version. I will be checking this.

Okey, thanks you

jkaflik:
@EpicStep I will release this today

PR #1062, opened 2023-08-07
Implement release connection in batch

### LINKED ISSUE #1047: Acquire connection in batch only when needed
Hi, now batch requires an acquired connection at the all time before `Send()`, but I think for `Append()` operation we doesn't need to keep connection. It would be good to have a function that would return the connection to the pool (in order not to break backward compatibility, in v3 it would be nice to see option in batch constructor). Also when calling `Send()` or first `Flush()` a connection from pool need to be acquired.

If you accept this feature request I can take and develop it.

EpicStep:
With this change, you can append to the batch directly without using helper structures (slice). Earlier this led to an empty pool, although it is not necessary to keep the connection in the batcher for the append operation.

In our implementation we saved about 2GB of memory per pod.

screen

Close #1047

### DIFF
--- a/tests/batch_release_connection_test.go
+++ b/tests/batch_release_connection_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"github.com/ClickHouse/clickhouse-go/v2"
+	"github.com/ClickHouse/clickhouse-go/v2/lib/driver"
 	"github.com/stretchr/testify/require"
 	"testing"
 )
@@ -27,13 +28,9 @@ func TestBatchReleaseConnection(t *testing.T) {
 		dropTable(conn, tableName)
 	}()
 	require.NoError(t, conn.Exec(ctx, ddl))
-	batch, err := conn.PrepareBatch(ctx, fmt.Sprintf("INSERT INTO %s", tableName))
+	batch, err := conn.PrepareBatch(ctx, fmt.Sprintf("INSERT INTO %s", tableName), driver.WithReleaseConnection())
 	require.NoError(t, err)
 	require.NoError(t, batch.Append(uint64(1), "test"))
-
-	require.NoError(t, batch.ReleaseConnection())
-	require.Error(t, batch.ReleaseConnection())
-
 	require.NoError(t, batch.Send())
 	require.Equal(t, uint64(1), getRowsCount(t, conn, tableName))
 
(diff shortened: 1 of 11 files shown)


tests/batch_release_connection_test.go
```diff
@@ -0,0 +1,76 @@
...
+	defer func() {
+		dropTable(conn, tableName)
+	}()
+	require.NoError(t, conn.Exec(ctx, ddl))
+	batch, err := conn.PrepareBatch(ctx, fmt.Sprintf("INSERT INTO %s", tableName))
+	require.NoError(t, err)
+	require.NoError(t, batch.Append(uint64(1), "test"))
+
+	require.NoError(t, batch.ReleaseConnection())
+	require.Error(t, batch.ReleaseConnection())
```
jkaflik, line 35: nit: assert error type/message

tests/batch_release_connection_test.go
```diff
@@ -0,0 +1,76 @@
...
+		dropTable(conn, tableName)
+	}()
+	require.NoError(t, conn.Exec(ctx, ddl))
+	batch, err := conn.PrepareBatch(ctx, fmt.Sprintf("INSERT INTO %s", tableName))
+	require.NoError(t, err)
+	require.NoError(t, batch.Append(uint64(1), "test"))
+
+	require.NoError(t, batch.ReleaseConnection())
+	require.NoError(t, batch.Flush())
+	require.Error(t, batch.ReleaseConnection())
```
jkaflik, line 72: nit: assert error type/message

jkaflik (changes requested):
Thanks for your contribution. I understand the need for this; however not sure if we should expose `ReleaseConnection`, but do it for the user instead. This might be overhead for short-live batches. 

Lets address comments and think if this approach is clean.

Fewer API functions we expose, the more clarity in the usage.

EpicStep:
I agree with you. I think we can add options in `PrepareBatch` (it won't break backwards compatibility).
Example: `PrepareBatch(context.Background(), "INSERT INTO some_table", WithReleaseConnection())`

EpicStep:
@jkaflik Thanks. I added this feature to README.md

EpicStep:
When will these changes be released?

jkaflik (approved):

jkaflik:
@EpicStep I plan to release it today. In a merge train, your PR is last and has conflicts. It's easy, but it would be great if you push from your fork.

EpicStep:
@jkaflik Yes, I can merge

EpicStep:
@jkaflik done :)

PR #1075, opened 2023-08-28
Fix fmt.Stringer.String() called on nil pointer

wilhelmliao:
## Summary
got error when call bind() using parameter `(*fmt.Stringer)(nil)`.

```golang
var v *json.Number  // fmt.Stringer
query := "SELECT ?"
params := []interface{}{
v,
}
sql, err := bindPositional(time.UTC, query, params...)
fmt.Println(sql)
```
want
```
SELECT NULL
```
got
```
panic: value method encoding/json.Number.String called using nil *Number pointer
```

## Checklist
- [v ] Unit and integration tests covering the common scenarios were added


bind_test.go
```diff
@@ -18,11 +18,13 @@
 package clickhouse
 
 import (
+	"encoding/json"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"go.openly.dev/pointy"
```
jkaflik, line 27: let's avoid introducing additional dependency

wilhelmliao: fixed

jkaflik:
@wilhelmliao could you remove conflicts?

wilhelmliao:
I fixed the conflicts, but the patched codes seem missing.

https://github.com/wilhelmliao/clickhouse-go/blob/main/bind.go#L308:313
```golang
	case fmt.Stringer:
		rv := reflect.ValueOf(v)
		if rv.Kind() == reflect.Pointer {
			if rv.IsNil() {
				return "NULL", nil
			}
		}
		return quote(v.String()), nil
```

jkaflik:
@wilhelmliao yes, there was a PR improving bind - https://github.com/ClickHouse/clickhouse-go/pull/1092.

Thanks for your effort!

PR #1092, opened 2023-09-11
Improving object allocation for (positional) parameter binding

mdonkers:
## Summary

I noticed _very_ high object allocation count with the ClickHouse Go client in our code. While with binding the parameters to a `String` this is logical to some extend, I found some easy improvements:

1) Mis-use of the `strings.NewReplacer`, as this instance is meant to be re-used but a new instance was created all the time
2) Replacement of positional arguments in the query string, for which not really a regex is needed. Specifically to find `\?` cases, two regex replacements were nested, creating a temporary string instance that shouldn't be needed. So I changed to code to a 'simple' loop that copies byte sequences.

Original object allocation:
2023-09-11-10-21-14_1899x227

Allocations after just changing `strings.NewReplacer` to be called once:
2023-09-11-10-21-56_1906x257

Final result:
2023-09-11-17-13-17_1901x169

(of course above numbers are case specific, but for our instance this meant going from being responsible for 70% of object allocations to around 30%)

## Checklist

### DIFF
--- a/bind.go
+++ b/bind.go
@@ -244,9 +244,11 @@ func formatTime(tz *time.Location, scale TimeUnit, value time.Time) (string, err
 	return fmt.Sprintf("toDateTime64('%s', %d, '%s')", value.Format(fmt.Sprintf("2006-01-02 15:04:05.%0*d", int(scale*3), 0)), int(scale*3), value.Location().String()), nil
 }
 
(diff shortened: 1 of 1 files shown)


jkaflik (changes requested):
@mdonkers this is great! Thanks. Please remove changes to CHANGELOG.md. It's generated automatically. We have quite unstable runners nowadays - i will take a look.

jkaflik (approved):

PR #1091, opened 2023-09-08
Add FreeBufOnConnRelease to clickhouse.Options

cergxx:
## Summary
Add a flag to `clickhouse.Options` which will trigger connection's buffer being dropped on query execution finish
Resolves issue #1086

## Checklist
- [x] Unit and integration tests covering the common scenarios were added
- [ ] A human-readable description of the changes was provided to include in CHANGELOG

Issue #1086, opened 2023-09-06
More control over connection's buffers


cergxx:
In my project I have a pool of long-living connections to ClickHouse.
Each of them holds a buffer `clickhouse.connect.buffer`, which leads to overall big memory usage

I would like to have more control over these buffers – to be able to drop them without closing the connection

Possible solutions:
Add an option to `clickhouse.Options` like `FreeBuffersOnRelease`, which would trigger drop of the `buffer` on every connection release
(Replace `b.Buf = b.Buf[:0]` with `b.Buf = nil`)

jkaflik:
Hi @cergxx 

Would you like to submit a PR with change?

cergxx:
> Would you like to submit a PR with change?

Yes, I would

PR #1100, opened 2023-09-21
Free compressor's buffer when FreeBufOnConnRelease enabled

cergxx:
## Summary
Additional fix to #1086
In #1091 I added a `FreeBufOnConnRelease` option to drop connections' buffers.
It turned out that this option would only make sense if it dropped compressor's buffer as well.


jkaflik (approved):
Thanks!

PR #1104, opened 2023-09-25
Fix: Block stream read process would be terminated by empty block with zero rows

crisismaple:
## Summary
Currently, empty row blocks will terminate block stream reading (i.e. `rows.Next() == false` ), this was caused by the fall through [row count increase logic](https://github.com/ClickHouse/clickhouse-go/blob/75abed29e88fa34064d97a63ae10a37c51394899/clickhouse_rows.go#L64C1-L67C9) after block reset.

## Checklist

### DIFF
--- a/clickhouse_rows.go
+++ b/clickhouse_rows.go
@@ -52,7 +52,6 @@ next:
 				r.err = err
 				return false
 			}
-			goto next
 		case block := <-r.stream:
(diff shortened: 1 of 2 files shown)


jkaflik (approved):
Thanks @crisismaple . Much appreciated work.

PR #1115, opened 2023-10-11
Fix an empty JSON map append using _dummy subcolumn

### LINKED ISSUE #1113: Mismatched len of columns when inserting empty map into JSON column
### Steps to reproduce
1. Enable the experimental JSON type (eg., `SET allow_experimental_object_type = 1`)
2. Give to a column the experimental JSON type
3. Insert an empty map (ie., `map[string]interface{}{}`)

### Expected behavior

I'd expect the JSON column to accept empty maps like `map[string]interface{}{}`.

The same way it works if inputting a string representing an empty JSON (eg., `"{}"`).

 Instead, it results in an error (see below) coming from [lib/proto/block.go](https://github.com/ClickHouse/clickhouse-go/blob/4f115fc2ea1ca12f43383643867eac632dc85760/lib/proto/block.go#L132).

### Code example

To prove this bug I've modified your example [clickhouse_api/json.go](https://github.com/ClickHouse/clickhouse-go/blob/main/examples/clickhouse_api/json.go) by adding a fourth column (`Col4`) (code below).

```go
package main

import (
	"context"
	"fmt"

)

func InsertReadJSON() error {
	opts := &clickhouse.Options{
		Settings: clickhouse.Settings{
			"allow_experimental_object_type": 1,
		},
	}

	conn, err := clickhouse.Open(opts)
	if err != nil {
		return err
	}
	ctx := context.Background()

	conn.Exec(ctx, "DROP TABLE IF EXISTS example")

	if err = conn.Exec(ctx, `
		CREATE TABLE example (
				Col1 JSON,
				Col2 JSON,
				Col3 JSON,
				Col4 JSON
			)
			Engine Memory
		`); err != nil {
		return err
	}

	type User struct {
		Name     string `json:"name"`
		Age      uint8  `json:"age"`
		Password string `ch:"-"`
	}

	defer func() {
		conn.Exec(ctx, "DROP TABLE example")
	}()
	batch, err := conn.PrepareBatch(ctx, "INSERT INTO example")
	if err != nil {
		return err
	}
	// we can insert JSON as either a string, struct or map
	col1Data := `{"name": "Clicky McClickHouse", "age": 40, "password": "password"}`
	col2Data := User{
		Name:     "Clicky McClickHouse Snr",
		Age:      uint8(80),
		Password: "random",
	}
	col3Data := map[string]any{
		"name":     "Clicky McClickHouse Jnr",
		"age":      uint8(10),
		"password": "clicky",
	}
	col4Data := map[string]interface{}{} // Notice that `{}` works resulting in an output of `map[_dummy:0]`

	// both named and unnamed can be added with slices
	if err = batch.Append(col1Data, col2Data, col3Data, col4Data); err != nil {
		return err
	}

	if err = batch.Send(); err != nil {
		return err
	}
	// we can scan JSON into either a map or struct
	var (
		col1 map[string]any
		col2 map[string]any
		col3 User
		col4 map[string]any
	)
	// named tuples can be retrieved into a map or slices, unnamed just slices
	if err = conn.QueryRow(ctx, "SELECT * FROM example").Scan(&col1, &col2, &col3, &col4); err != nil {
		return err
	}
	fmt.Printf("row: col1=%v, col2=%v, col3=%v\n, col4=%v\n", col1, col2, col3, col4)

	return nil
}

func main() {
	if err := InsertReadJSON(); err != nil {
		fmt.Println(err)
	}
}
```

### Error log

When running the code above you'll get:

```
clickhouse [Encode]:  mismatched len of columns - expected 1, received 0 for col Col4
```

Notice that using `col4 = "{}"` instead it works as expected -  ie. outputting the following:

```
row: col1=map[age:40 name:Clicky McClickHouse password:password], col2=map[age:80 name:Clicky McClickHouse Snr], col3={Clicky McClickHouse Jnr 10 }
, col4=map[_dummy:0]
```

#### Environment
* Client version: 23.9.1.1854 (official build)
* Language version: github.com/ClickHouse/clickhouse-go/v2 v2.14.2
* OS: 23.9.1 revision 54466
* Interface: ClickHouse API 

#### ClickHouse server
* ClickHouse Server version: 23.9.1 revision 54466
* ClickHouse Server non-default settings, if any: NONE
* `CREATE TABLE` statements for tables involved: NONE

jkaflik:
if map is empty, we need to create an empty Tuple to make sure subcolumns protocol is happy
_dummy is a ClickHouse internal name for empty Tuple subcolumn
 it has the same effect as `INSERT INTO single_json_type_table VALUES ('{}');`
 
 resolves #1113

### DIFF
--- a/lib/column/json.go
+++ b/lib/column/json.go
@@ -503,12 +503,17 @@ func appendStructOrMap(jCol *JSONObject, data any) error {
 		}
 
 		if vData.Len() == 0 {
-			// if map is empty, we need to create a dummy map to make ClickHouse Tuple happy
-			// this is exactly what ClickHouse does when it receives an empty `{}` object string
-			// JSON representation will become '{"_dummy":0}'
-			// JSON object type is experimental and may change in the future
-			vData = reflect.MakeMap(reflect.MapOf(reflect.TypeOf("_dummy"), reflect.TypeOf(uint8(0))))
-			vData.SetMapIndex(reflect.ValueOf("_dummy"), reflect.ValueOf(uint8(0)))
+			// if map is empty, we need to create an empty Tuple to make sure subcolumns protocol is happy
+			// _dummy is a ClickHouse internal name for empty Tuple subcolumn
+			// it has the same effect as `INSERT INTO single_json_type_table VALUES ('{}');`
+			emptyVal := &JSONValue{Interface: &UInt8{name: "_dummy"}}
+			if err := emptyVal.appendEmptyValue(); err != nil {
+				return err
+			}
+			jCol.columns = []JSON{
+				emptyVal,
+			}
+			return nil
 		}
 
 		return iterateMap(vData, jCol, 0)
--- a/lib/column/json.go
+++ b/lib/column/json.go
@@ -501,6 +501,16 @@ func appendStructOrMap(jCol *JSONObject, data any) error {
 				Err:        fmt.Errorf("map keys must be string for column %s", jCol.Name()),
 			}
 		}
+
+		if vData.Len() == 0 {
+			// if map is empty, we need to create a dummy map to make ClickHouse Tuple happy
+			// this is exactly what ClickHouse does when it receives an empty `{}` object string
+			// JSON representation will become '{"_dummy":0}'
+			// JSON object type is experimental and may change in the future
+			vData = reflect.MakeMap(reflect.MapOf(reflect.TypeOf("_dummy"), reflect.TypeOf(uint8(0))))
+			vData.SetMapIndex(reflect.ValueOf("_dummy"), reflect.ValueOf(uint8(0)))
+		}
+
 		return iterateMap(vData, jCol, 0)
 	}
 	return &UnsupportedColumnTypeError{
--- a/tests/issues/1113_test.go
+++ b/tests/issues/1113_test.go
@@ -19,12 +19,7 @@ func Test1113(t *testing.T) {
 	)
 	ctx := context.Background()
 	require.NoError(t, err)
-	const ddl = `
-		CREATE TABLE test_1113 (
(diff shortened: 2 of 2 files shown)


leodido:
I've also implemented a fix @jkaflik. 

Well, I've also implemented 2. I'm gonna write them here in case you find them useful.

The first fix is to replicate the behavior for the string `{}` in the map case (so in the `appendStructOrMap` function).

```go
// Handle empty maps
if jCol.columns == nil && vData.Len() == 0 {
	jCol.encoding = 1
	jCol.columns = append(jCol.columns, &JSONValue{Interface: &String{}})
	return jCol.columns[0].AppendRow(nil) // `{}` input also works
}
```

The alternative fix that I've developed is (always in the same spot in the `appendStructOrMap` function).

```go
if jCol.columns == nil && vData.Len() == 0 {
	jCol.upsertValue("_dummy", "Int8")
	jCol.insertEmptyColumn("_dummy")
	return nil
}
```

In terms of visual results they're totally equivalent to the one you've implemented. Ie, `col4=map[_dummy:0]`

jkaflik:
@leodido 

Thanks. Both proposals would work. The second one is cleaner but has a slight overhead on type resolve. In that case, I would not bother about it since the entire JSON implementation is costly anyway.

Feel free to submit your second proposal, but please add a few lines of comments, so anyone who reviews it later is not frustrated around this hack. :) If you submit it today, I am ok to merge yours 🚀

leodido:
On it!

leodido:
Here it is: https://github.com/ClickHouse/clickhouse-go/pull/1116

jkaflik:
Closed in favour of https://github.com/ClickHouse/clickhouse-go/pull/1116

PR #1116, opened 2023-10-11
Fix insertion of empty map into JSON column by using _dummy subcolumn

leodido:
## Summary

Fixes https://github.com/ClickHouse/clickhouse-go/issues/1113 (in the issue the code to reproduce it)

This PR correctly handles the inputting of an empty map,
by creating an empty tuple to ensure the subcolumns protocol is happy.

The `_dummy` key is a ClickHouse internal name for an empty tuple subcolumn. In fact, it is the same key it gets used when inputting `{}`.

With this PR,
inputting `map[string]any{}` from Go has the same effect as `INSERT INTO single_json_type_table VALUES ('{}');`.

In @jkaflik's PR https://github.com/ClickHouse/clickhouse-go/pull/1115 there are other evaluated fixes/solutions.

## Checklist

### DIFF
--- a/lib/column/json.go
+++ b/lib/column/json.go
@@ -501,6 +501,13 @@ func appendStructOrMap(jCol *JSONObject, data any) error {
 				Err:        fmt.Errorf("map keys must be string for column %s", jCol.Name()),
 			}
 		}
+		if jCol.columns == nil && vData.Len() == 0 {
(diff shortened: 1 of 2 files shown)


jkaflik:
Nice! Thanks @leodido

jkaflik (approved):

PR #1129, opened 2023-10-23
Support scan of custom types implementing sql.Scanner into array/tuple columns

### LINKED ISSUE #1128: Can't scan an array of non-`string` custom types
### Describe the bug

Can't read (ie., scan) an array of custom types when those custom types don't define the `string` type.

As you can observe in the reproducer below:

1. Writing/reading an array of a `string` custom type works
    * ie., `[]SomeString` -> `Array(LowCardinality(String))`
2. Writing/reading a custom type defining a type different than `string` work implementing `sql.Scanner` and `fmt.Stringer`
    * ie., `SomeUint64AsString` -> `String`
3. Reading an array of a custom type defining some type other than `string` **doesn't work**
    * writing works
    * can't scan values back
    * ie., `[]SomeUint64AsString` -> `Array(LowCardinality(String))`
    * See `Test3` field in the reproducer
    * Errors with `clickhouse [ScanRow]: (Col3) converting string to main.SomeUint64AsString is unsupported`

Sending a pull request with a fix soon.

I've also found that reading an array of pointers to a custom type **doesn't work** (see `Test4` and `Test5` fields in the reproducer), but maybe that's a topic for another issue/PR.

### Steps to reproduce

1. Write an array of non-`string` custom type
2. Try to scan them back

### Expected behavior

I'd expect it to work just like it works for non-array custom types (see `Test2` field in the reproducer).

### Code example

Here's a reproducer.

```go
package main

import (
	"context"
	"fmt"

)

const (
	A SomeUint64AsString = iota + 1
	B
	C
	D
	E
)

type SomeUint64AsString uint64

// Scan implements sql.Scanner for the type SomeUint64AsString.
func (f *SomeUint64AsString) Scan(src any) error {
	if t, ok := src.(uint64); ok {
		*f = SomeUint64AsString(t)
		return nil
	}
	if t, ok := src.(string); ok {
		switch t {
		case "A":
			*f = A
		case "B":
			*f = B
		case "C":
			*f = C
		default:
			return fmt.Errorf("cannot scan %s into SomeUint64AsString", t)
		}
		return nil
	}

	return fmt.Errorf("cannot scan %T into SomeUint64AsString", src)
}

func (f SomeUint64AsString) String() string {
	switch f {
	case A:
		return "A"
	case B:
		return "B"
	case C:
		return "C"
	}
	return ""
}

type SomeString string

func (s SomeString) String() string {
	return string(s)
}

func InsertAndRead() error {
	opts := &clickhouse.Options{
		Settings: clickhouse.Settings{},
	}

	conn, err := clickhouse.Open(opts)
	if err != nil {
		return err
	}
	ctx := context.Background()

	conn.Exec(ctx, "DROP TABLE IF EXISTS example")

	if err = conn.Exec(ctx, `
		CREATE TABLE example (
				Col1 Array(LowCardinality(String))
			, 	Col2 String
			,	Col3 Array(LowCardinality(String))
			//,	Col4 Array(LowCardinality(String))
			//, 	Col5 String
		)
		Engine Memory
		`); err != nil {
		return err
	}

	type Test struct {
		Test1 []SomeString         `ch:"Col1"` // Works
		Test2 SomeUint64AsString   `ch:"Col2"` // Works
		Test3 []SomeUint64AsString `ch:"Col3"` // DOESN'T WORK
		// Test4 []*SomeString        `ch:"Col4"` // DOESN'T WORK TOO
		// Test5 *SomeUint64AsString  `ch:"Col5"` // DOESN'T WORK TOO
	}

	defer func() {
		conn.Exec(ctx, "DROP TABLE example")
	}()
	batch, err := conn.PrepareBatch(ctx, "INSERT INTO example")
	if err != nil {
		return err
	}

	colData := Test{
		Test1: []SomeString{"one", "two"},
		Test2: A,
		Test3: []SomeUint64AsString{B, C},
		// Test4: []*SomeString{func() *SomeString { x := SomeString("three"); return &x }()},
		// Test5: func() *SomeUint64AsString { x := E; return &x }(),
	}
	// both named and unnamed can be added with slices
	if err = batch.AppendStruct(&colData); err != nil {
		return err
	}

	if err = batch.Send(); err != nil {
		return err
	}

	var (
		col Test
	)
	// named tuples can be retrieved into a map or slices, unnamed just slices
	if err = conn.QueryRow(ctx, "SELECT * FROM example").ScanStruct(&col); err != nil {
		return err
	}
	fmt.Printf("row: col=%v\n", col)

	return nil
}

func main() {
	if err := InsertAndRead(); err != nil {
		fmt.Println(err)
	}
}
```

#### Environment
* Client version: 23.9.1.1854 (official build)
* Language version: github.com/ClickHouse/clickhouse-go/v2 v2.14.3
* OS: Darwin 23.0.0 arm64
* Interface: ClickHouse API

#### ClickHouse server
* ClickHouse Server version: 23.9.1 revision 54466
* ClickHouse Server non-default settings, if any: NONE
* `CREATE TABLE` statements for tables involved: NONE

leodido:
## Summary

Fixes #1128

## Checklist

### DIFF
--- a/lib/column/tuple.go
+++ b/lib/column/tuple.go
@@ -18,14 +18,16 @@
 package column
 
 import (
+	"database/sql"
 	"fmt"
-	"github.com/ClickHouse/ch-go/proto"
-	"github.com/google/uuid"
-	"github.com/shopspring/decimal"
 	"net"
 	"reflect"
 	"strings"
 	"time"
+
+	"github.com/ClickHouse/ch-go/proto"
+	"github.com/google/uuid"
+	"github.com/shopspring/decimal"
 )
 
 type Tuple struct {
@@ -207,6 +209,15 @@ func setJSONFieldValue(field reflect.Value, value reflect.Value) error {
 	if value.CanConvert(field.Type()) {
 		field.Set(value.Convert(field.Type()))
 		return nil
+	}
+
+	// check if our target implements sql.Scanner
+	sqlScanner := reflect.TypeOf((*sql.Scanner)(nil)).Elem()
+	if fieldAddr := field.Addr(); field.Kind() != reflect.Ptr && fieldAddr.Type().Implements(sqlScanner) {
+		returns := fieldAddr.MethodByName("Scan").Call([]reflect.Value{value})
+		if len(returns) > 0 && returns[0].IsNil() {
+			return nil
+		}
 	}
 
 	return &ColumnConverterError{
--- a/tests/issues/1128_test.go
+++ b/tests/issues/1128_test.go
@@ -1,0 +1,88 @@
+package issues
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/ClickHouse/clickhouse-go/v2"
+	clickhouse_tests "github.com/ClickHouse/clickhouse-go/v2/tests"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	A SomeUint64AsString = iota + 1
+	B
+	C
+)
+
+type SomeUint64AsString uint64
+
+func (f *SomeUint64AsString) Scan(src any) error {
+	if t, ok := src.(uint64); ok {
+		*f = SomeUint64AsString(t)
+		return nil
+	}
+	if t, ok := src.(string); ok {
+		switch t {
+		case "A":
+			*f = A
+		case "B":
+			*f = B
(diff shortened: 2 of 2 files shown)


jkaflik (approved):
Thanks for your contribution! 💪

PR #1130, opened 2023-10-24
reflection-free AppendRow for plain arrays

### LINKED ISSUE #1122: Do not use reflect when writing plain array columns
**Is your feature request related to a problem? Please describe.**
In my application I mostly write big plain string arrays to CH e.g. `[]string`. I noticed, that using of reflection is responsible for 28% of all the allocations and at least 5% of cpu usage in my app.
allocs
cpu

**Describe the solution you'd like**
I believe it is possible to get rid of reflection here, at least for the plain arrays.
I have a draft proof of concept of what I think would work fine here: #1121

cergxx:
## Summary
Resolves #1122


jkaflik (approved):
This is super nice! Thank you!

Issue #997, opened 2023-05-15
Append does not honor Valuer interface for int32-backed enum.


todor-mutafov:
### Describe the bug
I'm using Gorm, and I have an enum backed by a `int32`. I'm trying to store it to a `Nullable(Int32)` by using a pointer to the enum value. When trying to store the row, I get the error attached below.
I tried implementing `driver.Valuer`, and it didn't help.

From debugging a bit I notice that `clickhouse-go` uses the valuer to determine the nil-or-not-nil state, but not the actual value of the field.

### Expected behaviour
The library should either automatically handle the conversion between the enum and the backing type, or at the very least it should use the `driver.Valuer` implementation properly to determine the value.

### Code example
```go
type MyEnum int32

// ...

func (m *MyEum) Value() (driver.Value, error) {
	converter := driver.Null{Converter: driver.Int32}
	if m == nil {
		return converter.ConvertValue(nil)
	}
	return converter.ConvertValue(int32(*m))
}

type MyEntity struct {
	Timestamp time.Time `gorm:"type:DateTime('UTC')"`
	MyEnumValue *MyEnum `gorm:"type:Nullable(Int32)"`
	// ...
}
```

### Error log
`clickhouse [AppendRow]: MyEnumValue clickhouse [AppendRow]: converting *example.MyEnum to Int32 is unsupported`

#### Environment
* Client version: `github.com/ClickHouse/clickhouse-go/v2 v2.8.2`
* Language version: `go version go1.20.4 linux/amd64`
* OS: `Linux d30d8d32178c 5.10.102.1-microsoft-standard-WSL2 #1 SMP Wed Mar 2 00:30:59 UTC 2022 x86_64 GNU/Linux`
* Interface: https://github.com/go-gorm/clickhouse

#### ClickHouse server
* ClickHouse Server version: `22.3.12.21.altinitystable`
* `CREATE TABLE` statements for tables involved:
```sql
CREATE TABLE clickhouse.MyEntity (
	`Timestamp` DateTime('UTC'),
	MyEnumValue Nullable(Int32)
) ENGINE = MergeTree
PARTITION BY toYYYYMM(`Timestamp`)
PRIMARY KEY (`Timestamp`)
ORDER BY (`Timestamp`)
SETTINGS index_granularity = 8192;
```

jkaflik:
Hi @todor-mutafov 

[...]

PR #1144, opened 2023-11-17
Add sql.Valuer support for all types

### LINKED ISSUE #1143: sql.Valuer support
### Describe the bug

It appears that `sql.Valuer` is only implemented for `String` and `FixedString`. `String` and `FixedString` also only check for the returned value from `sql.Valuer.Value()` call is a string instead of rechecking the supported types in the preceding logic.

### Steps to reproduce
1. Wrap any type such as Float64 in valuer supported type such as [this Option generic type](https://github.com/go-playground/pkg/blob/master/values/option/option.go#L49C6-L49C12) that implements the `sql.Valuer` type.
2. Attempt to insert a record into the database.
3. Error occurs about unsupported type.

### Expected behaviour
The code to detect `sql.Valuer`, just like the code does for `sql.Scanner` when reading data out, to put data in.

### Code example
```go
type testDateTimeSerializer struct {
	val time.Time
}

func (c testDateTimeSerializer) Value() (driver.Value, error) {
	return c.val, nil
}

func (c *testDateTimeSerializer) Scan(src any) error {
	if t, ok := src.(time.Time); ok {
		*c = testDateTimeSerializer{val: t}
		return nil
	}
	return fmt.Errorf("cannot scan %T into testDateTimeSerializer", src)
}

func TestDateTimeValuer(t *testing.T) {
	conn, err := GetNativeConnection(nil, nil, &clickhouse.Compression{
	})
	ctx := context.Background()
	require.NoError(t, err)
	defer func() {
		conn.Exec(ctx, "DROP TABLE datetime_valuer")
	}()
	const ddl = `
		CREATE TABLE datetime_valuer (
			  Col1 DateTime
		) Engine MergeTree() ORDER BY tuple()
		`
	require.NoError(t, conn.Exec(ctx, ddl))
	batch, err := conn.PrepareBatch(ctx, "INSERT INTO datetime_valuer")
	require.NoError(t, err)
	vals := [1000]time.Time{}
	var now = time.Now()
	for i := 0; i < 1000; i++ {
		vals[i] = now.Add(time.Duration(i) * time.Hour).Truncate(time.Second)
		batch.Append(testDateTimeSerializer{val: vals[i]})
		require.Equal(t, 1, batch.Rows())
		batch.Flush()
	}
	require.Equal(t, 0, batch.Rows())
	batch.Send()
	rows, err := conn.Query(ctx, "SELECT * FROM datetime_valuer")
	require.NoError(t, err)
	i := 0
	for rows.Next() {
		var col1 time.Time
		require.NoError(t, rows.Scan(&col1))
		assert.Equal(t, vals[i].In(time.UTC), col1)
		i += 1
	}
}
```
### Error log
```go
return &ColumnConverterError{
	Op:   "AppendRow",
	To:   "DateTime",
	From: fmt.Sprintf("%T", v),
}
```

#### Environment
* Client version: v2.15.0
* Language version: Go v1.21.4
* OS: All
* Interface: database/sql compatible driver

#### ClickHouse server
* ClickHouse Server version: v23.8
* ClickHouse Server non-default settings, if any: N/A

deankarn:
## Summary

This PR adds `sql.Valuer` support for all types, except experimental type JSON.

Resolves #1143 #997 

## Checklist


jkaflik:
Hi @deankarn 
Thanks for submitting. I will review PR this week.

lib/column/array.go
```diff
@@ -191,13 +191,23 @@ func appendNullableRowPlain[T any](col *Array, arr []*T) error {
...
+				}
 			}
+			return nil
+		}
+		return &ColumnConverterError{
+			Op:   "AppendRow",
+			To:   "Array",
+			From: fmt.Sprintf("%T", elem),
 		}
-		return nil
```
jkaflik, line 200: could you elaborate on this change please?

jkaflik (commented):
Overall looks good. Can you please see my question?

lib/column/array.go
deankarn: While I was adding tests for the enum types it reached this code path and library panic’d on calling `elem.Len()` as the reflected value was not valid to call that method, so I added this check so at least an error would be returned instead of entire application panicing.

iirc in my case adding `Valuer` support to the Array type caused my specific tests not to hit this issue anymore, however decided to leave this safeguard as it was still possible for this panic to happen to others custom types.

jkaflik: Makes sense! thanks for answering

jkaflik (approved):
Good piece of work. Thank you!

jkaflik:
@deankarn there is a conflict to be resolved. Can you take a look? I would like to release tomorrow.

deankarn:
@jkaflik I will make it my priority as soon as I am at my desk in an hour.

Issue #1154, opened 2023-12-08
Native Driver support for Custom Bool type in struct 


yogasw:
**Is your feature request related to a problem? Please describe.**
I have a structure with custom boolean type CustomBool that I use to handle data from JSON and data from clickhouse.

```go
import (
	"database/sql"
	"gopkg.in/guregu/null.v3"
)

type CustomBool sql.NullBool
type Data struct {
	ID      null.Int `json:"id" ch:"id"`
	IsValid CustomBool `json:"is_valid" ch:"is_valid"`
}
```

However, I get an error with this code:

```go
var res Data
rows.ScanStruct(&res)
```

```log
clickhouse [ScanRow]: (is_valid) converting Bool to *model.CustomBool is unsupported
```
**Describe the solution you'd like**
a solution like [this](https://github.com/ClickHouse/clickhouse-go/issues/742#issuecomment-1260673158) should be able to solve my issue

PR #1156, opened 2023-12-08
Support bool alias type

yogasw:
## Summary
closes https://github.com/ClickHouse/clickhouse-go/issues/1154

### DIFF
--- a/lib/column/bool.go
+++ b/lib/column/bool.go
@@ -65,12 +65,9 @@ func (col *Bool) ScanRow(dest any, row int) error {
 	case **bool:
 		*d = new(bool)
 		**d = col.row(row)
-	case *sql.NullBool:
+	case sql.Scanner:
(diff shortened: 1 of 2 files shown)


lib/column/bool.go
```diff
@@ -68,6 +68,9 @@ func (col *Bool) ScanRow(dest any, row int) error {
 	case *sql.NullBool:
 		return d.Scan(col.row(row))
 	default:
+		if scan, ok := dest.(sql.Scanner); ok {
+			return scan.Scan(col.row(row))
+		}
```
jkaflik, line 73:
```suggestion
	case sql.Scanner:
		return d.Scan(col.row(row))
	default:
```

jkaflik (commented):
There is one code improvement. I will submit it.

jkaflik (approved):

PR #1205, opened 2024-02-27
Support [n]byte/[]byte type Scan/Append to FixedString column

### LINKED ISSUE #1155: converting []uint8 to FixedString is unsupported
Would it be possible to accept `[]byte` and `[n]byte` for the `FixedString(n)` type?

It is already accepting a `encoding.BinaryMarshaler` which returns `[]byte`, like this:

```go
type binBytes []byte

func (b binBytes) MarshalBinary() ([]byte, error) {
	return b, nil
}
```

rogeryk:
## Summary
Support \[n]byte/[]byte type Scan/Append to FixedString column

## Checklist

closes #1155

### DIFF
--- a/lib/column/fixed_string.go
+++ b/lib/column/fixed_string.go
@@ -22,8 +22,9 @@ import (
 	"database/sql/driver"
 	"encoding"
 	"fmt"
-	"github.com/ClickHouse/ch-go/proto"
 	"reflect"
+
+	"github.com/ClickHouse/ch-go/proto"
 
 	"github.com/ClickHouse/clickhouse-go/v2/lib/binary"
 )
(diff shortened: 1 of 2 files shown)


lib/column/fixed_string.go
```diff
@@ -150,6 +203,8 @@ func (col *FixedString) Append(v any) (nulls []uint8, err error) {
 func (col *FixedString) AppendRow(v any) (err error) {
 	data := make([]byte, col.col.Size)
 	switch v := v.(type) {
+	case []byte:
+		copy(data, v)
```
jkaflik, line 207: Why copy instead of assign slice?

jkaflik (approved):

lib/column/fixed_string.go
rogeryk: The length of slice may not equal col.col.Size.  Use copy to cover overlap scenarios

jerbob92:
Just to let you know: this has broken our code to save [XID's](https://github.com/rs/xid) as `FixedString(20)` in CH. This is due to that the XID is actually an alias for a byte array (`type ID [rawLen]byte`), so it's now hitting the new code. XID's in byte format are 12 bytes, but in string format it's 20. Since XID has implemented the `Stringer` interface, it would normally have hit this block:

```
if s, ok := v.(fmt.Stringer); ok {
	return col.AppendRow(s.String())
}
```

But since this change it's hitting the byte array block and resulting in the following error:

```
clickhouse [AppendRow]: converting xid.ID to FixedString is unsupported. invalid size 12, expect 20
```

My suggestion would be to move the handling of the byte array to below the `Stringer` handling, imho in case of FixedString, the `Stringer` should have precedence over other handling.

rogeryk:
@jerbob92 Sounds right, I think should create a new issue and file a PR  to fix it.

jerbob92:
@rogeryk or maybe @jkaflik should revert it and you should make a new PR...

PR #1217, opened 2024-03-03
Fix column with double quotes PrepareBatch failed.(#1216)

### LINKED ISSUE #1216: Got panic: "block cannot be sorted - missing columns in requested order" when a column identifiers with double quotes
### Describe the bug
I define a table with fields that have double quotes, and when I insert a record, I get an error using backticks. This bug was introduced in clickhouse-go 2.14.3 which introduced in the PR #1083,  and has not been fixed so far. While it's not a good idea to define fields this way, I'm having this issue in production and hope it can be fixed.

### Code example
```go
package main

import (
	"context"
	"fmt"
	"log"

)

func main() {
	})
	if err != nil {
		log.Fatal(err)
	}

	ctx := context.Background()
        _ = conn.Exec(ctx, "DROP TABLE IF EXISTS example SYNC")
	createSql := "CREATE TABLE IF NOT EXISTS example (`@id` String,`\"@id_with_quotes\"` String) Engine = Memory"
	err = conn.Exec(ctx, createSql)
	if err != nil {
		log.Fatal(err)
	}

	_, err = conn.PrepareBatch(ctx, "INSERT INTO example (`@id`, `\"@id_with_quotes\"`)")
	if err != nil {
		log.Fatal(err)
	}
}

```
### Error log
```bash
2024/03/03 09:13:42 block cannot be sorted - missing columns in requested order: ["@id_with_quotes"]
exit status 1
```

#### Environment
* Client version: 22.8.8.3
* Language version: clickhouse-go2.14.3+
* OS: CentOS 7 x86_64
* Interface: ClickHouse API / database/sql compatible driver

#### ClickHouse server
* ClickHouse Server version:
* ClickHouse Server non-default settings, if any:
* `CREATE TABLE` statements for tables involved:
* Sample data for all these tables, use [clickhouse-obfuscator](https://github.com/ClickHouse/ClickHouse/blob/master/programs/obfuscator/Obfuscator.cpp#L42-L80) if necessary

YenchangChan:
## Summary
Resolves #1216 

## Checklist

### DIFF
--- a/conn_batch.go
+++ b/conn_batch.go
@@ -47,7 +47,9 @@ func (c *connect) prepareBatch(ctx context.Context, query string, opts driver.Pr
 	if len(colMatch) == 2 {
 		columns = strings.Split(colMatch[1], ",")
 		for i := range columns {
-			columns[i] = strings.Trim(strings.TrimSpace(columns[i]), "`\"")
+			// refers to https://clickhouse.com/docs/en/sql-reference/syntax#identifiers
+			// we can use identifiers with double quotes or backticks, for example: "id", `id`, but not both, like `"id"`.
+			columns[i] = strings.Trim(strings.Trim(strings.TrimSpace(columns[i]), "\""), "`")
 		}
 	}
 	if !strings.HasSuffix(strings.TrimSpace(strings.ToUpper(query)), "VALUES") {
--- a/tests/issues/1216_test.go
+++ b/tests/issues/1216_test.go
@@ -1,0 +1,31 @@
+package issues
+
+import (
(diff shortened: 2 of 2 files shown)


YenchangChan:
unit tests failed:
```
2024/03/03 01:48:35 🐳 Creating container for image testcontainers/ryuk:0.6.0
panic: look up reaper container returned nil although creation failed due to name conflict: creating reaper failed: failed to create container
```
But it doesn't look like it's caused by my code.

jkaflik (approved):

PR #1237, opened 2024-03-18
Make errors channel buffered inside query() 

### LINKED ISSUE #1229: "acquire conn timeout" issue
This issue is a little different. I'll be able to describe it, but won't be able to outline steps to reproduce it. 

### Describe the bug
In my app, I've set max open/idle connections to 25 and 20 respectively. In the beginning I had a problem that every nth query ended up returning error "acquire conn timeout: ...". I tried to fix it by setting max conn lifetime to 5s and passing context with timeout. Nevertheless I still see the same problem, less often than it used to be fortunately, but it still exists, and bothers me a little. 

### Expected behaviour
I expected to see no problem with settings above. 

## Side notes
Inserts happen in batches once in every 20s. Reads are rare.

threadedstream:
## Summary
This PR closes #1229 

While debugging this issue, I noticed the following line of code inside conn_query.go
```go
		err := c.process(ctx, onProcess)
		if err != nil {
			c.debugf("[query] process error: %v", err)
			errors <- err
```

here, err gets transferred to errors channel in case if err != nil. In my case, it happens when context gets canceled. This resulted into situation where certain goroutines couldn't pass through this errors <- err and code never got to the connection release point. So I made this errors channel a buffered one, tested it, and was satisfied with the result - connections now don't get stuck in an open pool.

### DIFF
--- a/clickhouse.go
+++ b/clickhouse.go
@@ -258,6 +258,11 @@ func (ch *clickhouse) acquire(ctx context.Context) (conn *connect, err error) {
 	timer := time.NewTimer(ch.opt.DialTimeout)
 	defer timer.Stop()
 	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	default:
+	}
+	select {
 	case <-timer.C:
 		return nil, ErrAcquireConnTimeout
 	case <-ctx.Done():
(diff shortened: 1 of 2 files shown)


jkaflik (approved):
great finding. do you want to provide a test case for this?

threadedstream:
Yeah, I'll try providing test for it asap

PR #1266, opened 2024-04-05
Add driver.CollectRows and driver.AppendRows

EpicStep:
These auxiliary functions help with rows processing, examples of use can be seen in the tests

### DIFF
--- a/lib/driver/rows.go
+++ b/lib/driver/rows.go
@@ -1,0 +1,41 @@
+package driver
+
+// CollectableRow is the subset of Rows methods that a RowToFunc is allowed to call.
+type CollectableRow interface {
+	Scan(dest ...any) error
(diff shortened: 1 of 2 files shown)


EpicStep:
Hi @jkaflik. I think auxiliary functions that I published in this PR are inappropriate inside the database driver and it's better to put them in a separate repository, what do you think?

jkaflik (commented):
I would rather defer it to a separate repository. If we want to keep this as a part of this, we need to consider it as a `util` subpackage.

EpicStep:
Okey, sounds good, I will check this PR after vacation

EpicStep:
A separate library is ready
https://github.com/EpicStep/clickhouse-go-rows-utils

Issue #1150, opened 2023-12-01
An obscure error in the HTTP protocol when using compression


vitalyisaev2:
When trying to access ClickHouse server with invalid credentials using HTTP protocol with LZ4 Compression enabled, an unclear error occurs. Consider this small example:

```go
package main

import (
	"fmt"
	"github.com/ClickHouse/clickhouse-go/v2"
)

func makeOptions(compression bool) *clickhouse.Options {
	opts := &clickhouse.Options{
		Addr: []string{"localhost:8123"},
		Auth: clickhouse.Auth{
			Database: "dqrun",
			Username: "crab",
			Password: "qwerty12345", // Wrong password
		},
		Protocol: clickhouse.HTTP,
	}

	if compression {
		opts.Compression = &clickhouse.Compression{
			Method: clickhouse.CompressionLZ4,
		}
	}

	return opts
}

func main() {
	conn1 := clickhouse.OpenDB(makeOptions(false))
	err := conn1.Ping()
	fmt.Printf("Without compression:\n%v\n\n", err)

	conn2 := clickhouse.OpenDB(makeOptions(true))
	err = conn2.Ping()
	fmt.Printf("With compression:\n%v\n\n", err)

	return
}
```

Without compression the error is straightforward:
```
clickhouse [execute]:: 403 code: Code: 516. DB::Exception: crab: Authentication failed: password is incorrect, or there is no user with such name. (AUTHENTICATION_FAILED) (version 23.4.2.11 (official build))
```

Otherwise user gets obscure error resembling protocol corruption:
```
clickhouse [execute]:: failed to read the response: read: read next block: data size should be 0 < 980316009 < 134217728
```

#### Environment
* Client version: github.com/ClickHouse/clickhouse-go 1.5.4
* Language version: go1.21.3 linux/amd64
* OS: Ubuntu 20

#### ClickHouse server
* ClickHouse Server version: 23.4.2.11

wade-liwei:
how do you resolve the problem?

PR #1230, opened 2024-03-08
Fix an obscure error in HTTP protocol with LZ4 compression

YenchangChan:
## Summary
Resolves #1150 

Now return error like:
```
clickhouse [execute]:: 403 code: failed to read the response: Code: 516. DB::Exception: default: Authentication failed: password is incorrect, or there is no user with such name.

If you have installed ClickHouse and forgot password you can reset it in the configuration file.
The password for default user is typically located at /etc/clickhouse-server/users.d/default-password.xml
and deleting this file will reset the password.
See also /etc/clickhouse-server/users.xml on the server where ClickHouse is installed.

. (AUTHENTICATION_FAILED) (version 23.8.9.54 (official build))
```

## Checklist
Delete items not relevant to your PR:
- [ ] Unit and integration tests covering the common scenarios were added
- [ ] A human-readable description of the changes was provided to include in CHANGELOG
- [ ] For significant changes, documentation in https://github.com/ClickHouse/clickhouse-docs was updated with further explanations or tutorials

PR #1286, opened 2024-04-27
Always compress responses when the client compression is on

zhkvia:
## Summary
* It resolves https://github.com/ClickHouse/clickhouse-go/issues/1150 again, since [the fix](https://github.com/ClickHouse/clickhouse-go/pull/1230) was broken at some time (`body` after `return` [is empty](https://github.com/ClickHouse/clickhouse-go/blob/main/conn_http.go#L428) since it hasn't been read yet)
* The previous fix covered only the 403 code, but this one covers all non-200
* As a bonus, with this fix the error doesn't state that it `failed to read the response` (which is technically not true since a response is successfully read)
Now it looks like:
```
clickhouse [execute]:: 403 code: Code: 516. DB::Exception: default: Authentication failed: password is incorrect, or there is no user with such name.
<and the rest>
```

### DIFF
--- a/conn_http.go
+++ b/conn_http.go
@@ -424,14 +424,14 @@ func (h *httpConnect) readRawResponse(response *http.Response) (body []byte, err
 	if err != nil {
 		return nil, err
 	}
-	if response.StatusCode == http.StatusOK && (h.compression == CompressionLZ4 || h.compression == CompressionZSTD) {
+	if h.compression == CompressionLZ4 || h.compression == CompressionZSTD {
 		chReader := chproto.NewReader(reader)
 		chReader.EnableCompression()
 		reader = chReader
 	}
 
 	body, err = io.ReadAll(reader)
-	if err != nil {
+	if err != nil && !errors.Is(err, io.EOF) {
 		return nil, err
 	}
 	return body, nil
--- a/conn_http.go
+++ b/conn_http.go
@@ -424,10 +424,7 @@ func (h *httpConnect) readRawResponse(response *http.Response) (body []byte, err
 	if err != nil {
 		return nil, err
 	}
-	if response.StatusCode == http.StatusForbidden {
-		return nil, errors.New(string(body))
-	}
-	if h.compression == CompressionLZ4 || h.compression == CompressionZSTD {
+	if response.StatusCode == http.StatusOK && (h.compression == CompressionLZ4 || h.compression == CompressionZSTD) {
 		chReader := chproto.NewReader(reader)
 		chReader.EnableCompression()
 		reader = chReader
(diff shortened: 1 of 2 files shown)


zhkvia:
Actually after applying my fix and running it for some time in production I noticed that errors returned from [a failed DESCRIBE](https://github.com/ClickHouse/clickhouse-go/blob/main/conn_http_batch.go#L54) don't get decompressed:
```
clickhouse [execute]:: 404 code: ~k��"x�
�h��`������Code: 60. DB::Exception: Table logs.�_shard does not exist. Maybe you meant0�0_trace?. (UNKNOWN_TABLE) (version 23.8.12.13 (official build))
```

So I looked at the source code more closely and noticed the difference between `DESCRIBE` and `INSERT` queries:
1. responses from `DESCRIBE` queries are always compressed:
    1. https://github.com/ClickHouse/clickhouse-go/blob/main/conn_http_batch.go#L54
    2. https://github.com/ClickHouse/clickhouse-go/blob/main/conn_http_query.go#L39
2. while responses from `INSERT` queries are not:
    1. https://github.com/ClickHouse/clickhouse-go/blob/main/conn_http_batch.go#L52
    2. https://github.com/ClickHouse/clickhouse-go/blob/main/conn_http_batch.go#L108
    3. https://github.com/ClickHouse/clickhouse-go/blob/main/conn_http_batch.go#L201

But responses are always decompressed no matter what query it was: https://github.com/ClickHouse/clickhouse-go/blob/main/conn_http.go#L432

So in order to fix this, I explicitly set `compress` option for `INSERT` queries as well

--------
More than that, I noticed that in older versions of the library `EOF` was treated explicitly because it was wrapped with other errors several times in a row:
https://github.com/ClickHouse/clickhouse-go/blob/v2.14.0/conn_http.go#L453

But now this error is not checked and `ReadAll` can't handle it by itself since it doesn't look inside wrapped errors:
https://github.com/ClickHouse/clickhouse-go/blob/main/conn_http.go#L436

So I added this too

jkaflik (approved):

zhkvia:
```@github-actions
run-tests / single-node (1.21, 24.1) (pull_request) Failing after 1m
```

Well, I couldn't reproduce the same results:
```
 λ  CLICKHOUSE_VERSION=24.1 go test .
ok      github.com/ClickHouse/clickhouse-go/v2  0.392s
 λ  cd tests/issues
 λ  CLICKHOUSE_VERSION=24.1 go test .
ok      github.com/ClickHouse/clickhouse-go/v2/tests/issues     0.565s
 λ  go version
go version go1.21.9 darwin/amd64
```

PR #1291, opened 2024-05-03
Don't recreate keys from LC columns from direct stream

genzgd:
## Summary
Fixes an issue with low cardinality columns with a large number of distinct values when using experimental SELECT streaming

## Checklist


jkaflik (approved):
🍻

PR #1317, opened 2024-06-02
Make custom debug log function on-par with the built-in one

vespian:
## Summary
The built-in debug-log function was printing more information than the custom one could. This PR makes them on-par by providing the wrapper.

I think this can be considered backwards-compatible, but I would be happy to hear your opinion on this. This is just a debug message after all and we just prefix it with extra data.


vespian:
@jkaflik PTAL :pray:

jkaflik (approved):
It's not fully backward compatible, but given impact is low we can release it as next minor version.

@vespian please take a look on our CLA agreement.

vespian:
@jkaflik I have run tests locally and they passed. I am unfamiliar with your testing setup and whether the failures here are legit. 

Is there anything I could help here with?

jkaflik:
No worries. We have CI runners instability - I will take care of it.

PR #1373, opened 2024-08-15
Fix column list parsing for multiline INSERT statements

Fiery-Fenix:
## Summary
This PR is fixing an issue when multiline INSERT statement is used and driver is using list of columns from ClickHouse table instead of provided in statement list of columns
As a result of this bug error "clickhouse: expected N arguments, got M" produced when columns count mismatch in ClickHouse table and INSERT statement

## Checklist

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -26,9 +26,9 @@ import (
 )
 
 var normalizeInsertQueryMatch = regexp.MustCompile(`(?i)(INSERT\s+INTO\s+([^(]+)(?:\s*\([^()]*(?:\([^()]*\)[^()]*)*\))?)(?:\s*VALUES)?`)
-var truncateFormat = regexp.MustCompile(`\sFORMAT\s+[^\s]+`)
+var truncateFormat = regexp.MustCompile(`(?i)\sFORMAT\s+[^\s]+`)
 var truncateValues = regexp.MustCompile(`\sVALUES\s.*$`)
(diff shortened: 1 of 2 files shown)


SpencerTorres (approved):
Thanks for finding this. It looks like this fix works by enabling newlines in the regex?

It would be good to add another test that mixes inline with newlines just to validate there's no quirks for poorly formatted queries

Fiery-Fenix:
> Thanks for finding this. It looks like this fix works by enabling newlines in the regex?
> 
> It would be good to add another test that mixes inline with newlines just to validate there's no quirks for poorly formatted queries

Thanks for suggestion, by adding few additional test cases I was able to identify another bug with lowercase statements ;)
Now this PR is:
* Fixing bug with multiline INSERT statements
* Fixing bug with lowercase INSERT statements
* Adding few additional test cases to cover previous issues in different combinations

SpencerTorres:
Awesome! Glad to hear it handles more cases now. Regex makes me nervous (especially for parsing language syntax) so it's great to have those extra cases in there. Thanks for the update!

jkaflik (approved):

jkaflik:
@Fiery-Fenix, thanks for your contribution. Just for a record, this regression has been introduced in https://github.com/ClickHouse/clickhouse-go/pull/1341

Issue #1388, opened 2024-08-27
There's has no way to cancel a running query using context's cancel functions.


tinybit:
## Observed

Suppose we have a query that takes more than 10s to run.

Example 1: query will timeout after 10s, even though `cancel()` is called after 5s.
```
ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)

go func() {
   <...>
   err = conn.Exec(ctx, query)
   <...>
}

go func() {
    time.Sleep(5 * time.Second)
    cancel()
}()
```

Example 2: query will timeout after 5m, even though `cancel()` is called after 5s.
```
ctx, cancel := context.WithCancel(context.Background())

go func() {
   <...>
   err = conn.Exec(ctx, query)
   <...>
}

go func() {
    time.Sleep(5 * time.Second)
    cancel()
}()
```

## Expected behaviour
Stop query execution immediately after context was cancelled.

## Code example

```
package main

import (
	"context"
	"crypto/tls"
	"fmt"
	"log"
	"time"

	"github.com/ClickHouse/clickhouse-go/v2"
)

func main() {
	// connect
	options := &clickhouse.Options{
		Addr: []string{"o2q8185ekk.us-east-2.aws.clickhouse-staging.com:9440"},
		Auth: clickhouse.Auth{
			Database: "default",
			Username: "default",
			Password: "pg0H6o0ikVyJFbu9KgYKXTII1u5HHdLLsvCg18MI7rSsf4ukLFUZcNzMoMjFfQaY",
		},
		Compression: &clickhouse.Compression{
			Method: clickhouse.CompressionLZ4,
		},
		TLS: &tls.Config{
			InsecureSkipVerify: false,
		},
	}

	conn, err := clickhouse.Open(options)
	if err != nil {
		return
	}

	if err = conn.Ping(context.Background()); err != nil {
		return
	}

	log.Println("Connected.")

	// prepare context
	ctx, cancelCtx := context.WithCancel(context.Background())
	defer cancelCtx()

	doneCh := make(chan bool, 1)

	// run query in background
	go func() {
		defer func() { doneCh <- true }()
		log.Println("Running query that will take 3s to execute...")

		err = conn.Exec(ctx, "SELECT sleep(3)")
		if err != nil {
			log.Printf("Finished with error: %v\n", err)
			conn.Close()
			return
		}

		fmt.Println("Finished ok.")
	}()

	// let workers run for awhile and stop
	go func() {
[...]

PR #1389, opened 2024-08-27
Add ability to handle context cancellations for TCP protocol

tinybit:
## Summary
Issue:
- https://github.com/ClickHouse/clickhouse-go/issues/1388

## Checklist


conn.go
```diff
@@ -218,6 +237,18 @@ func (c *connect) compressBuffer(start int) error {
 }
 
 func (c *connect) sendData(block *proto.Block, name string) error {
+	if c.closed {
+		err := errors.New("attempted sending on closed connection")
+		c.debugf("[send data] err: %v", err)
+		return err
+	}
+
+	if c.buffer == nil {
```
jkaflik, line 246: Is it something you run into when debugging or it's something you want to presumably safe?

conn.go
```diff
@@ -131,6 +139,8 @@ type connect struct {
 	readTimeout          time.Duration
 	blockBufferSize      uint8
 	maxCompressionBuffer int
+	mutex                sync.Mutex
```
jkaflik, line 142: do you think it makes sense to name it explicit?
```suggestion
	readerMutex                sync.Mutex
```

conn.go
```diff
@@ -218,6 +255,18 @@ func (c *connect) compressBuffer(start int) error {
...
+		err := errors.New("attempted sending on closed connection")
+		c.debugf("[send data] err: %v", err)
+		return err
+	}
+
+	if c.buffer == nil {
+		err := errors.New("attempted sending on nil buffer")
+		c.debugf("[send data] err: %v", err)
+		return err
+	}
```
jkaflik, line 268: is it something you found when while debugging or only added for safety?

jkaflik (commented):
LGTM. I will do a small testing on my side and then we can merge.

conn.go
tinybit: yeah, good idea

conn.go
```diff
@@ -218,6 +237,18 @@ func (c *connect) compressBuffer(start int) error {
...
```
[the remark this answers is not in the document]
tinybit: yup

conn.go
tinybit: yup

conn.go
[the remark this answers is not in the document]
tinybit: hmm, this is outdated code, please check the latest changes
https://github.com/ClickHouse/clickhouse-go/pull/1389/files

conn.go
tinybit: done

conn.go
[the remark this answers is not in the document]
tinybit: also, safety is not presumed, it's necessary here.
previously, in conn_process.go access to c.close and c.buffer was single-threaded.
now i'm reading data in background goroutine and in foreground thread i'm waiting on select:

``` go
// do reads in background
errCh := make(chan error, 1)
doneCh := make(chan bool, 1)

go func() {
	err := c.processImpl(ctx, on) // this accesses c.reader internally (reads data)
	if err != nil {
		errCh <- err
		return
	}

	doneCh <- true
}()

// select on context or read channel (errors)
select {
case <-ctx.Done():
	c.cancel() // calls c.close(), it accesses c.reader internally (sets c.reader to nil)
	return ctx.Err()

case err := <-errCh:
	return err

case <-doneCh:
	return nil
}
```

1) Closing connections must be done very carefully. Currently **conn.Close()** is broken because multiple calls to **conn.Close()** have issues deadlocking the thread it was called from

2) Also multiple calls to **c.close()** from multiple threads is not thread safe, doing so will result in data races.

So, this needed to be fixed from this:
```
func (c *connect) close() error {
    if c.closed {
        return nil
    }
    c.closed = true
<...>
```

To a thread-safe version of it
```
func (c *connect) close() error {
	c.closeMutex.Lock()
	if c.closed {
		c.closeMutex.Unlock()
		return nil
	}
	c.closed = true
	c.closeMutex.Unlock()
<...>
```

This guarantees that under no circumstances **c.close()** can be called twice no matter what.

3) Besides this we need to guarantee threadsafe access to **c.reader** from **c.processImpl()**, **c.cancel()** and  **c.close()**, hence I've added **c.readerMutex()**

4) One more thing: the order of operations during closing a connection:
Old version:
```
func (c *connect) close() error {
	c.closeMutex.Lock()
	if c.closed {
		c.closeMutex.Unlock()
		return nil
	}
	c.closed = true
	c.closeMutex.Unlock()

	c.buffer = nil

	c.readerMutex.Lock()
	c.reader = nil
	c.readerMutex.Unlock()

	if err := c.conn.Close(); err != nil {
		return err
	}
	
	return nil
}
```

As you can see here, we're changing **c.buffer** and **c.reader** that might be currently reading/sending data, before we close the network connection. Connection must be closed first, in order to stop any i/o operations on **c.buffer** and **c.reader** (and related contexts), then it's safe to deallocale **c.buffer** and **c.reader**.

```
func (c *connect) close() error {
	c.closeMutex.Lock()
	if c.closed {
		c.closeMutex.Unlock()
		return nil
	}
	c.closed = true
	c.closeMutex.Unlock()

	if err := c.conn.Close(); err != nil {
		return err
	}

	c.buffer = nil

	c.readerMutex.Lock()
	c.reader = nil
	c.readerMutex.Unlock()

	
	return nil
}
```

jkaflik:
I conducted basic tests. This will be merged with next minor release in two weeks.

PR #1417, opened 2024-10-09
Default implementation of column.IterableOrderedMap

earwin:
## Summary
As discussed in #1402: most of column.IterableOrderedMap use cases can be covered by a simple default implementation.
This pull request aims to provide the implementation.


jkaflik (commented):
Thanks @earwin ! Could you provide a minimal unit test for your implementation?

PR #1426, opened 2024-10-17
[improvement] Some performance related changes to evaluate

mdonkers:
## Summary

Based on some CPU and memory analysis we ran on our production systems, we noticed a few things that could be improved.

I'd like to present these and discuss whether it makes sense to merge these into the project (otherwise we'll likely keep these on a local fork). Also some changes perhaps need a bit of work or could be done in another way. All feedback welcome!

### Allow String column provider to take the name, so that it can use different settings

For `String` column types its possible to define a `ColStrProvider`, but this can only be defined once and globally. While `ColStrProvider` is great for pre-sizing buffers when creating batches, you can imagine not every column needs as much space. Some may be limited to a few characters, while others (e.g. log bodies) can be hundreds of characters.
Adding the name of the column as input to the `ColStrProvider` allows the internal logic to size buffers differently, or maybe include other logic.
This change does not affect users of the API.

### Prevent hashing enums just for checking validity

We are creating quite big batches, and have quite a few `enum`s in our records. In CPU profiles we noticed a lot of time was spent in map lookups, just for adding the enum to the batch. This is done to check whether the enum is valid, since they don't need to be a continuous range in ClickHouse.

The optimization takes a bit of overhead up-front, by checking if the Enum definition _is_ a continuous range and capturing the lower and upper bound. Because if continuous, then for validating we _only_ need to check if the number falls between lower and upper bound. This simple boolean logic is much faster overall than the map lookup.

## Checklist


mdonkers:
@jkaflik by the way I don't expect the PR to be ready for merging in its current state.
But wanted to discuss on the approach etc before spending a lot of time. Everything _works_ and compiles though.

jkaflik:
@SpencerTorres could you take a look please?

lib/column/enum.go
```diff
@@ -20,10 +20,11 @@ package column
 import (
 	"bytes"
 	"errors"
+	"github.com/ClickHouse/ch-go/proto"
```
serprex, line 23: don't mix non-std imports with std imports

lib/column/enum.go
```diff
@@ -47,6 +48,9 @@ func Enum(chType Type, name string) (Interface, error) {
 			enum.iv[values[i]] = proto.Enum8(v)
 			enum.vi[proto.Enum8(v)] = values[i]
 		}
+		enum.minEnum = int8(slices.Min(maps.Keys(enum.vi)))
+		enum.maxEnum = int8(slices.Max(maps.Keys(enum.vi)))
+		enum.continuous = (enum.maxEnum-enum.minEnum)+1 == int8(len(enum.vi))
```
serprex, line 53: For Enum8 non-continuous & continous could both use a 256 bitmap (only 32 bytes, could skip heap by being in struct as `[4]uint64`)

mdonkers: Not sure I understand your suggestion. Do you have some (semi-code) example of how you would use that, and which library? (Since I don't know of any bitmap implementations in the Golang standard library itself.)

lib/column/enum.go
```diff
@@ -47,6 +48,9 @@ func Enum(chType Type, name string) (Interface, error) {
 			enum.iv[values[i]] = proto.Enum8(v)
 			enum.vi[proto.Enum8(v)] = values[i]
 		}
+		enum.minEnum = int8(slices.Min(maps.Keys(enum.vi)))
```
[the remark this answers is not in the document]
mdonkers: Agree, I can make those changes if the solution in general is accepted (something I wasn't sure about).

lib/column/tuple.go
```diff
@@ -566,6 +570,23 @@ func (col *Tuple) AppendRow(v any) error {
 		return nil
 	}
 
+	if tuple2, ok := v.(Tuple2); ok {
```
serprex, line 573: Have you tried adding a case to switch above this for `reflect.Array`? Could work to have `v` be `[2]any`, tho I'm doubtful it'll ever be very good having to go through `reflect`

lib/column/tuple.go
```diff
@@ -566,6 +570,23 @@ func (col *Tuple) AppendRow(v any) error {
...
+		if 2 != len(col.columns) {
+			return &Error{
+				ColumnType: string(col.chType),
+				Err:        fmt.Errorf("invalid size. expected %d got %d", len(col.columns), 2),
```
serprex, line 577:
```suggestion
				Err:        fmt.Errorf("invalid size. expected %d, got %d", len(col.columns), 2),
```

lib/column/tuple.go
serprex, line 573: `switch valuer := v.(type)` since we're checking between Tuple2/driver.Valuer/default

lib/column/tuple.go
```diff
@@ -39,6 +39,10 @@ type Tuple struct {
 	index   map[string]int // map from col name to offset in columns
 }
 
+type Tuple2 interface {
+	Get() (any, any)
```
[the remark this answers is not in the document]
mdonkers: Yeah, this is probably the least meaningful improvement, so happy to drop it.

lib/column/enum.go
serprex: No need to use library, can be done in loop above, something like
```
var bitset [4]uint64
for i := range values {
  bitIndex := uint8(indexes[i])
  bitset[bitIndex>>3] |= 1 << (bitIndex&7)
}
```

lib/column/enum.go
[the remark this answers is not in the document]
serprex: just noticed `range values` loop above, can calc min/max in there

SpencerTorres (changes requested):
I appreciate your effort in finding these optimizations along with a good explanation for each.

@serprex found most of the major concerns I would say. Here's my summary after reviewing each file:
- The String column provider change is great, very convenient for supplying your own function and switching on column name.
- The enum changes are good too, but I agree with @serprex on the bitmap solution. It's somewhat obscure, but once they're wrapped in functions it should be clear what is happening. This seems like it would be the fastest and most accurate approach for checking enum validity. The implementation would need to be adjusted for each enum size too (8 vs 16).
- Let's remove the Tuple change for now. I like the idea of having optimized definitons for common sizes, but I don't want to hold back the other optimizations in this PR while we decide on that implementation. Maybe we can move it to an issue for discussion?
- And then finally once everything is updated/finalized, some good test coverage (especially for the enums.)

Thanks!

lib/column/enum.go
mdonkers: Gave this solution some thought, but I think its defeating the purpose. The whole idea was to skip the map lookup (`col.vi[v]`) which is the thing causing lots of overhead.
Using the bitmap as you proposed, it still needs to do a map lookup: `bitset[bitIndex>>6]`. I think whether the map appears on the heap or not doesn't matter.

If you think it does, I can try creating a benchmark and compare the different solutions. But would need to find a bit of time to put in that effort.

lib/column/enum.go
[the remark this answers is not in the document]
mdonkers: Changed this for `Enum16`, but will wait a bit for the `Enum8` case for a response here: https://github.com/ClickHouse/clickhouse-go/pull/1426/files#r1905195485

lib/column/enum.go
SpencerTorres: I believe the expense on the CPU would be different between a standard map type and the bitmap. It should be similarly as fast as the integer bounds check. 

If the solution works as-is then I'm okay with keeping it, if it's invalid then the server will throw an error. Most clients are already fairly in sync with their data model, especially with enums. I wouldn't want to hold the PR back due to this, we can iterate on it later

serprex: `[4]uint64` isn't a map. Array access will as fast as field access for on continuous/min/max

mdonkers: :man_facepalming: of course. Somehow I had a map in my mind due to the bit**map**

I'll make the changes accordingly and verify

mdonkers:
@SpencerTorres @serprex Could you have another look please? I think I addressed all comments and also added some tests to check the `enum8` usage with the bitset for checking existence.

serprex (approved):

mdonkers:
Separated the `Tuple` improvement proposal to https://github.com/ClickHouse/clickhouse-go/issues/1473

SpencerTorres (approved):
Thanks for updating!

srikanthccv:
Hi @SpencerTorres / @mdonkers, can you also please update the template `lib/column/codegen/column.tpl` for `case "String"` to use `name` and regenerate. The `name` is missing in the template, resulting in invalid code with `go run lib/column/codegen/main.go`

Issue #1195, opened 2024-02-16
Add support for experimental Variant data type


mshustov:
[The type is experimental]. See the docs https://clickhouse.com/docs/en/sql-reference/data-types/variant
There will be more added for better support of the semistructured data https://github.com/ClickHouse/ClickHouse/issues/54864

Gictorbit:
I also have this error when using `Variant`:
```sh
assertion failed: error is not nil: clickhouse: unsupported column type "Variant(Bool, Float64, String, UInt64)"
```

PR #1453, opened 2024-12-20
Add Variant Type

### LINKED ISSUE #1430: Add support for new JSON type
Scope:
- [x] Variant type https://github.com/ClickHouse/clickhouse-go/pull/1453
- [x] Dynamic type https://github.com/ClickHouse/clickhouse-go/pull/1454
- [x] JSON type https://github.com/ClickHouse/clickhouse-go/pull/1455
- [ ] Deprecation of `Object('json')`
- [ ] performance benchmarking for string- and binary-based implementation

from @SpencerTorres  preliminary research

What works:
- Inserting JSON as Go `string`
- Selecting `JSON` when it is first cast to a string (`::String`). Must be scanned into a Go `string` type.
- Selecting `Dynamic` column (from within a `JSON` object) when it is first cast to a string (`::String`). Must be scanned into a Go `string` type.

What doesn't work:
- Inserting as Go `json.RawMessage` (`[]byte`) does not work (although I don't believe this works for `String` either)
- Selecting does not work for non-casted `JSON` type column, the Go driver reports it as an unsupported type
- Selecting does not work for non-casted `Dynamic` type column, the Go driver reports it as an unsupported type

Sample SQL:
```sql
SET allow_experimental_json_type = 1;

CREATE TABLE json_test (id String, obj JSON) Engine = MergeTree ORDER BY id;

INSERT INTO json_test (id, obj) VALUES ('a', '{ "key1": "value1", "nestedKey2": { "nestedValue2": 5 }, "arrayKey3": [4, 5, 6] }')

SELECT id, obj.key1, obj.nestedKey2, obj.arrayKey3, obj.nestedKey2.nestedValue2, obj FROM json_test;
```

Sample code:
```go
// regular struct
type jsonRow struct {
	ID  string `ch:"id"`
	Obj string `ch:"obj"`
}

// row to insert, note that the JSON data is a `string`
var insertRow = jsonRow{
	ID:  "c",
	Obj: `{ "key1": "value1", "nestedKey2": { "nestedValue2": 5 }, "arrayKey3": [4, 5, 6] }`,
}

// async insert, blocking
err = db.AsyncInsert(context.Background(), "INSERT INTO json_test (id, obj) VALUES (?, ?)", true, insertRow.ID, insertRow.Obj)
if err != nil {
	fmt.Println(err)
	return
}

// This is where things break. The JSON column MUST be cast to a string here.
// Also renamed to fit struct tags.
// If the JSON column is not cast to string, the following error is printed: clickhouse: unsupported column type "JSON"
var rows []jsonRow
err = db.Select(context.Background(), &rows, "SELECT id, obj::String as obj FROM json_test")
if err != nil {
	fmt.Println(err)
	return
}

// Print object
fmt.Printf("%+v\n", rows)
// Output (formatted):
// [{
// 	ID: c
// 	Obj: {"arrayKey3":["4","5","6"],"key1":"value1","nestedKey2":{"nestedValue2":"5"}}
// }]
```

Example of selecting an object key in Grafana (which uses our latest Go driver):
grafana object select key

SpencerTorres:
## Summary
Implement `Variant` column type. Partially resolves #1430.
closes https://github.com/ClickHouse/clickhouse-go/issues/1195

# Implementation

This implementation adds 2 major types to the module:
- `column.Variant` - the column implementation for (de)serialization
- `clickhouse.Variant` - a container to hold variant values (optional for (de)serialization). This type also has the ability to provide a preferred type in cases where it is ambiguous to existing column type detection (such as `Array(UInt8)` vs `String`)

#### Serialization

```go
// Variant(Array(Map(String, String)), Array(UInt8), Bool, Int64, String)
batch, err := conn.PrepareBatch(ctx, "INSERT INTO test_variant (c)")
require.NoError(t, err)
require.NoError(t, batch.Append(int64(42))) // Accepts primitives
require.NoError(t, batch.Append(clickhouse.NewVariantWithType("test", "String"))) // Accepts Variants with type preference
require.NoError(t, batch.Append(true))
require.NoError(t, batch.Append(clickhouse.NewVariant([]uint8{0xA, 0xB, 0xC}).WithType("Array(UInt8)"))) 
require.NoError(t, batch.Append(nil)) // Accepts nil
require.NoError(t, batch.Append([]map[string]string{{"key1": "val1"}, {"key2": "val2"}})) // Accepts complex types
```

When values are appended via `col.AppendRow()`, the input `v interface{}` type is checked. If it is `nil`, a `Null` discriminator is appended. If it is a `clickhouse.Variant` with a preferred type, then the specified column type will be appended along with its matching discriminator. The underlying column's `AppendRow` function is re-used so that we don't need to re-implement its logic.

As a catch-all, the input value will be tested against each column type until it succeeds. For example, `Variant(Bool, Int64, String)` will try to append as `bool`, `int64`, then `string`. If a value does not fit into any column type, it will return an error.

Sometimes types will conflict. Due to alphabetical sorting of the type, `Array(UInt8)` would be used before `String` since `Array` allows for `string` input. I have researched different solutions to this, including a [type priority system](https://github.com/ClickHouse/ClickHouse/blob/1aceb608f3863822277fed9f34021178118e4c5b/src/DataTypes/Serializations/SerializationVariant.cpp#L671-L815), but it would be complex to implement. For now it is easiest to let the user simply input `NewVariantWithType(int64(42), "Int64")` or `NewVariant(int64(42)).WithType("Int64")` if they want a specific type within the variant. For complex types like maps, reflection will be used if a type isn't specified.

After all rows are appended, the Native format is used to serialize the data into the buffer. First with `serializationVersion`, then the `uint8` array for `discriminators`, then each column's `Encode` function is re-used as usual (similar to `Tuple`).

#### Deserialization

The Native format deserializes the `discriminators` and builds a set of `offsets` for each column. This allows for storing multiple columns with mixed lengths. When the user wants to read a row, we can index into the correct row of each column to get the corresponding type.

In practice this looks like this:

```go
var row clickhouse.Variant // Scan into variant

require.True(t, rows.Next())
err = rows.Scan(&row)
require.NoError(t, err)
require.Equal(t, int64(42), row.Any())
```

Or, if you know your types ahead of time, you can also scan directly into it:
```go
var i int64 // Scan directly into int64
require.True(t, rows.Next())
err = rows.Scan(&i)
require.NoError(t, err)
require.Equal(t, int64(84), i)
```

This pattern works by simply calling the underlying column's `ScanRow` function. It is safest to scan into `Variant` however.
If you need to switch types on `Variant` for your own type detection, you can use `variantRow.Any()` to return `any`.
You can also switch on ClickHouse type strings by using `variantRow.Type()`.

### `clickhouse.Variant`

`clickhouse.Variant` is simply a wrapper around `any` that also includes an optional `string` to represent the preferred ClickHouse type. It implements stdlib sql interfaces such as `driver.Value` and `Scan`. If you need to access the underlying value you can use `Any()`. This type can be constructed with the `clickhouse.NewVariant(v)` function, or with `clickhouse.NewVariantWithType(v, "Array(String)")` if you need to provide a preferred type.

The `clickhouse.Variant` type should be used in structs and when scanning from `column.Variant`. It can also be used for insertion, although a variant type may be required if there's overlap between types.

You can use the preferred type for insertion when the Variant column has types that overlap. For example if you had `Variant(Array(UInt8), String)`, a Go `string` would be inserted as an `Array(UInt8)`. If you wanted to force this to be a ClickHouse `String`, you could use `clickhouse.NewVariantWithType(v, "String")` to provide the preferred type. If the preferred type is not present in the Variant, the row will fail to append to the block. Types can be added on an existing `Variant` by calling `exampleVariant.WithType(t string)`, which will return a new `clickhouse.Variant` with the preferred type set.

## Checklist


lib/chcol/variant.go
```diff
@@ -0,0 +1,126 @@
...
+
+// Any returns the underlying value as any. Same as Interface.
+func (v Variant) Any() any {
+	return v.value
+}
+
+// Interface returns the underlying value as interface{}. Same as Any.
+func (v Variant) Interface() interface{} {
+	return v.value
+}
```
jkaflik, line 48: What is the point? `interface{}` is an alias to `any`.

lib/chcol/variant.go
```diff
@@ -0,0 +1,126 @@
...
+
+// Bool returns the value as an bool if possible
+func (v Variant) Bool() (bool, bool) {
+	if b, ok := v.value.(bool); ok {
+		return b, true
+	}
+
+	return false, false
+}
```
jkaflik, line 84: I just wonder if we need these functions at all. See:

```go
package main

import "fmt"

type Variant struct {
	value any
}

func (v *Variant) Value() any {
	return v.value
}

func main() {
	v := Variant{}
	i, ok := v.Value().(int)
	fmt.Println(i, ok)
}
```

We can just expose value and let the user freely type assert.

jkaflik: More: If we make `value` public, we don't need any value access functions.

lib/column/variant.go
```diff
@@ -0,0 +1,385 @@
...
+			Op:   "Append",
+			To:   string(c.chType),
+			From: fmt.Sprintf("%T", v),
+		}
+	}
+}
+
+func (c *Variant) AppendRow(v any) error {
+	var requestedType string
+	switch v.(type) {
```
jkaflik, line 244:
```suggestion
	switch vv := v.(type) {
```

This will give you a type asserted value ready to use inside case blocks without the need for additional type assertions.

lib/column/variant.go
```diff
@@ -0,0 +1,385 @@
...
+	}
+}
+
+func (c *Variant) decodeHeader(reader *proto.Reader) error {
+	variantSerializationVersion, err := reader.UInt64()
+	if err != nil {
+		return fmt.Errorf("failed to read variant discriminator version: %w", err)
+	} else if variantSerializationVersion != SupportedVariantSerializationVersion {
+		return fmt.Errorf("unsupported variant discriminator version: %d", variantSerializationVersion)
+	}
```
jkaflik, line 337: nit 
```suggestion
	if err != nil {
		return fmt.Errorf("failed to read variant discriminator version: %w", err)
	} 
	
	if variantSerializationVersion != SupportedVariantSerializationVersion {
		return fmt.Errorf("unsupported variant discriminator version: %d", variantSerializationVersion)
	}
```

IMO more readable

lib/column/variant_test.go
```diff
@@ -0,0 +1,91 @@
...
+		require.Equal(t, c.typ, col.chType)
+		require.Equal(t, len(c.expectedTypes), len(col.columns))
+		require.Equal(t, len(c.expectedTypes), len(col.columnTypeIndex))
+
+		for j, subCol := range col.columns {
+			expectedType := c.expectedTypes[j]
+			actualType := subCol.Type()
+			if actualType != expectedType {
+				t.Fatalf("case index %d Variant type index %d column type does not match: expected: \"%s\" actual: \"%s\"", i, j, expectedType, actualType)
+			}
```
jkaflik, line 36: why not testify assert function?

tests/variant_test.go
```diff
@@ -0,0 +1,247 @@
...
+
+	if !CheckMinServerServerVersion(conn, 24, 1, 0) {
+		t.Skip(fmt.Errorf("unsupported clickhouse version for Variant type"))
+		return nil
+	}
+
+	return conn
+}
+
+func TestVariant(t *testing.T) {
```
jkaflik, line 50: Is there a need to test Variant:

- stdlib driver?
- with HTTP protocol? Do you expect any differences?

jkaflik (commented):
1st review

lib/chcol/variant.go
SpencerTorres: I put both so users could choose whichever name they prefer within their application. Some apps prefer using `any` and others `interface{}`

lib/chcol/variant.go
SpencerTorres: I'm considering removing `Variant` in favor of `VariantWithType` as the default. There are many cases where the ClickHouse type needs to be provided. If `value` were public then it doesn't make this type any different from a regular `any`.

For now I agree it may be best to simply make `value` public

lib/column/variant.go
SpencerTorres: Ah I've seen this elsewhere in the code, good point. I can simplify this switch

lib/column/variant_test.go
SpencerTorres: In this case I wanted to be able to specify an index via `%d`. I can check the testify asserts to see if there's a function for this

lib/chcol/variant.go
jkaflik: It's only a name. Why have both functions?

jkaflik (approved):
Overall LGTM. Let's push ongoing conversations before merging.

lib/chcol/variant.go
SpencerTorres: Deleted `Interface` in favor of `Any`

lib/chcol/variant.go
SpencerTorres: I have merged `VariantWithType` and `Variant` since it makes the usage more consistent for reading/writing.

`Value` being public conflicts with `Value()` which is required by `driver.Valuer` interface, I could rename it to something else but I like keeping it private since I merged the types

tests/variant_test.go
SpencerTorres: Good call, found an issue with scanning. I'll add this to the other PRs too 👍

lib/column/variant_test.go
SpencerTorres: updated all tests to use testify functions

SpencerTorres:
Updated to resolve all review notes. Thanks! 👍 

I will apply these changes to the other PRs too

PR #1455, opened 2024-12-20
Add JSON Type

### LINKED ISSUE #1430: Add support for new JSON type

SpencerTorres:
*Note: There are some server-side bugs that need to be fixed/merged, including a protocol update. However, this PR can still be reviewed as is.*

## Summary
Implement `JSON` column type. Fully resolves #1430.

## Usage

The user can use JSON in 3 ways:
1. By using the new `clickhouse.JSON` type to assemble JSON data by path strings.
2. By using a struct or map. The struct can be any nested structure, but the map must be `map[string]any`.
3. By using strings. In this mode the user handles their own `json.(Un)Marshal` and the strings get sent to the server as-is.

The JSON column also supports the `clickhouse.Dynamic` type for handling a path with multiple data types. Structs can also include a tag to specify type, such as `json:"my_field" chType:"Array(String)"`.

See examples folder for sample code.

## Checklist


SpencerTorres:
Updated with fixes and suggestions from Variant PR. There are still multiple server side bugs that will prevent certain features from working, but common scenarios work such as:
- Read/Write JSON using key/value pairs
- Write JSON using strings
- Read/Write JSON using structs (except stdlib sql)


Once #1454 is merged then we can see if any tests are broken in the test matrix. Tests are passing on latest Go/ClickHouse versions.

SpencerTorres:
Looks like the tests are passing against main.

Note that JSON string tests are skipped while we wait for: https://github.com/ClickHouse/ClickHouse/pull/73924

@jkaflik ready for review

jkaflik (approved):
LGTM. Lets have it merged and iterate in next steps.

PR #1463, opened 2025-01-02
reduced number of docker containers 

romanchechyotkin:


jkaflik:
This will not reduce the number of containers. There will be run one per package, same as it works now.

Issue #1477, opened 2025-01-21
JSON tags without name override inserts/selects wrong data


matevzmihalic:
## Observed

Inserting a struct to JSON field where struct fields have json tag formatted as `json:",omitempty"` uses empty string as field name.

## Expected behaviour

Use the actual field name if the name isn't overridden in json tag.

## Code example

Replace this struct in [/examples/clickhouse_api/json_structs.go](https://github.com/ClickHouse/clickhouse-go/blob/main/examples/clickhouse_api/json_structs.go):

```go
type ProductPricing struct {
	Price    int64  `json:",omitempty"`
	Currency string `json:",omitempty"`
}
```

Running `TestJSONStructExample` will fail with `clickhouse []: error filling nested struct: fillStruct failed to scan dynamic path: failed to scan Dynamic column into dynamic path "pricing.": clickhouse [ScanRow]: converting String to *int64 is unsupported`

### Environment
* `clickhouse-go` version: main

SpencerTorres:
Thanks for the contribution! Glad to see people trying out the new JSON type

PR #1478, opened 2025-01-21
Handle json tag without name override

matevzmihalic:
## Summary
A field in struct that has a json tag without field name override eg.`json:",omitempty"` should use the actual field name instead of empty string.

Issue #1477 

## Checklist

### DIFF
--- a/examples/clickhouse_api/json_structs.go
+++ b/examples/clickhouse_api/json_structs.go
@@ -21,13 +21,14 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/ClickHouse/clickhouse-go/v2"
 	"time"
+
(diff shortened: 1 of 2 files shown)


SpencerTorres (approved):
Looks good, I think this change will cover the common cases in the stdlib docs:

```go
// Field appears in JSON as key "myName".
Field int `json:"myName"`

// Field appears in JSON as key "myName" and
// the field is omitted from the object if its value is empty,
// as defined above.
Field int `json:"myName,omitempty"`

// Field appears in JSON as key "Field" (the default), but
// the field is skipped if empty.
// Note the leading comma.
Field int `json:",omitempty"`

// Field is ignored by this package.
Field int `json:"-"`

// Field appears in JSON as key "-".
Field int `json:"-,"`
```

Issue #1468, opened 2025-01-09
why ExceptionWhileProcessing didn't return err


btyABC:
### Question

i am using go client "github.com/ClickHouse/clickhouse-go/v2",

this is the go code:
```
	deadlineCtx, cancel := context.WithDeadline(context.Background(), time.Now().Add(queryTimeout))
	defer cancel()
	rows, err := ch.Query(deadlineCtx, sql)
	if err != nil {
		return 0, err
	}
	defer rows.Close()
```
ch.Query didn't  return err, but there are 0 rows in return.
I am sure query result of sql is not null.

this is the records in system.query_log
<html>
<body>
<html xmlns:v="urn:schemas-microsoft-com:vml" xmlns:o="urn:schemas-microsoft-com:office:office" xmlns:x="urn:schemas-microsoft-com:office:excel" xmlns="http://www.w3.org/TR/REC-html40">
<head>

<meta name=Generator content="Microsoft Excel">

<style>

</style>
</head>
<body><div ksdocclipboard="ksDocClipboardId:'{a5b80668-646d-ac84-17a6-5d0c71e796c7-358894163888}';from:'et';priorityFormat:'';mimetypes:'text/plain;text/html;WPS Spreadsheets 6.0 Format';dataType:'';srcRange:'sheetIdx:0,rowFrom:0,rowTo:14,colFrom:0,colTo:8,objSheet:94875357210048';ext:'eyJ0eXBlIjoidXBkYXRlUmFuZ2VzIiwiZmlsZUlkIjoiMzU4ODk0MTYzODg4IiwibGlua0lkIjoiY3JoaThnOW4zZTU5IiwicmdzIjpbeyJzaGVldElkeCI6MCwicm93RnJvbSI6MCwicm93VG8iOjE0LCJjb2xGcm9tIjowLCJjb2xUbyI6OCwib2JqU2hlZXQiOjk0ODc1MzU3MjEwMDQ4LCJzaGVldFN0SWQiOjEsInV1aWQiOiJleUp5WmlJNk1Dd2ljblFpT2pFMExDSmpaaUk2TUN3aVkzUWlPamdzSW5OMFNXUWlPakY5Iiwid2lkdGgiOjEyODIsImhlaWdodCI6NTcwfV19'">

query_id | initial_query_id | event_time | type | query_duration_ms | read_rows | result_rows | exception | exception_code
-- | -- | -- | -- | -- | -- | -- | -- | --
[...]

PR #1481, opened 2025-01-24
Exception fix

Roma12344321:
## Summary
Added exception handling while executing query
https://github.com/ClickHouse/clickhouse-go/issues/1468

## Checklist
- Unit tests covering the common scenarios were added


SpencerTorres:
Hello! Thank you for submitting this PR.

It looks like this code would work for some cases, however I don't think it would be performant to do multiple regexes for every query exec. Ultimately this needs to be solved at the server level. We have an issue open discussing it here: https://github.com/ClickHouse/ClickHouse/issues/75175

Depending on the server-side fix, some code in this PR may be relevant again, but for now I will close it. I appreciate you looking into this though, if you have any suggestions on how to improve the error detection please leave a comment on the linked issue.

PR #1491, opened 2025-02-06
feat: `clickhouse.JSON` Serializer interface

SpencerTorres:
## Summary
After running benchmarks (#1490), it's clear that the `clickhouse.JSON` type is the fastest way to append JSON data (excluding strings).

Right now the only way to convert a struct to this JSON data is to use reflection magic and walk the struct recursively. A user could also write their own functions to convert their struct to `clickhouse.JSON`. To make this optimization more apparent, I have added interfaces for `clickhouse.JSONSerializer` and `clickhouse.JSONDeserializer`.

If you have a custom struct you can implement these and the JSON column will make use of them when reading/writing data.
A helper function has also been included for easily reading these paths with a specific type in mind (`clickhouse.ExtractJSONPathAs[T](jsonObj, path)`). The user can also do this manually if they choose to.

Other changes:
- Added test for `json.Marshal` to confirm it is using the `NestedMap` func
- Reduced duplicated struct code for JSON tests

## Example

Example:
```go

type ProductPricing struct {
	Price    int64  `json:",omitempty"`
	Currency string `json:",omitempty"`
}

type Product struct {
	ID        clickhouse.Dynamic `json:"id"`
	Name      string             `json:"name"`
	Tags      []string           `json:"tags"`
	Pricing   ProductPricing     `json:"pricing"`
	Metadata  map[string]any     `json:"metadata"`
	CreatedAt time.Time          `json:"created_at" chType:"DateTime64(3)"`
}

// ToClickHouseJSON implements clickhouse.JSONSerializer for faster struct appending
func (p *Product) ToClickHouseJSON() (*clickhouse.JSON, error) {
	obj := clickhouse.NewJSON()
	obj.SetValueAtPath("id", p.ID)
	obj.SetValueAtPath("name", p.Name)
	obj.SetValueAtPath("tags", p.Tags)
	obj.SetValueAtPath("pricing.price", p.Pricing.Price)
	obj.SetValueAtPath("pricing.currency", p.Pricing.Currency)
	obj.SetValueAtPath("metadata.region", p.Metadata["region"])
	obj.SetValueAtPath("metadata.page_count", p.Metadata["page_count"])
	obj.SetValueAtPath("created_at", p.CreatedAt)

	return obj, nil
}

// FromClickHouseJSON implements clickhouse.JSONDeserializer for faster struct scanning
func (p *Product) FromClickHouseJSON(obj *clickhouse.JSON) error {
	p.ID, _ = clickhouse.ExtractJSONPathAs[clickhouse.Dynamic](obj, "id")
	p.Name, _ = clickhouse.ExtractJSONPathAs[string](obj, "name")
	p.Tags, _ = clickhouse.ExtractJSONPathAs[[]string](obj, "tags")
	p.Pricing.Price, _ = clickhouse.ExtractJSONPathAs[int64](obj, "pricing.price")
	p.Pricing.Currency, _ = clickhouse.ExtractJSONPathAs[string](obj, "pricing.currency")
	p.Metadata = make(map[string]any, 2)
	p.Metadata["region"], _ = clickhouse.ExtractJSONPathAs[string](obj, "metadata.region")
	p.Metadata["page_count"], _ = clickhouse.ExtractJSONPathAs[int64](obj, "metadata.page_count")
	p.CreatedAt, _ = clickhouse.ExtractJSONPathAs[time.Time](obj, "created_at")

	return nil
}
```

Then inside a batch append:

```go
batch.Append(&product)
```

The underlying column implementation will then choose to use the user's `clickhouse.JSON` instead of building its own from reflection.

The same applies to `Scan`:

```go
var row TestStruct
err := rows.Scan(&row)
```

The test struct will be populated using the implemented interface.

## Performance

By having the user choose exactly how the object is serialized/deserialized, we can save on CPU and memory allocations:

Serialization:
```
BenchmarkJSONInsert/structs-32       	  275416	      3923 ns/op	    5421 B/op	      55 allocs/op
BenchmarkJSONInsert/fast_structs-32  	  468788	      2497 ns/op	    4968 B/op	      27 allocs/op
```

Deserialization:
```
BenchmarkJSONRead/structs-32                   	 2793265	       410.0 ns/op	     416 B/op	      15 allocs/op
BenchmarkJSONRead/fast_structs-32              	 3078796	       346.9 ns/op	    1040 B/op	       8 allocs/op
```

## Checklist

### DIFF
--- a/examples/clickhouse_api/main_test.go
+++ b/examples/clickhouse_api/main_test.go
@@ -230,6 +230,11 @@ func TestJSONStructExample(t *testing.T) {
 	require.NoError(t, JSONStructExample())
 }
 
+func TestJSONFastStructExample(t *testing.T) {
+	clickhouse_tests.SkipOnCloud(t, "cannot modify JSON settings on cloud")
+	require.NoError(t, JSONFastStructExample())
+}
+
 func TestJSONStringExample(t *testing.T) {
 	clickhouse_tests.SkipOnCloud(t, "cannot modify JSON settings on cloud")
 	t.Skip("client cannot receive JSON strings")
--- a/lib/chcol/json.go
+++ b/lib/chcol/json.go
@@ -25,6 +25,37 @@ import (
 	"strings"
 )
 
+// JSONSerializer interface allows a struct to be manually converted to an optimized JSON structure instead of relying
+// on recursive reflection.
+type JSONSerializer interface {
+	ToClickHouseJSON() (*JSON, error)
+}
+
+// JSONDeserializer interface allows a struct to load its data from an optimized JSON structure instead of relying
+// on recursive reflection to set its fields.
+type JSONDeserializer interface {
+	FromClickHouseJSON(*JSON) error
+}
+
+// ExtractJSONPathAs is a convenience function for asserting a path to a specific type.
+// The underlying value is also extracted from its Dynamic wrapper if present.
+func ExtractJSONPathAs[T any](o *JSON, path string) (T, bool) {
+	value, ok := o.valuesByPath[path]
+	if !ok || value == nil {
+		var empty T
+		return empty, false
+	}
+
+	dynValue, ok := value.(Dynamic)
+	if !ok {
+		valueAs, ok := value.(T)
+		return valueAs, ok
+	}
+
+	valueAs, ok := dynValue.value.(T)
+	return valueAs, ok
+}
+
 // JSON represents a ClickHouse JSON type that can hold multiple possible types
 type JSON struct {
 	valuesByPath map[string]any
--- a/lib/chcol/json.go
+++ b/lib/chcol/json.go
@@ -29,13 +29,13 @@ import (
 // on recursive reflection.
 // Note that the struct must be a pointer in order for the interface to be matched, reflection will be used otherwise.
 type JSONSerializer interface {
-	ToClickHouseJSON() (*JSON, error)
+	SerializeClickHouseJSON() (*JSON, error)
 }
 
 // JSONDeserializer interface allows a struct to load its data from an optimized JSON structure instead of relying
 // on recursive reflection to set its fields.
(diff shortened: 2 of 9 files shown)


lib/chcol/json.go
```diff
@@ -25,6 +25,38 @@ import (
 	"strings"
 )
 
+// JSONSerializer interface allows a struct to be manually converted to an optimized JSON structure instead of relying
+// on recursive reflection.
+// Note that the struct must be a pointer in order for the interface to be matched, reflection will be used otherwise.
+type JSONSerializer interface {
+	ToClickHouseJSON() (*JSON, error)
```
jkaflik, line 32:
```suggestion
	SerializeClickHouseJSON() (*JSON, error)
```

nit: I wonder if it should be named after interface

examples/clickhouse_api/main_test.go
```diff
@@ -230,6 +230,11 @@ func TestJSONStructExample(t *testing.T) {
 	require.NoError(t, JSONStructExample())
 }
 
+func TestJSONFastStructExample(t *testing.T) {
+	clickhouse_tests.SkipOnCloud(t, "cannot modify JSON settings on cloud")
```
jkaflik, line 234: I think we should find a way to make it run against Cloud soon.

jkaflik (approved):
Looks good 🚀

lib/chcol/json.go
SpencerTorres: Good idea. I tried to make these names as distinct as possible so that they wouldn't conflict with any other interfaces such as `SerializeJSON` or `MarshalJSON`. It's a bit long but it is clearer

examples/clickhouse_api/main_test.go
SpencerTorres: Once JSON is stable we should be able to change it in cloud, not sure why it's restricted

PR #1493, opened 2025-02-07
use unsafe.Slice/unsafe.StringData

serprex:
## Summary
https://pkg.go.dev/unsafe#Slice introduced in 1.17


SpencerTorres (approved):
> `safer-unsafe`

Good branch name, good update 🚀

PR #1495, opened 2025-02-11
Release compressor on close

pablomatiasgomez:
## Summary

`compressor` is used along with `buffer` to compress buffer contents.
We have seen that when the `conn_max_lifetime` is somewhat low, the compressor still holds references to big hashtables that are no longer used.
I'm still not sure if there's actually a memory leak somewhere.  

## Checklist

### DIFF
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -127,7 +127,7 @@ func CreateClickHouseTestEnvironment(testSet string) (ClickHouseTestEnvironment,
 	// create a ClickHouse Container
 	ctx := context.Background()
 	// attempt use docker for CI
-	provider, err := testcontainers.ProviderDocker.GetProvider()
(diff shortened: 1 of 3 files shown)


tests/utils.go
```diff
@@ -127,7 +127,7 @@ func CreateClickHouseTestEnvironment(testSet string) (ClickHouseTestEnvironment,
 	// create a ClickHouse Container
 	ctx := context.Background()
 	// attempt use docker for CI
-	provider, err := testcontainers.ProviderDocker.GetProvider()
+	provider, err := testcontainers.ProviderDefault.GetProvider()
```
pablomatiasgomez, line 130: This simply makes testcontainer work with both Docker and Podman (internally it knows how to resolve it)

PR #1497, opened 2025-02-11
change container provider, verify close conn error in tests

pablomatiasgomez:
## Summary

Currently, even if no compression method is configured, we are allocating a new compress.Writer, which by default allocates more than 1MB (See: https://github.com/ClickHouse/ch-go/pull/1039)

Given that we know in advance the method that is going to be used when using the compressor, we can create the compressor for just that method, using the newly introduced constructor `compress.NewWriterWithMethods` in https://github.com/ClickHouse/ch-go/pull/1039


SpencerTorres:
Hey thanks for updating clickhouse-go too!

https://github.com/ClickHouse/ch-go/pull/1039 is merged but not released yet, once it's released we can update this PR. I'll wait to review until then.

Just a note, we do have a couple other references to the compression for the HTTP interface:
https://github.com/ClickHouse/clickhouse-go/blob/a8501776a1b9fa633896005b526970e7636b1598/conn_http.go#L233
https://github.com/ClickHouse/clickhouse-go/blob/a8501776a1b9fa633896005b526970e7636b1598/conn_http.go#L259

Can we verify if it's safe for the compressor to be nil? I haven't checked but it's possible that compressor still gets called with method `None` without nil checks.

Also with your ch-go change we can update that switch statement to only init the selected compressor (as you suggested in this PR). I don't think it should ever be nil though, it may need to be init with only `None` enabled

tests/utils.go
```diff
@@ -127,7 +127,7 @@ func CreateClickHouseTestEnvironment(testSet string) (ClickHouseTestEnvironment,
 	// create a ClickHouse Container
 	ctx := context.Background()
 	// attempt use docker for CI
-	provider, err := testcontainers.ProviderDocker.GetProvider()
+	provider, err := testcontainers.ProviderDefault.GetProvider()
```
pablomatiasgomez, line 130: This simply makes testcontainers work with both Docker and Podman (internally it knows how to resolve it)

pablomatiasgomez:
@SpencerTorres I see the dependency was released so I just updated this PR, let me know your thoughts!

SpencerTorres:
@pablomatiasgomez that was fast! Due to all the other changes I made, I had already started an update from scratch here: https://github.com/ClickHouse/clickhouse-go/pull/1499

I see you have a couple other fixes included though, maybe once that is merged we can merge this to get the other small changes

SpencerTorres:
Merged #1499, let me know if there's anything from this PR you wish to keep

pablomatiasgomez:
@SpencerTorres I just updated the branch, I think the change to use default provider in tests so that Podman also works out of the box is ok, and then freeing the compression reference on close() I think it's a good add, what do you think?

SpencerTorres (approved):
I'm not sure why that block of code assigns those to nil, not sure we need to add it for compressor but it's fine for now. It makes it more consistent with the other pointers on that struct I suppose

Thanks!

PR #1499, opened 2025-02-14
Add LZ4HC compression, update compressor functions, security patch

SpencerTorres:
## Summary
- Update to ch-go v0.65.0
- Included security patch from ch-go
- Adds LZ4HC compression from ch-go
- Reduce memory usage by ~1MB using new ch-go compressor functions
- Updated compression tests to use lz4hc + levels

Related ch-go PRs:
- https://github.com/ClickHouse/ch-go/pull/1039
- https://github.com/ClickHouse/ch-go/pull/1040
- https://github.com/ClickHouse/ch-go/pull/1041

## Checklist

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -73,14 +73,22 @@ func dial(ctx context.Context, addr string, num int, opt *Options) (*connect, er
 		}
 	}
 
-	compression := CompressionNone
+	var (
+		compression CompressionMethod
+		compressor  *compress.Writer
+	)
 	if opt.Compression != nil {
 		switch opt.Compression.Method {
-		case CompressionLZ4, CompressionZSTD, CompressionNone:
+		case CompressionLZ4, CompressionLZ4HC, CompressionZSTD, CompressionNone:
 			compression = opt.Compression.Method
 		default:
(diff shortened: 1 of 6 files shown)


conn.go
```diff
@@ -73,14 +73,22 @@ func dial(ctx context.Context, addr string, num int, opt *Options) (*connect, er
...
-		case CompressionLZ4, CompressionZSTD, CompressionNone:
+		case CompressionLZ4, CompressionLZ4HC, CompressionZSTD, CompressionNone:
 			compression = opt.Compression.Method
 		default:
 			return nil, fmt.Errorf("unsupported compression method for native protocol")
 		}
+
+		compressor = compress.NewWriter(compress.Level(opt.Compression.Level), compress.Method(opt.Compression.Method))
+	} else {
+		compression = CompressionNone
```
pablomatiasgomez, line 90: Is it really needed to initialize this given that the if in line https://github.com/ClickHouse/clickhouse-go/blob/main/conn.go#L247 always check for `!= CompressionNone`

I guess the point was to avoid having a nil compressor just in case?

pablomatiasgomez (approved):

PR #1500, opened 2025-02-15
Add handling for struct in named tuples

### LINKED ISSUE #1010: AppendStruct() and Tuple support
Hi, looks like `AppendStruct` API does not support structs with nested fields. Here's an extended version of [official example](https://github.com/ClickHouse/clickhouse-go/blob/main/benchmark/v2/write-native-struct/main.go):

```go
type Nested struct {
	Foo int64
}

type Row struct {
	Col1       uint64
	Col4       time.Time
	Col2       string
	Col3       []uint8
	Col5       *Nested
	ColIgnored string
}

func TestAppendStructInsert() {
	conn, err := OpenDB()
	if err != nil {
		log.Fatal(err)
	}

	ctx := context.Background()

	if err := conn.Exec(ctx, `DROP TABLE IF EXISTS example`); err != nil {
		log.Fatal(err)
	}

	if err := conn.Exec(ctx, `
		CREATE TABLE example (
			  Col1 UInt64
			, Col2 String
			, Col3 Array(UInt8)
			, Col4 DateTime
			, Col5 Tuple(Foo Int64)
		) Engine = Memory
		`); err != nil {
		log.Fatal(err)
	}

	batch, err := conn.PrepareBatch(context.Background(), "INSERT INTO example")
	if err != nil {
		log.Fatal(err)
	}

	err = batch.AppendStruct(&Row{
		Col1:       uint64(100),
		Col2:       "Golang SQL database driver",
		Col3:       []uint8{1, 2, 3, 4, 5, 6, 7, 8, 9},
		Col4:       time.Now(),
		ColIgnored: "this will be ignored",
		Col5: &Nested{
			Foo: 1000000,
		},
	})
	if err != nil {
		log.Fatal(err)
	}
}

```
The message I get says: 
`clickhouse [AppendRow]: Col5 clickhouse [AppendRow]: converting *main.Nested to Tuple(Foo Int64) is unsupported`

Is there any workaround or any future plans to support this feature?

Exca-DK:
## Summary
Add support for converting Go structs to ClickHouse named tuples.
Closes #1010 

## Checklist

### DIFF
--- a/lib/column/tuple.go
+++ b/lib/column/tuple.go
@@ -511,6 +511,55 @@ func (col *Tuple) AppendRow(v any) error {
 		value = value.Elem()
 	}
 	switch value.Kind() {
+	case reflect.Struct:
+		if !col.isNamed {
+			return &Error{
+				ColumnType: string(col.chType),
+				Err:        fmt.Errorf("converting from %T is not supported for unnamed tuples - use a slice", v),
+			}
+		}
+
+		valueType := value.Type()
+		fieldNames := make(map[string]struct{}, value.NumField())
+		for i := 0; i < value.NumField(); i++ {
+			if !value.Field(i).CanInterface() {
+				// can't interface - likely not exported so ignore the field
+				continue
+			}
+			name, omit := getStructFieldName(valueType.Field(i))
+			if omit {
+				continue
+			}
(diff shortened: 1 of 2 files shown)


SpencerTorres (approved):
Thanks 🚀

PR #1523, opened 2025-03-18
fix concurrent map write errors with context

### LINKED ISSUE #1503: Obscure Concurrent Map Write Panic from Clickhouse-go
## Observed

1. Panic for concurrent map writes hit for the private settings object.
2. Running the query using: `db.Query("SELECT sleep(1), * FROM system.numbers limit 10")` where db is the database/sql DB struct with the Clickhouse driver.

## Expected behaviour

Clickhouse doesn't panic.

## Code example

```go
// nothing really to share.  We're using the database/sql interface to query clickhouse though.
```

## Error log

```log
fatal error: concurrent map writes

goroutine 146 [running]:
internal/runtime/maps.fatal({0x105cbc062?, 0x1063657c0?})
        /opt/homebrew/Cellar/go/1.24.0/libexec/src/runtime/panic.go:1053 +0x20
github.com/ClickHouse/clickhouse-go/v2.queryOptions({0x106578490, 0x1400051b680})
        /Users/abraithwaite/Projects/github.com/ClickHouse/clickhouse-go/context.go:178 +0x158
github.com/ClickHouse/clickhouse-go/v2.(*connect).query(0x1400019ea80, {0x106578490, 0x1400051b680}, 0x106555d30, {0x14000764150, 0x2f}, {0x1072869a0, 0x0, 0x0})
        /Users/abraithwaite/Projects/github.com/ClickHouse/clickhouse-go/conn_query.go:29 +0x74
github.com/ClickHouse/clickhouse-go/v2.(*stdDriver).QueryContext(0x140009aec40, {0x106578490, 0x1400051b680}, {0x14000764150, 0x2f}, {0x1072869a0, 0x0, 0x12e306448?})
        /Users/abraithwaite/Projects/github.com/ClickHouse/clickhouse-go/clickhouse_std.go:290 +0x230
database/sql.ctxDriverQuery({0x106578490?, 0x1400051b680?}, {0x12e3e5f00?, 0x140009aec40?}, {0x0?, 0x0?}, {0x14000764150?, 0x105754048?}, {0x1072869a0?, 0x14000730070?, ...})
        /opt/homebrew/Cellar/go/1.24.0/libexec/src/database/sql/ctxutil.go:48 +0xac
database/sql.(*DB).queryDC.func1()
        /opt/homebrew/Cellar/go/1.24.0/libexec/src/database/sql/sql.go:1786 +0x114
database/sql.withLock({0x1065687f8, 0x14000990700}, 0x140009bb6d8)
        /opt/homebrew/Cellar/go/1.24.0/libexec/src/database/sql/sql.go:3574 +0x74
database/sql.(*DB).queryDC(0x14000598680?, {0x106578490, 0x1400051b680}, {0x0, 0x0}, 0x14000990700, 0x140003c4ab0, {0x14000764150, 0x2f}, {0x0, ...})
        /opt/homebrew/Cellar/go/1.24.0/libexec/src/database/sql/sql.go:1781 +0x144
database/sql.(*Conn).QueryContext(0x14000598680, {0x106578490, 0x1400051b680}, {0x14000764150, 0x2f}, {0x0, 0x0, 0x0})
        /opt/homebrew/Cellar/go/1.24.0/libexec/src/database/sql/sql.go:2037 +0x94
github.com/runreveal/runreveal/api/ch.(*Store).LogsQueryV2(0x140009064c0, {0x106578490, 0x1400051b680}, {0x14000047e80, 0x1b}, {{0x14000764150, 0x2f}, 0x0, 0x0, {0x0, ...}, ...}, ...)
        /Users/abraithwaite/Projects/github.com/runreveal/runreveal/api/ch/query.go:456 +0x854
```

### Environment
* [x] `clickhouse-go` version:
```
grep -i clickhouse go.mod
        github.com/ClickHouse/clickhouse-go/v2 v2.23.2
        github.com/ClickHouse/ch-go v0.61.5 // indirect
```
* [x] Interface: ClickHouse API / `database/sql` compatible driver
* [x] Go version: go version go1.24.0 darwin/arm64
* [x] Operating system: macOS 15.3 24D60 arm64 
* [x] ClickHouse version: 24.8
* [x] Is it a ClickHouse Cloud? No, it's a local container.
* [x] ClickHouse Server non-default settings, if any:  None
* [x] `CREATE TABLE` statements for tables involved:  n/a, system tables & functions only.
* [x] Sample data for all these tables, use [clickhouse-obfuscator](https://github.com/ClickHouse/ClickHouse/blob/master/programs/obfuscator/Obfuscator.cpp#L42-L80) if necessary

SpencerTorres:
## Summary
closes #1503 

This PR fixes 3 bugs:
1. A concurrent map write error where `clickhouse.WithSettings()` maps are re-used between context/calls
2. Per-query settings will no longer override subsequent queries
3. An edge case where an empty `clickhouse.WithSettings(nil)` map could result in a nil write panic

The root cause of #1503 is that maps were being shared between contexts. This is fine when nesting contexts, but there are some cases where the driver will override settings. For example `max_execution_time` is always overridden when the context has a deadline. Another example is `async_insert` being changed when using `AsyncInsert`. When the driver overrides these settings, the settings are persisted for the following queries. This leads to unexpected behavior as well as concurrency issues when overriding settings.

As a fix, the `Settings` map will now always be copied when read in the driver. This is because the deadline context is almost always modified, and settings are almost always overridden (this will increase garbage but it's negligible).

There were two places in the code where the settings are read but not modified: reading async status and reading user timezone location. I added two functions for these cases so we don't need to make an unnecessary copy.

I added many tests to solidify the behavior of the driver's context wrapper. I also added a test case that forces a concurrent map write issue (which doesn't fail anymore since it's fixed).

Function documentation has been updated too to make it clearer in the future. The code is also a bit more readable and easy to follow when seeing where the `Settings` map is initialized.

## Checklist

### DIFF
--- a/context.go
+++ b/context.go
@@ -19,6 +19,7 @@ package clickhouse
 
 import (
 	"context"
+	"maps"
 	"time"
 
 	"github.com/ClickHouse/clickhouse-go/v2/ext"
@@ -43,12 +44,13 @@ type CustomSetting struct {
 type Parameters map[string]string
 type (
 	QueryOption  func(*QueryOptions) error
+	AsyncOptions struct {
+		ok   bool
+		wait bool
+	}
 	QueryOptions struct {
-		span  trace.SpanContext
-		async struct {
-			ok   bool
-			wait bool
-		}
+		span     trace.SpanContext
+		async    AsyncOptions
 		queryID  string
 		quotaKey string
 		events   struct {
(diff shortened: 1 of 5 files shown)


context.go
```diff
@@ -163,26 +165,67 @@ func ignoreExternalTables() QueryOption {
 	}
 }
 
+// Context applies ClickHouse QueryOptions to the current context.
```
abraithwaite, line 168:
```suggestion
// Context returns a new context derived from the given parent, and sets the
// Clickhouse options on the context for the options passed in.  If the parent
// context has options already set on it, we clone those options first.  New
// options given override the old.
```

abraithwaite (approved):
Nice, this is waay cleaner than the slop I wrote!  Just suggested a slightly more detailed doc comment but otherwise looks fantastic, thank you!

PR #1532, opened 2025-03-27
allow to build decimal from string

rutaka-n:
## Summary
Some decimal libraries implements database/sql.Valuer interface by using string representation
allowing it in the driver helps to support wider list of 3d-party libraries for decimal type

## Checklist

### DIFF
--- a/lib/column/decimal.go
+++ b/lib/column/decimal.go
@@ -201,6 +201,12 @@ func (col *Decimal) AppendRow(v any) error {
 		if v != nil {
 			value = *v
 		}
+	case string:
+		d, err := decimal.NewFromString(v)
+		if err != nil {
+			return fmt.Errorf("%v could not be converted to decimal %w", v, err)
+		}
+		value = d
 	case nil:
 	default:
 		if valuer, ok := v.(driver.Valuer); ok {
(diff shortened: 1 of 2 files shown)


rutaka-n:
I am not sure how to test this change without introducing other library, so there is any idea, I would be happy to cover this by test.

SpencerTorres:
> I am not sure how to test this change without introducing other library, so there is any idea, I would be happy to cover this by test.

Haven't looked at the code yet but is there a way to make a test struct that implements `sql.Valuer` ?

steden:
Hi, I am currently using a similar implementation.
Because I have my own decimal type. Instead of shopspring/decimal. 
I had to abandon this package because shopspring/decimal was too memory intensive when calculating.

In my own implementation of the decimal type, Value() returns a string type. Of course I also refer to the design of shopspring/decimal.

Clickhouse-go, as the underlying driver package, should not restrict us on what decimal package to use. Rather, it should allow us to use a more liberal approach, such as string type (numeric) 

When will this PR be merged, please?

rutaka-n:
> > I am not sure how to test this change without introducing other library, so there is any idea, I would be happy to cover this by test.
> 
> Haven't looked at the code yet but is there a way to make a test struct that implements `sql.Valuer` ?
> 
> Edit: I see it's a small change. Yeah maybe a test struct that implements one of these interfaces will work. As long as it evaluates to a string. So maybe something that starts as a `float64` but then does `fmt.Sprintf("%d", v)` just to simulate the stringifying
> 
> Edit 2: It might be good to `fmt.Errorf` wrap that error so we know where to look from the error message

Thank you for looking into code and for your feedback.
I've pushed some changes:
- added error wrapping 
- added test with dummy implementation
Please review changes when you have time.
And let me know if any improvements might be made.

SpencerTorres (approved):

PR #1538, opened 2025-04-12
JWT Authentication

### LINKED ISSUE #1445: Support JWT auth
prerequisites:
adjust Cloud integration tests to run them `GitHub Actions` instance instead of an instance created by the TF provider. JWT-based auth is enabled on the `GitHub Actions` instance only. 

scope:
- ability to attach a JWT token
- ability to overwrite the token on the client instance
- works for both HTTP and TCP connections
- test against ClickHouse Cloud. https://github.com/ClickHouse/clickhouse-js/blob/main/.github/workflows/tests.yml#L218-L220

Reference implementation for HTTP interface:
- nodejs client https://github.com/ClickHouse/clickhouse-js/pull/360
- java client https://github.com/ClickHouse/clickhouse-java/pull/1904

SpencerTorres:
## Summary
closes #1445

This PR adds support for JWT authentication to Native/HTTP interfaces, and the `database/sql` connector.

To use JWT, provide a `GetJWT` implementation in the `clickhouse.Options` struct:
```go
getJWT := func(ctx context.Context) (string, error) {
    return GetEnv("example jwt"), nil
}

opt := &clickhouse.Options{
    GetJWT: jwtFunc,
}
```

If this function is present in the config, the driver will ignore username/password in favor of the JWT.

Whenever a new TCP connection or HTTPS request starts, this function will be called. The JWT can be cached by the application if necessary, or generated each time.

For HTTPS connections, we can update the token immediately per request. For Native/TCP connections we have to wait for the connection to close or be forcibly closed by the server (which happens on token expiry). To force a Native/TCP connection to update, simply make a new DB instance.

If you want to use a specific JWT per query, you must be using HTTPS and not the Native/TCP interface. You can then use the following function to build a context with the given JWT:
```go
ctx := clickhouse.Context(context.Background(), clickhouse.WithJWT("example jwt"))
```

Use this `ctx` in your query and it will use the provided JWT.

Other changes:
- I updated the HTTP request implementation to apply auth headers per-request instead of calculating them once on client initialization. This lets us update the JWT per-request, and makes the code a bit simpler. In the previous version, some functions were manually re-adding these headers which seems very backwards. The updated version cleans this up a bit.

## Checklist


SpencerTorres:
@mshustov let me know if there's any context or functionality missing for this. The native TCP test is failing due to `i/o timeout`, but the `database/sql` tests are passing for both TCP and HTTP which is odd. I need to find a better way to test this locally so I can fix it outside of GitHub Actions 👀

.github/workflows/run-tests-cloud.yml
```diff
@@ -6,6 +6,7 @@ on:
   push:
     branches:
       - main
+      - jwt-auth
```
jkaflik, line 9: Label can be used to trigger this from PR

clickhouse.go
```diff
@@ -375,3 +375,8 @@ func (ch *clickhouse) Close() error {
 		}
 	}
 }
+
+func (ch *clickhouse) UpdateJWT(jwt string) error {
```
jkaflik, line 379: If we let use to provide a callback instead of token value, library can handle all necessary logic related to refresh and error handling.
I'd recommend having a look how other libraries are solving this issue.

jkaflik (commented):
Let's see if we can change approach a little bit.

How is library dealing with authentication errors?

clickhouse.go
SpencerTorres: I agree a callback would be best, I think I avoided that solution originally because I thought it would be too complex for the user. It will definitely be more consistent + easier for the std/sql interface though

SpencerTorres: Also other libraries are using HTTP so it's easier to swap the token immediately, there's still the timeout issue for active TCP connections. With the current implementation it could be possible to trigger a manual token update instead of waiting for the driver to call the callback.

But the connection would be closed either way, so the callback works best

SpencerTorres:
@jkaflik Updated to use a callback, also added `context.Context` override for HTTPS connections. This lets the user control the JWT via the function. They can decide whether to cache it or generate/fetch it each time. They can also swap it out between contexts to support multiple tokens over a single client instance.

Let me know what you think, if there's any concerns, or if something isn't named clearly (such as the primary `GetJWT` function).

Thanks!

conn_http_batch.go
```diff
@@ -207,9 +207,7 @@ func (b *httpBatch) Send() (err error) {
 
 	options.settings["query"] = b.query
 	headers["Content-Type"] = "application/octet-stream"
-	for k, v := range b.conn.headers {
-		headers[k] = v
-	}
+
```
jkaflik, line 210: Why is this removed?

jkaflik (commented):
This approach looks much better. Do you want to provide a README entry?

conn_http_batch.go
SpencerTorres: See comment in thread

SpencerTorres:
> I just noticed this PR does not include happy path test coverage. Do you plan to introduce it?

Yeah we'll expand on this PR later, just want to get the main functionality working first.

Some of the tests are mixed where they work and then intentionally break, I feel like this covers the realistic use of JWT where tokens aren't always valid, but I suppose it's fair to split these into distinct pass/fail logic

PR #1553, opened 2025-05-07
Custom proto revision

titanproger:
## Summary
Add ability to setup in options preferable Native TCP protocol version, to optimise data size, downloaded from server.

https://github.com/ClickHouse/clickhouse-go/discussions/1552

## Checklist
- [ ] A human-readable description of the changes was provided to include in CHANGELOG -  comment: not sure should i change CHANGELOG without version changing - it looks like generated automatically by some script
- [ ] For significant changes, documentation in https://github.com/ClickHouse/clickhouse-docs was updated with further explanations or tutorials - comment: not a significant change


titanproger:
As a justification that the library is supporting different revisions of protocol:
[here](https://github.com/ClickHouse/clickhouse-go/blob/main/conn_handshake.go#L78) revision can be downgraded from server side

SpencerTorres:
Hey! I reached out to your team in Slack to see if we can get some more details about your use case. But until that channel is available I'll make some notes here:

Why would you want to change the protocol version without also adding the features in the code? I fear this will break client compatibility. If it were safe to override the client revision, why can't we simply update the client revision?

I followed the link in the discussion as well, the linked line didn't seem relevant for what you were referring to. I think I need more context. I don't want to expose your implementation details on a public PR so feel free to reach out in the Slack channel.

Thanks! 😄

PR #1564, opened 2025-06-02
Add gotData and endOfProcess handlers

velom:
## Summary
Add two new handlers to track the processing:
1. Helper `WithGotData` allows to pass a function will be triggered on every processing data block.
2. Helper `WithEndOfProcess` allows to pass a function will be triggered when the whole processing going to be finished.

## Checklist

### DIFF
--- a/context.go
+++ b/context.go
@@ -122,6 +122,8 @@ func WithParameters(params Parameters) QueryOption {
 	}
 }
 
+// WithLogs sets a callback function to handle log events during query execution.
+// The provided function `fn` will be called synchronous in the handler goroutine so must not block.
 func WithLogs(fn func(*Log)) QueryOption {
 	return func(o *QueryOptions) error {
 		o.events.logs = fn
@@ -129,6 +131,8 @@ func WithLogs(fn func(*Log)) QueryOption {
 	}
 }
 
+// WithProgress sets a callback function to handle progress events during query execution.
+// The provided function `fn` will be called synchronous in the handler goroutine so must not block.
 func WithProgress(fn func(*Progress)) QueryOption {
 	return func(o *QueryOptions) error {
 		o.events.progress = fn
@@ -136,6 +140,8 @@ func WithProgress(fn func(*Progress)) QueryOption {
 	}
 }
 
+// WithProfileInfo sets a callback function to handle profile information events during query execution.
+// The provided function `fn` will be called synchronous in the handler goroutine so must not block.
 func WithProfileInfo(fn func(*ProfileInfo)) QueryOption {
 	return func(o *QueryOptions) error {
 		o.events.profileInfo = fn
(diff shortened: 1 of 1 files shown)


context.go
```diff
@@ -147,6 +150,20 @@ func WithProfileEvents(fn func([]ProfileEvent)) QueryOption {
 	}
 }
 
+func WithGotData(fn func()) QueryOption {
```
Garbett1, line 163: I know the other options don't have this yet, but do you think we could add a `go doc` comment that explains that these functions are synchronous in the handler goroutine, so must not block?

conn_process.go
```diff
@@ -32,6 +32,8 @@ type onProcess struct {
 	progress      func(*Progress)
 	profileInfo   func(*ProfileInfo)
 	profileEvents func([]ProfileEvent)
+	gotData       func()
```
Garbett1, line 35: What's the difference of `gotData` vs `data`? Other than the function signature?

context.go
velom: Added even for existing helpers

conn_process.go
velom: `data` is "internal" handler set by library itself during query start:
https://github.com/ClickHouse/clickhouse-go/blob/46d229807929e95ba318dda0a6be40328a1a1c05/conn_query.go#L67-L84

conn_process.go
```diff
@@ -178,6 +183,9 @@ func (c *connect) handle(ctx context.Context, packet byte, on *onProcess) error
 		}
 		if block.Rows() != 0 && on.data != nil {
 			on.data(block)
+			if on.gotData != nil {
+				on.gotData()
```
serprex, line 187:
```suggestion
				on.gotData(block)
```
why doesn't gotData take block?

velom: If you'd like to, I can add it

SpencerTorres (changes requested):
*(I sent a response in Slack but I'll copy here for the record)*

I looked at the code here, and I'm hesitant to make these changes since they feel very application-specific (to our internal app). I want to learn more about what the goal of these hooks are.

Specifically these things stand out:
1. Having hooks for `data` and `gotData` is confusing, the names need to be clearer. We could rename `data` to `block` and `gotData` to `firstBlock` (assuming the goal is to track the first block, or count blocks). Maybe we could rename `data` to some internal function like `_data` since it's the only one that is always overridden by the driver.
2. I saw the code suggestion to add `*Block` to `gotData`. I think this would cause some data races if the user had access to these outside of the driver, and so we shouldn't expose them.
3. `endOfProcess` can be renamed `endOfStream` since the server packet is called `EndOfStream`. This depends on whether your goal is to track the end of processing, or the actual `EndOfStream` packet.
4. Do we need `endOfProcess` to know if a query ended? This should already be detectable through the responding `driver.Rows` return value.

Overall I want to make sure we aren't adding a solution to something really specific for one application. If we add it we need to test/maintain it, and I am suspicious about these lower level hooks. I feel like there's another way to expose these events, perhaps in the `driver.Rows` interface.

velom:
Implemented in the `driver.Rows` interface
https://github.com/ClickHouse/clickhouse-go/pull/1796

PR #1573, opened 2025-06-07
fix: avoid placeholder detection in  SQL comments

### LINKED ISSUE #1537: Question mark in comment section recognized as query param
Hello, I found issue #1489 and it may be similar, but i think my problem need different fix.
driver: clickhouse-go/v2 v2.30.0

### Example
```golang
func main() {
	conn, err := clickhouse.Open(&clickhouse.Options{Addr: []string{"localhost:9001"}})
	if err != nil {
		panic(err)
	}

	row := conn.QueryRow(context.TODO(), "select $1 --some comment with ?", "clickhouse")
	fmt.Println(row.Err())
}

```
Output: 
```
clickhouse [bind]: mixed named, numeric or positional parameters
```
I think method [bind()](https://github.com/ClickHouse/clickhouse-go/blob/main/bind.go#L67) has incorrect regexps. Needed to edit regexp or pass query in it without comments

tosolveit:
## Summary

This pull request fixes issue #1537, where parameter placeholders such as `?` were incorrectly detected inside `--` single-line SQL comments. This led to false-positive errors like:

    clickhouse [bind]: mixed named, numeric or positional parameters

when using `$1` placeholders together with comments.

### What this patch does

- Adds a `stripLineComments` helper in `bind.go` that removes `--` comments before applying placeholder-detection regexes.
- Ensures that `?` or `$1` inside comments are not interpreted as actual parameter placeholders.
- Adds a regression test under `tests/issues/1537_test.go` to verify the fix and prevent future regressions.

### Limitations

This patch only addresses `--` style comments. It does not currently handle:
- `/* ... */` block comments
- `'string literals'` containing `?` or `$1`

These could be addressed in future improvements if needed.

---

Fixes #1537

## Checklist

### DIFF
--- a/bind.go
+++ b/bind.go
@@ -82,8 +82,9 @@ func bind(tz *time.Location, query string, args ...any) (string, error) {
 		return bindNamed(tz, query, args...)
 	}
 
-	haveNumeric = bindNumericRe.MatchString(query)
-	havePositional = bindPositionalRe.MatchString(query)
+	cleanQuery := stripLineComments(query)
+	haveNumeric = bindNumericRe.MatchString(cleanQuery)
+	havePositional = bindPositionalRe.MatchString(cleanQuery)
 	if haveNumeric && havePositional {
 		return "", ErrBindMixedParamsFormats
 	}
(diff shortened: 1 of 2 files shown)


SpencerTorres:
I wonder if any queries would be broken by merging this. It might be safer to tell users to remove comments like this one.

I need to follow up in the original issue, but is there any valid case to have comments like this in the query?

tosolveit:
Hi,
I see your point, there are many cases that needs to be handled and yes I agree an elegant solution might better, I approached pragramatically.

I also understand that you don't want to touch to the query not to introduce some risks, you also mentioned about adding comments, one possibility is that the user might have many SQL files and maybe comment are already added before, It would be weird to say delete your comments when submitting queries in my opinion.

I am very new to the project and I'm **totally** fine if we skip to merge this and get back to this later.

zamai:
It's a pity that this fix wasn't merged because I faced this issue today, spend 2 hours debugging just to realize that client was expecting me to pass parameter to the value I commented out...

PR #1575, opened 2025-06-10
fix: namedValue namedDatevalue usage error

### LINKED ISSUE #1555: Cannot use NamedValue and  NamedDateValue together
example query

```
select count() from db.{table:Identifier} WHERE datetime BETWEEN @start AND @end
err = client.Select(
		context.TODO(),
		&items,
		query,
		clickhouse.Named("table", table),
		clickhouse.DateNamed("start", time.Unix(input.Since, 0), clickhouse.Seconds),
		clickhouse.DateNamed("end", time.Unix(input.Till, 0), clickhouse.Seconds)
	)
```

breaks in query_parameters.go

```
if paramsProtocolSupport &&
		len(args) > 0 &&
		hasQueryParamsRe.MatchString(query) {
		options.parameters = make(Parameters, len(args))
		for _, a := range args {
			if p, ok := a.(driver.NamedValue); ok {
				if str, ok := p.Value.(string); ok {
					options.parameters[p.Name] = str
					continue
				}
			}

			return "", ErrExpectedStringValueInNamedValueForQueryParameter
		}

		return query, nil
	}
```

gives error `expected string value in NamedValue for query parameter`

tosolveit:
## Summary
Fixes  #1555 query parameter binding for queries using named date values such as clickhouse.DateNamed(...).

Previously, if the query contained Named parameters (e.g., {table:Identifier}) and also used DateNamed, the driver returned ErrExpectedStringValueInNamedValueForQueryParameter.

This update extends bindQueryOrAppendParameters to support driver.NamedDateValue types and formats them using formatTimeWithScale, ensuring proper formatting and compatibility with the query parameter protocol.

## Checklist

### DIFF
--- a/query_parameters.go
+++ b/query_parameters.go
@@ -27,6 +27,8 @@ import (
 
 var (
 	ErrExpectedStringValueInNamedValueForQueryParameter = errors.New("expected string value in NamedValue for query parameter")
+	ErrInvalidValueInNamedDateValue                     = errors.New("invalid value in NamedDateValue for query parameter")
+	ErrUnsupportedQueryParameter                        = errors.New("unsupported query parameter type")
 
 	hasQueryParamsRe = regexp.MustCompile("{.+:.+}")
 )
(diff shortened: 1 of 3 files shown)


SpencerTorres (approved):
Thanks for the fix! We will likely deprecate the type of parameters used in the original issue in favor of ClickHouse's syntax, but this is still helpful for existing users. https://clickhouse.com/docs/sql-reference/syntax#defining-and-using-query-parameters

Issue #1579, opened 2025-06-18
Add support for Time and Time64 types


mshustov:
### Use case
https://github.com/ClickHouse/ClickHouse/pull/81217 brings `Time` and `Time64` support into ClickHouse. See the PR for docs and coding/encoding logic.

### Describe the solution you'd like
The client and driver support the new types: allow using in DDL + write / read operations.

tosolveit:
Hi,
Thanks for your reply, i was working on this and got errors when testing and wanted to clarify here.

I will review my work and create the PR then, thanks

tosolveit:
Hi @Slach ,

As of now If I pull the clickhouse, it returns version `25.5.2.47`
```bash
docker pull clickhouse/clickhouse-server:latest
```

I tried with a more recent release.
```sql
SELECT version(); -- 25.5.3.75
```

 And it returnes zero rows for 25.5.3.75
```sql
SELECT name, value, description
FROM system.settings
WHERE name LIKE 'allow_experimental_time_time64_type%';
```

or when I run
```sql
SET allow_experimental_time_time64_type=1
```
I get

```text
SQL Error [22000]: Code: 115. DB::Exception: Unknown setting 'allow_experimental_time_time64_type': Maybe you meant ['allow_experimental_object_type','allow_experimental_time_series_table']. (UNKNOWN_SETTING) (version 25.5.3.75 (official build)) 
```

So as you see we can't set this experimental feature, please let me know if I'm missing something.

mshustov:
> As of now If I pull the clickhouse, it returns version 25.5.2.47

@tosolveit  It will be part of `25.6`, see https://github.com/ClickHouse/ClickHouse/pull/81217

tosolveit:
Hi,
I looked into implementing the types requested here and realized that they first need to be added to ch-go. I’ve already worked on an initial implementation on this side and feel confident about the approach.

I've also offered help on the corresponding [issue](https://github.com/ClickHouse/ch-go/issues/1070) in ch-go and am currently waiting for a response from the maintainer. 

Just wanted to share this context in case it helps clarify the current status.

Thanks!

shivanshuraj1333:
PRs, which adds the support in our go client https://github.com/ClickHouse/clickhouse-go/pull/1596, corresponding ch-go changes https://github.com/ClickHouse/ch-go/pull/1074

PR #1596, opened 2025-07-11
feat: add support for time/time64

shivanshuraj1333:
fixes: https://github.com/ClickHouse/clickhouse-go/issues/1579

related to https://github.com/ClickHouse/ClickHouse/pull/81217

depends on https://github.com/ClickHouse/ch-go/pull/1074

## Checklist


go.mod
```diff
@@ -22,6 +22,10 @@ require (
 
 require go.opentelemetry.io/otel v1.37.0 // indirect
 
+replace github.com/ClickHouse/ch-go => ../ch-go
```
shivanshuraj1333: will update once https://github.com/ClickHouse/ch-go/pull/1074 is merged

shivanshuraj1333:
Hey @SpencerTorres when you get some time, can you please take a look at this one?

(also tagging you as some other PRs which can get some help from your reviews)

Thanks!

go.mod
```diff
@@ -5,7 +5,7 @@ go 1.23.0
 toolchain go1.24.1
 
 require (
-	github.com/ClickHouse/ch-go v0.66.1
+	github.com/ClickHouse/ch-go v0.66.2-0.20250716011122-97e525b8cb83
```
shivanshuraj1333: updated to bring changes from https://github.com/ClickHouse/ch-go/pull/1074

this is using commit id, a new tag is not created.

shivanshuraj1333:
https://github.com/ClickHouse/ch-go/pull/1074 is now, merged, this is now ready for review/merge.

lib/column/time.go
```diff
@@ -0,0 +1,348 @@
...
+
+func (col *Time) Reset() {
+	col.col.Reset()
+}
+
+func (col *Time) Name() string {
+	return col.name
+}
+
+func (col *Time) parse(t Type, tz *time.Location) (_ Interface, err error) {
```
SpencerTorres, line 50: Update function signature to the latest. I recently changed this on `main`, the `parse` function is now expected to accept a `sc *ServerContext`. You could then use `sc.Timezone`.

You could keep the param as timezone though, see my other comment about `column.tpl`.

lib/column/time.go
```diff
@@ -0,0 +1,348 @@
...
+		if col.timezone != nil {
+			timezone = col.timezone
+		}
+		return time.Date(1970, 1, 1, int(hours), int(minutes), int(secs), 0, timezone), nil
+	}
+
+	return time.Time{}, fmt.Errorf("cannot parse time value: %s", value)
+}
+
+func (col *Time) WriteStatePrefix(buffer *proto.Buffer) error {
```
SpencerTorres, line 342: I believe this is an optional interface, along with `ReadStatePrefix`. You should be able to remove these.

lib/column/time.go
SpencerTorres: Same for Time64

lib/column/time.go
SpencerTorres: Same for Time64

lib/proto/block.go
```diff
@@ -229,9 +229,12 @@ func (b *Block) Decode(reader *proto.Reader, revision uint64) (err error) {
 				return err
 			}
 			if hasCustom {
-				return &BlockError{
-					Op:  "Decode",
-					Err: errors.New(fmt.Sprintf("custom serialization for column %s. not supported by clickhouse-go driver", columnName)),
+				// Allow Time and Time64 columns with custom serialization
+				if columnType != "Time" && columnType != "Time64" {
```
SpencerTorres, line 233: What is this? How is Time and Time64 encoded where this is something that needs to be changed? I never saw what `hasCustom` is in the protocol, is this really the first column that has the bool set to `true`?

If we need to support some columns with custom encoding let's make an interface similar to `CustomSerialization` with `WriteStatePrefix` and `ReadStatePrefix`. Perhaps an interface that has a single `AllowCustom() bool` on it. If it returns false then we throw this error.

Also note that `columnType` could be complex such as `Time64(precision)`, so `Time` and `Time64` would never match if they had any parameters.

tests/time_test.go
```diff
@@ -0,0 +1,406 @@
...
+	"testing"
+	"time"
+
+	"github.com/ClickHouse/clickhouse-go/v2"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestTimeBasic(t *testing.T) {
+	// Skip if no ClickHouse server is available
```
SpencerTorres: ClickHouse server is always available in this testing folder, these are integration tests. There are other places we could put this logic, let's not duplicate these skips everywhere.

If you're testing logic related to the column type itself, you can put it in the `column` package next to the source file. `time_test.go` and such.

tests/time_test.go
```diff
@@ -0,0 +1,406 @@
...
+	if testing.Short() {
+		t.Skip("Skipping integration test in short mode")
+	}
+
+	ctx := context.Background()
+	conn, err := clickhouse.Open(&clickhouse.Options{
```
SpencerTorres: You should reference `variant_test` or `dynamic_test` or `json_test`, these are also experimental/new similar to the time type. 

This connection will not run on CI due to the server being unavailable. Our local machines and CI machines usually have docker so we can run multiple versions of ClickHouse through the same tests.

Some considerations/notes:
- Use a setup function. This will be useful for skipping versions that don't have this column type.
- The setup function can also be used to include the `enable_time_timet64_type` setting
- Use the `GetNativeConnection` function as seen in the other tests
- We must test TCP and HTTP connections, use `TestProtocols` wapper in each test as seen in the other tests

tests/time_test.go
```diff
@@ -0,0 +1,406 @@
...
+	if err != nil {
+		t.Skipf("Skipping test - no ClickHouse server available: %v", err)
+	}
+	defer conn.Close()
+
+	// Test that we can create a table with Time and Time64 columns
+	require.NoError(t, conn.Exec(ctx, "DROP TABLE IF EXISTS test_time_registration"))
+
+	const ddl = `
+		CREATE TABLE test_time_registration (
```
SpencerTorres: This is good that the table names are different. This prevents session/table locking issues for HTTP tests. It's best that each test has its own table name

lib/column/time.go
```diff
@@ -0,0 +1,348 @@
...
+	"strings"
+	"time"
+
+	"github.com/ClickHouse/ch-go/proto"
+	"github.com/ClickHouse/clickhouse-go/v2/lib/timezone"
+)
+
+const defaultTimeFormat = "15:04:05"
+
+type Time struct {
```
SpencerTorres, line 35: You will need to add `Time` and `Time64` to `lib/column/codegen/column.tpl` and run the codegen to re-generate `column_gen.go`. This will allow our column parser to automatically initialize the `Time` and `Time64` types.

This is critical for actually adding the type, otherwise the column won't be recognized in the driver. It also helps other wrapper types like `Array(Time)` init the type.

SpencerTorres (changes requested):
Hey! Thanks for submitting this.

The implementation itself looks good, but there's definitely some changes needed.

Apologies for the testing issues as well, I'm not sure if you have docker available but we have a really easy way of running integration tests to where you don't need to run your own instance. Let me know if you have any questions about this or any other notes I've left.

Overall I'd recommend you take a look at the other column types and integration tests.

Also it would be good to get some tests for the standard library `database/sql` interface. You can find those in `tests/std`. These are important for ensuring our driver works with other SQL wrappers.

PR #1597, opened 2025-07-13
feat: add Session api for stateful connection and set role support

### LINKED ISSUE #1443: Support `SET ROLE` for `clickhouse.Conn`
### Discussed in https://github.com/ClickHouse/clickhouse-go/discussions/1391

<div type='discussions-op-text'>

<sup>Originally posted by **jpugliesi** August 29, 2024</sup>
We have a use case where we'd like to be able to dynamically set the connection `ROLE` prior to executing a query (and unset it after). Ex:
```
SET ROLE some_role;
SELECT ...
```

This does not seem currently possible with the native/`clickhouse` `clickhouse.Conn`, because it pools connections under the hood and does not expose an API to retrieve and use the same, stateful connection for multiple statements. 

It seems like the only way to achieve this is to use the `database/sql` driver's `BeginTx` to execute multiple statements on the same connection.

Is there some way to support `SET ROLE` with the native/`clickhouse` driver? If not, are there plans to support this functionality somehow?

This relates to https://github.com/ClickHouse/clickhouse-go/issues/772, where exposing connection pool's `acquire` (or hooks into connection acquire/release from the pool) would allow this</div>

shivanshuraj1333:
fixes: https://github.com/ClickHouse/clickhouse-go/issues/1443

please check the end to end implementation with this toy program https://github.com/shivanshuraj1333/clickhuose-test

make sure to update the go.mod to point to the changes present in this PR

## Summary:
- Problem: ClickHouse Go driver lacks support for SET ROLE functionality due to connection pooling, which doesn't maintain connection state across multiple statements.

- Solution: Added session management API that provides stateful connections while maintaining backward compatibility.

Key Changes
- New Session Interface: Added Session interface with SetRole(), UnsetRole(), and Close() methods
- Connection Enhancement: Added AcquireSession() method to main connection interface

Implementation:
- session struct for native driver
- stdSession struct for standard SQL driver
- Proper resource management and error handling

example usage:
```go
// Acquire a stateful session
session, err := conn.AcquireSession()
if err != nil {
    return err
}
defer session.Close()

// Set role for this session
if err := session.SetRole("admin"); err != nil {
    return err
}

// Execute queries - role persists across operations
rows, err := session.Query("SELECT currentUser()")
// ... more operations

// Unset role when done
session.UnsetRole()
```


shivanshuraj1333:
@SpencerTorres need your help with reviews on this one.
Thanks!

shivanshuraj1333:
These are some great feedbacks, IMO if we want to role this out, we can do this Iteratively like below.

1) => AFAIU
Now, when using HTTP, driver does not automatically manage session_id for sessions, so session state only persists if the user manually sets and reuses the same session_id for all requests, else each HTTP request is stateless even when using the session API. Maybe we can document that and as an enhancement add a followup which updates the driver so that it will automatically generate and inject a unique session_id for each HTTP session. Does that sound right approach to you?

2) => Sure

3) => Shall we add these in docs?
Always call .Close() on your session when done, to avoid leaking connections.
If a session encounters a fatal error (e.g., connection closed), it becomes unusable and should be closed. Do not attempt to reuse a session after a fatal error.

4) => Have made a slight change which closes the connection instead of returning it to the pool

lmk your thoughts, and thanks for the feedbacks!

must108 (changes requested):
Hi @shivanshuraj1333, thanks for the contribution. Could you please address the github-bot and Claude comments? Also, please address the conflicts, thanks.

PR #1602, opened 2025-07-14
feat: add ssh key authentication support

shivanshuraj1333:
Adds fix for https://github.com/ClickHouse/clickhouse-go/discussions/1601
related to:
https://github.com/ClickHouse/ch-go/issues/350
https://github.com/ClickHouse/ch-go/pull/1075
https://github.com/ClickHouse/ClickHouse/pull/41109

## Summary
Adds ssh key auth support

## Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.23.0
 toolchain go1.24.1
 
 require (
-	github.com/ClickHouse/ch-go v0.66.1
+	github.com/ClickHouse/ch-go v0.67.0
 	github.com/andybalholm/brotli v1.2.0
 	github.com/docker/docker v28.3.2+incompatible
 	github.com/docker/go-units v0.5.0
@@ -16,6 +16,7 @@ require (
 	github.com/stretchr/testify v1.10.0
 	github.com/testcontainers/testcontainers-go v0.38.0
 	go.opentelemetry.io/otel/trace v1.37.0
+	golang.org/x/crypto v0.40.0
 	golang.org/x/net v0.42.0
 	gopkg.in/yaml.v3 v3.0.1
 )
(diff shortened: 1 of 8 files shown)


go.mod
```diff
@@ -74,7 +77,14 @@ require (
...
+	go.uber.org/zap v1.27.0 // indirect
+	golang.org/x/crypto v0.40.0 // indirect
+	golang.org/x/mod v0.23.0 // indirect
+	golang.org/x/sync v0.16.0 // indirect
+	golang.org/x/sys v0.34.0 // indirect
+	golang.org/x/tools v0.30.0 // indirect
 	google.golang.org/genproto/googleapis/api v0.0.0-20240318140521-94a12d6c2237 // indirect
 )
+
+replace github.com/ClickHouse/ch-go => ../ch-go
```
shivanshuraj1333: would update it once https://github.com/ClickHouse/ch-go/pull/1075 is merged and a tag is released

shivanshuraj1333: updated to `v0.67.0`

shivanshuraj1333:
@ernado ptal when you've some bandwidth

SpencerTorres (changes requested):
Similar comments to #1596, I think we should try to improve the testing here.

The unit tests can stay where they are, but it would be nice to have an integration test for this. We have a `resources` folder where we could mount and test SSH auth with the ClickHouse server container.

I like the client options, being able to provide a file path via URL seems convenient for external tools dependent on the driver + a URL. I think the custom signer is more useful though, I don't want to have too many file system operations within the driver so it's good to give the user that option.

I'm actually not familiar with the authentication process with SSH and ClickHouse. Can you provide some more info on the structure of the exchange and the limitations?

must108 (commented):
ping @shivanshuraj1333

Issue #1607, opened 2025-07-17
`ReadTimeout` appears to apply to the first read block only


GeorgeMac:
## Observed

We have recently observed calls `rows.Next()` blocking indefinitely.

We noticed that a query, which appears to have at-least returned its first block, get stuck on the CH server side, then never results in a read timeout. Instead, our code was permanently blocked on a call to `rows.Next()`.

```
goroutine 1098 [select, 72 minutes]:
github.com/ClickHouse/clickhouse-go/v2.(*rows).Next(0xc019ae3c80)
	/go/pkg/mod/github.com/!click!house/clickhouse-go/v2@v2.37.2/clickhouse_rows.go:52 +0x239
```

> Taken from a goroutine profile. Here is where we observed the rows.Next call being blocked for 72 minutes.

Blocking indefinitely on this `select` statement:
https://github.com/ClickHouse/clickhouse-go/blob/2482badc9eeb6815f8bfc2afe4a8b68399f3e8d4/clickhouse_rows.go#L52-L67

Digging into the client code it appears that `ReadTimeout` is currently only being applied to the first block:
https://github.com/ClickHouse/clickhouse-go/blob/2482badc9eeb6815f8bfc2afe4a8b68399f3e8d4/conn_query.go#L41-L48

This is because we're deferring resetting the read (and read/write) deadline to the zero value.

Since the rest of the connection is processed asynchronously in the goroutine below, this read timeout can be cleared regardless whether or not a subsequent is scheduled before returning the rows. Even so, as the docs for `SetReadDeadline` state, setting this deadline appears for future and current block read calls anyway:

```go
	// SetReadDeadline sets the deadline for future Read calls
	// and any currently-blocked Read call.
	// A zero value for t means Read will not time out.
	SetReadDeadline(t time.Time) error
```

## Expected behaviour

Our understanding is that the read deadline will apply to all reads performed during the lifetime of a query.
Such that if CH server fails to deliver subsequent bytes for whatever reason, the client will fail with a timeout and can move on.

Here are a couple thoughts on how `ReadTimeout` could be interpretted.
I am unsure what the expectation is, I can't see any documentation for it:

### read deadline per call to `conn.Read`

Maybe what we want is to apply the read timeout per call to Read.
Setting the deadline explicit before starting each call.
So that a long query continues to run if we're still getting bytes back from CH.

### read deadline on rows.Close

Or perhaps clearing the read timeout should move to occur within `rows.Close()`.
[...]

PR #1608, opened 2025-07-17
Reset connection deadline before connection is released

jkaflik:
## Summary

It's not well well-thought-out fix attempt for https://github.com/ClickHouse/clickhouse-go/issues/1607.

Requires deep dive into edge cases.

## Checklist

### DIFF
--- a/conn_query.go
+++ b/conn_query.go
@@ -44,7 +44,6 @@ func (c *connect) query(ctx context.Context, release nativeTransportRelease, que
 	// context level deadlines override any read deadline
 	if deadline, ok := ctx.Deadline(); ok {
 		c.conn.SetDeadline(deadline)
-		defer c.conn.SetDeadline(time.Time{})
 	}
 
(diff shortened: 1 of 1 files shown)


conn_query.go
```diff
@@ -80,6 +79,8 @@ func (c *connect) query(ctx context.Context, release nativeTransportRelease, que
 		}
 		close(stream)
 		close(errors)
+
+		c.conn.SetDeadline(time.Time{})
```
GeorgeMac, line 83: I think we might need to do this for `SetReadDeadline` too. As the SetDeadline is an override on the context. I am unsure what we're using exactly in our case.

p.s. thanks for jumping on this so quickly!

SpencerTorres:
Thanks for the info + discussion in this thread!

@GeorgeMac's solution is what we need. I have merged a snippet from his solution in https://github.com/ClickHouse/clickhouse-go/pull/1612 in https://github.com/ClickHouse/clickhouse-go/pull/1617

PR #1643, opened 2025-09-03
Division by zero panic fix

artemseleznev:
The Svace static analyzer (https://www.ispras.ru/en/technologies/svace/) flagged this code as suspicious: it could lead to a panic (division by zero) when data is empty.

### DIFF
--- a/lib/column/fixed_string.go
+++ b/lib/column/fixed_string.go
@@ -178,7 +178,11 @@ func (col *FixedString) Append(v any) (nulls []uint8, err error) {
 			return nil, err
 		}
 
-		nulls = make([]uint8, len(data)/col.col.Size)
+		if col.col.Size == 0 {
+			nulls = make([]uint8, 0)
+		} else {
+			nulls = make([]uint8, len(data)/col.col.Size)
+		}
 	case [][]byte:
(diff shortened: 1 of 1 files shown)


lib/column/fixed_string.go
```diff
@@ -178,7 +178,11 @@ func (col *FixedString) Append(v any) (nulls []uint8, err error) {
 			return nil, err
 		}
 
-		nulls = make([]uint8, len(data)/col.col.Size)
+		if col.col.Size == 0 {
```
kavirajk, line 181: can we do something like following?
```
size := 0
if col.col.Size != 0 {
    size = len(data)/col.col.Size
}
nulls = make([]uint8, size)
```

Rationale being, this makes it more readable IMHO. Some additional advantages like
1. Less duplication with `make()`
2. Bit simpler control flow
3. Single place to calculate the value of `size`.

what do you think?

kavirajk:
closing in favor of #1665. Thanks for the contribution @artemseleznev :)

lib/column/fixed_string.go
artemseleznev: yes, you're right! thanks!

PR #1655, opened 2025-09-14
Fix `ConnOpenRandom` case in `ConnOpenStrategy` during connection dial

### LINKED ISSUE #1653: Upgrade to `math/rand/v2`
We have some tests that depend on random number generation, mainly for connection host failover. These tests were never really safely picking random numbers, they were inconsistent. I hacked some `rand.Seed` calls in to get it working, but never liked the solution.

With Go 1.25, `rand.Seed` doesn't work anymore and broke our main branch. I've skipped these tests for now (#1652) but we should look into updating to the new [math/rand/v2](https://go.dev/pkg/math/rand/v2) package.

Related blog post:

https://go.dev/blog/randv2

kavirajk:
Fixes: #1653 

## Summary

This makes sure the random open strategy always tries all given addresses to find working server eventually.

Previously there is a chance, that same non-working server could have been picked twice thus ending up not picking working server at all (by not exploring all given addresses)

Basically, [implemented the proposal (1) as mentioned in this comment.](https://github.com/ClickHouse/clickhouse-go/issues/1653#issuecomment-3289976523)

#### Main changes
1. Create `rand.Int()` outside the loop once. And use it with `rand + i` to cover all given addresses correctly
2. Remove the need for any `seed()` in the tests.

Now the test passes without any extra maintenance like seeding.
```
go test ./tests/... -run TestConnFailoverRandom -count=1 -v
```
This always passes now without random failure.

## Checklist


must108 (changes requested):
LGTM, please just address agent comments, and fix test cases

PR #1656, opened 2025-09-16
feat: add custom http transport

r0bobo:
## Summary

This makes it possible to set a custom transport for the underlying http client.

## Checklist

### DIFF
--- a/clickhouse_options.go
+++ b/clickhouse_options.go
@@ -168,6 +168,10 @@ type Options struct {
 	// to respond to a single Read call for bytes over the connection.
 	// Can be overridden with context.WithDeadline.
 	ReadTimeout time.Duration
+
+	// Set a custom transport for the http client.
+	// The default transport configured by the library is passed in as an argument.
+	TransportFunc func(*http.Transport) (http.RoundTripper, error)
 }
 
 func (o *Options) fromDSN(in string) error {
--- a/conn_http.go
+++ b/conn_http.go
@@ -224,7 +224,7 @@ func dialHttp(ctx context.Context, addr string, num int, opt *Options) (*httpCon
 		httpProxy = http.ProxyURL(opt.HTTPProxyURL)
 	}
 
-	t := &http.Transport{
+	tr := &http.Transport{
 		Proxy: httpProxy,
 		DialContext: (&net.Dialer{
 			Timeout: opt.DialTimeout,
@@ -237,9 +237,20 @@ func dialHttp(ctx context.Context, addr string, num int, opt *Options) (*httpCon
 	}
 
 	if opt.DialContext != nil {
-		t.DialContext = func(ctx context.Context, network, addr string) (net.Conn, error) {
+		tr.DialContext = func(ctx context.Context, network, addr string) (net.Conn, error) {
 			return opt.DialContext(ctx, addr)
 		}
+	}
+
+	var rt http.RoundTripper
+
+	if opt.TransportFunc != nil {
+		rt, err = opt.TransportFunc(tr)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		rt = tr
 	}
 
 	conn := httpConnect{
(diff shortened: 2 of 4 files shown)


conn_http.go
```diff
@@ -237,19 +237,30 @@ func dialHttp(ctx context.Context, addr string, num int, opt *Options) (*httpCon
...
 	if opt.DialContext != nil {
-		t.DialContext = func(ctx context.Context, network, addr string) (net.Conn, error) {
+		tr.DialContext = func(ctx context.Context, network, addr string) (net.Conn, error) {
 			return opt.DialContext(ctx, addr)
 		}
 	}
 
+	var rt http.RoundTripper
+
+	if opt.TransportFunc != nil {
```
kavirajk: Curious if we should assign the transport right after the creation of default transport (`&http.Transport{}`? Because otherwise any additional mutation on transport (like `tr.DialContext`) won't be added correctly for custom transport?

Sure, the `TransportFunc` takes current `http.Transport` but no guarantee, the consumer copies all the options correctly thus bit more error prone I think? If we decide the transport right after default transport creation, then we can be sure additional options on transport are correctly applied?

May be I miss your exact use-case of overriding the transport at the very end? what do you think?

Also anyway we can lock this behavior via some tests as well?

clickhouse_options.go
```diff
@@ -168,6 +168,10 @@ type Options struct {
 	// to respond to a single Read call for bytes over the connection.
 	// Can be overridden with context.WithDeadline.
 	ReadTimeout time.Duration
+
+	// Set a custom transport for the http client.
+	// The default transport configured by the library is passed in as an argument.
+	TransportFunc func(*http.Transport) (http.RoundTripper, error)
```
kavirajk, line 174: curious the rationale behind accepting `*http.Transport` type and returning `http.RoundTripper`?. 

Given the use case and name `TransportFunc`, I would prefer to keep it simple like `func(*http.Transport) *http.Transport`. Even we don't need additional `error` return I think?.

conn_http.go
r0bobo: Our usecase is to inject a "middleware" roundtripper that wraps the upstream transport created by `clickhouse-go` so that we can modify the request.
Something like:
```go
type MyRoundTripper struct {
	Transport *http.Transport
}

func (t *MyRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
	// Mutate the req before passing it on to the upstream transport.
	return t.Transport.RoundTrip(req)
}

var options = &clickhouse.Options{
	TransportFunc: func(t *http.Transport) (http.RoundTripper, error) {
		return &MyRoundTripper{
			Transport: t,
		}, nil
	},
}
```

Since we want to expose our own custom struct that implements the [http.RoundTripper interface](https://pkg.go.dev/net/http#RoundTripper) rather than a new concrete [http.Transport struct](https://pkg.go.dev/net/http#Transport) we can't treat it as `http.Transport` and run the `tr.DialContext()` functions against it.

I however understand your point about the behavior potentially being a bit surprising if you completely replace the transport. :thinking:

clickhouse_options.go
r0bobo: We need to return a `http.RoundTripper` because of the reason described in the previous comment. We could also just pass in the transport as a `http.RoundTripper` so the signature becomes:
```go
TransportFunc func(http.RoundTripper) http.RoundTripper
```
But then you can't access the fields of the transport, making the function less powerful. It would however cover our usecase regardless so I don't have a strong opinion regarding having a `*http.Transport` vs `http.RoundTripper` as the input argument.

Regarding the return error I just thought it's practical to have the option to return an error. But I guess you can do all fallible operations outside of that function.

kavirajk: Ok. I see the rationale. I think that's fine. Sounds good to me to leave the signature as is. Can we just add a small test case? happy to merge after that :)

conn_http.go
kavirajk: yea I'm just worried what if `DialContext()` is applied on old transport but get lost when overwriting the transport layer.

I think we can also leave it to user to persist the passed in transport in their custom RoundTripper (like the way you did above).

I'm happy with keep it like this.

clickhouse_options.go
r0bobo: Sounds good. I'll try to cook up a test case.

r0bobo: What do you think of the latest commit? I factored out `createHTTPRoundTripper()` into its own function to make it testable without having to provide a working Clickhouse instance to connect to. The test case might seem banally simple but I couldn't think of something more clever to test. :thinking:

kavirajk: Looks good :) thanks @r0bobo.

Issue #1638, opened 2025-08-20
Inserting Null into a Nullable(JSON) Column Via A Batch Appears To Insert A Non-Null Value


d80tb7:
## Observed

- Define a Clickhouse table contianing a `Nullable(JSON)` column
- Insert a `null` value into this table using a sql batch
- Read the value back from clickhouse and see that you get a non-null value.

## Expected behaviour

- A `null` value shoudl be inserted into the table

## Code example

```go
package code

import (
	"context"
	"testing"

	"github.com/ClickHouse/clickhouse-go/v2"
	"github.com/stretchr/testify/require"
)

func TestNullableJSON(t *testing.T) {
	ctx := context.Background()

	// Adjust address/auth to match your environment.
	conn, err := clickhouse.Open(&clickhouse.Options{
		Addr: []string{"127.0.0.1:9000"},
		Auth: clickhouse.Auth{
			Database: "default",
			Username: "clickhouse",
			Password: "",
		},
	})
	require.NoError(t, err, "open clickhouse")

	// Fresh table for the test.
	_ = conn.Exec(ctx, `DROP TABLE IF EXISTS test_nullable_json`)
	err = conn.Exec(ctx, `
		CREATE TABLE test_nullable_json
		(
			id Int32,
			payload Nullable(JSON)
		)
		ENGINE = MergeTree
		ORDER BY (id)
	`)
	require.NoError(t, err, "create table")

	batch, err := conn.PrepareBatch(ctx, `INSERT INTO test_nullable_json (id, payload) VALUES (?, ?)`)
	require.NoError(t, err, "prepare batch")
	var nilMap map[string]string
	require.NoError(t, batch.Append(1, nilMap))
	require.NoError(t, batch.Send(), "batch send")

	var retrievedPayload map[string]string
	err = conn.QueryRow(ctx, `SELECT payload FROM test_nullable_json WHERE id = 1`, 2).Scan(&retrievedPayload)
	require.NoError(t, err, "select payload")

	// retrievedPayload should be nil but is instead the non-nil (an empty map)
	require.Nil(t, retrievedPayload, "payload for id=2 should be NULL (batched insert). Got: %#v", retrievedPayload)
}

```

Note that if we change the above test to use a simple statement rather than a batch, thios does work as expected.
```
conn.Exec(ctx, `INSERT INTO test_nullable_json (id, payload) VALUES (?, ?)`, 1, nil)
```

### Environment
[...]

PR #1667, opened 2025-09-22
bug: Fix Nullable(JSON) with `nil` map

kavirajk:
Fixes: #1638

The actual bug is in the `Nullable` [column here.](https://github.com/ClickHouse/clickhouse-go/blob/12088fe6b4a62f27e7cdb43d3bbebbcfccd9a5cb/lib/column/nullable.go?plain=1#L134-L136)

We check if the value is nil only if type is `reflect.Pointer`. In our case, it's `reflect.Map` (which can also be nil).

Following is verification using `clickhouse-client` with and without patch.

```
SELECT *
FROM test_nullable_json

Query id: 485b3606-fe0e-45e9-8719-f7a88e0daaf9

   ┌─id─┬─payload─┐
1. │  2 │ {}      │
   └────┴─────────┘

1 row in set. Elapsed: 0.006 sec.
```

```
SELECT *
FROM test_nullable_json

Query id: f626b8e0-e71e-4ec0-ad29-256903d4a348

   ┌─id─┬─payload─┐
1. │  2 │ ᴺᵁᴸᴸ    │
   └────┴─────────┘

1 row in set. Elapsed: 0.003 sec.
```

Notice how after the patch, the column value is populated correctly as `NULL`.

Also verified with original code in the description. We are able to get back the NULL column correctly as `nil` value. Locked the behavior via integration tests

## Checklist

### DIFF
--- a/lib/column/nullable.go
+++ b/lib/column/nullable.go
@@ -1,4 +1,3 @@
-
 package column
 
 import (
@@ -131,7 +130,7 @@ func (col *Nullable) AppendRow(v any) error {
 		rv = reflect.ValueOf(v)
 	}
 
-	if v == nil || (rv.Kind() == reflect.Pointer && rv.IsNil()) {
+	if v == nil || ((rv.Kind() == reflect.Pointer || rv.Kind() == reflect.Map) && rv.IsNil()) {
 		col.nulls.Append(1)
 		// used to detect sql.Null* types
 	} else if val, ok := v.(driver.Valuer); ok {
--- a/tests/issues/1638_test.go
+++ b/tests/issues/1638_test.go
@@ -1,0 +1,43 @@
+package issues
+
+import (
+	"context"
+	"testing"
+
+	clickhouse_tests "github.com/ClickHouse/clickhouse-go/v2/tests"
+	"github.com/stretchr/testify/require"
+)
+
+func TestIssue1638_NullableJSON(t *testing.T) {
+	ctx := context.Background()
+
+	conn, err := clickhouse_tests.GetConnectionTCP("issues", nil, nil, nil)
+	require.NoError(t, err)
+	require.NoError(t, err, "open clickhouse")
+
+	// Fresh table for the test.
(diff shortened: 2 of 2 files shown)


tests/issues/1638_test.go
```diff
@@ -0,0 +1,49 @@
...
+	if !clickhouse_tests.CheckMinServerServerVersion(conn, 25, 2, 0) {
+		// https://clickhouse.com/docs/ru/whats-new/changelog#new-feature-3
+		t.Skip(fmt.Errorf("unsupported clickhouse version. JSON not supported in Nullable"))
+		return
+	}
+
+	// Fresh table for the test.
+	_ = conn.Exec(ctx, `DROP TABLE IF EXISTS test_nullable_json`)
+	err = conn.Exec(ctx, `
+		CREATE TABLE test_nullable_json
```
chernser: please add test from this comment when there are predefined paths in JSON: 
```
Nullable(JSON(Name String, Age Int64, KeysNumbers Map(String, Int64), SKIP fake.field))
```

lib/column/nullable.go
```diff
@@ -131,7 +130,7 @@ func (col *Nullable) AppendRow(v any) error {
 		rv = reflect.ValueOf(v)
 	}
 
-	if v == nil || (rv.Kind() == reflect.Pointer && rv.IsNil()) {
+	if v == nil || ((rv.Kind() == reflect.Pointer || rv.Kind() == reflect.Map) && rv.IsNil()) {
```
chernser, line 133: will nil Slice cause similar issue?

tests/issues/1638_test.go
kavirajk: Addressed[ here](https://github.com/ClickHouse/clickhouse-go/pull/1667/commits/63b78917b4ea7a830d19f854096f562ce8f7b829)

lib/column/nullable.go
kavirajk: True. it can happen for other nested types like `Array` and `Struct`. I c[reated another issue](https://github.com/ClickHouse/clickhouse-go/issues/1668) to tackle this separately

PR #1669, opened 2025-09-25
Add support to `time` and `time64` datatypes

kavirajk:
## Summary

This is built on top of [other PR ](https://github.com/ClickHouse/clickhouse-go/pull/1596) preserving original authors commits.

## Checklist


lib/column/codegen/column.tpl
```diff
@@ -126,6 +110,10 @@ func (t Type) Column(name string, sc *ServerContext) (Interface, error) {
 		return &SharedVariant{name: name}, nil
 	case "Object('json')":
 	    return &JSONObject{name: name, root: true, sc: sc}, nil
+	case "Time":
```
SpencerTorres, line 111: It looks like the changes here are inconsistent with what was generated. Was `column_gen.go` manually edited?

Also in `column_gen.go` it looks like `Time` has a prefix check, does it need a prefix check? I looked at the docs and it looks like Time is just `Time`. The only one that needs a prefix is `Time64` since it has the precision defined

lib/column/column_gen.go
```diff
@@ -161,6 +160,10 @@ func (t Type) Column(name string, sc *ServerContext) (Interface, error) {
 		return (&DateTime64{name: name}).parse(t, sc.Timezone)
 	case strings.HasPrefix(strType, "DateTime") && !strings.HasPrefix(strType, "DateTime64"):
 		return (&DateTime{name: name}).parse(t, sc.Timezone)
+	case strings.HasPrefix(strType, "Time64"):
+		return (&Time64{name: name}).parse(t)
+	case strings.HasPrefix(strType, "Time"):
```
SpencerTorres: See comment on `column.tpl`

lib/column/time.go
```diff
@@ -0,0 +1,308 @@
...
+	col.col.Reset()
+}
+
+func (col *Time) Name() string {
+	return col.name
+}
+
+func (col *Time) parse(t Type) (_ Interface, err error) {
+	col.chType = t
+	if string(t) == "Time" {
```
SpencerTorres: Not sure if `Time` needs a parse function at all, you can probably initialize it without parsing. The reason we have parsing is for types like `Tuple` and `Time64` where other syntax is included within the type string.

We don't need to return unsupported type error here since the column initialization function would never call `parse` if the string isn't already equal to `Time`

lib/column/time.go
```diff
@@ -0,0 +1,308 @@
...
+		*d = new(time.Time)
+		**d = col.row(row)
+	case *int64:
+		t := col.row(row)
+		*d = timeToSeconds(t)
+	case **int64:
+		*d = new(int64)
+		t := col.row(row)
+		**d = timeToSeconds(t)
+	case *sql.NullTime:
```
SpencerTorres: I see `NullTime` here but it makes me wonder if there's a regular SQL `Time` type or if that's just the Go type

lib/column/time.go
```diff
@@ -0,0 +1,308 @@
...
+		}
+	}
+	return
+}
+
+func (col *Time) AppendRow(v any) error {
+	switch v := v.(type) {
+	case int64:
+		seconds := v
+		hours := seconds / 3600
```
SpencerTorres: It looks like this math is copied around a bunch. Can we make a function or struct that does it? Maybe a function with multiple return values?

lib/column/time.go
```diff
@@ -0,0 +1,308 @@
...
+			minutes := (seconds % 3600) / 60
+			secs := seconds % 60
+			col.col.Append(proto.FromTime32(time.Date(1970, 1, 1, int(hours), int(minutes), int(secs), 0, time.UTC)))
+		default:
+			col.col.Append(proto.FromTime32(time.Time{}))
+		}
+	case time.Time:
+		col.col.Append(proto.FromTime32(v))
+	case *time.Time:
+		switch {
```
SpencerTorres: Strange use of `switch` here, is this not just an `if`? Is this used anywhere else in this file or `Time64`?

lib/column/time.go
```diff
@@ -0,0 +1,308 @@
...
+		case v != nil:
+			col.col.Append(proto.FromTime32(*v))
+		default:
+			col.col.Append(proto.FromTime32(time.Time{}))
+		}
+	case sql.NullTime:
+		if v.Valid {
+			col.col.Append(proto.FromTime32(v.Time))
```
SpencerTorres: If `v` isn't valid does that mean it defaults to `time.Time{}`? Do we need this `if` at all?

lib/column/time.go
```diff
@@ -0,0 +1,308 @@
...
+	return col.col.Row(i).ToTime32()
+}
+
+func (col *Time) parseTime(value string) (tv time.Time, err error) {
+	value = strings.TrimSpace(value)
+	if value == "" {
+		return time.Time{}, nil
+	}
+
+	formats := []string{
```
SpencerTorres: I wonder how likely it is that the format is different between `parseTime` calls. Maybe after the first successful format match we can set it for the column instance, sort of caching it? Then subsequent appends won't have to loop through.

Just an idea, probably not necessary. Safest to treat every row as a possibly different format

lib/column/time64.go
```diff
@@ -0,0 +1,311 @@
...
+	return col.col.Row(i).ToTime()
+}
+
+func (col *Time64) parseTime(value string) (tv time.Time, err error) {
+	value = strings.TrimSpace(value)
+	if value == "" {
+		return time.Time{}, nil
+	}
+
+	formats := []string{
```
SpencerTorres: Is this list/function duplicated from `Time`? Perhaps Time64 can borrow some helper functions from Time

lib/proto/block.go
```diff
@@ -213,9 +213,12 @@ func (b *Block) Decode(reader *proto.Reader, revision uint64) (err error) {
 				return err
 			}
 			if hasCustom {
-				return &BlockError{
-					Op:  "Decode",
-					Err: errors.New(fmt.Sprintf("custom serialization for column %s. not supported by clickhouse-go driver", columnName)),
+				// Allow Time and Time64 columns with custom serialization
```
SpencerTorres: I am so incredibly suspicious of this line, I saw it in the first PR and was concerned. Is this really the first column type that has this `hasCustom` flag set?

I don't recall if I wrote it in a comment on that PR but perhaps we can have an interface on the column type to declare whether it `hasCustom` or not, because I don't like the idea of these types being hardcoded here.

Does this even match on `Time64(3)` ? Is this necessary? Let's ask the core team to confirm, or maybe try a hexdump of Native format to see if this flag is really being set.

tests/time_test.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func setupTimeTest(t *testing.T, protocol clickhouse.Protocol) (context.Context, func(), clickhouse.Conn) {
+	ctx := context.Background()
+	conn, err := GetNativeConnection(t, protocol, map[string]any{
+		"enable_time_time64_type": 1,
+	}, nil, nil)
+	require.NoError(t, err)
+	if !CheckMinServerServerVersion(conn, 25, 6, 0) {
```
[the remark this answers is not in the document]
SpencerTorres: I think I've seen this before but was too deep in a different PR to fix it. Let's correct this typo at some point. Should be a quick and easy refactor

TYPES.md
```diff
@@ -74,9 +74,53 @@ All types can be read into a pointer or pointer to a pointer.
...
+## Time and Time64 Example
+
+```go
+// Insert and read Time and Time64 columns
+colTime := &column.Time{name: "t"}
+colTime64 := &column.Time64{name: "t64"}
+colTime64.col.WithPrecision(9)
+
+// Append values
+colTime.AppendRow(time.Date(2024, 7, 11, 12, 34, 56, 0, time.UTC))
```
SpencerTorres: Let's make sure the docs are super clear about how the date component + timezone behaves. It's not immediately clear from here that it starts at 1970 or whatever we talked about

tests/time_test.go
```diff
@@ -0,0 +1,87 @@
...
+	ctx := context.Background()
+	conn, err := GetNativeConnection(t, protocol, nil, nil, nil)
+	require.NoError(t, err)
+	if !CheckMinServerServerVersion(conn, 25, 6, 0) {
+		t.Skip("Time/Time64 not supported on this ClickHouse version")
+	}
+	return ctx, func() {}, conn
+}
+
+func TestTimeAndTime64(t *testing.T) {
```
SpencerTorres: For the sake of test coverage and isolating issues with these types, I would suggest splitting these into separate tests, maybe even separate files since `Time` is different than `Time64`. I would use Date/DateTime/DateTime64 as reference (unless those are all in one test too, haven't looked recently)

Also considering the feature is experimental, the setup function is the correct way to go, glad to see it being used.

Also also, and I know it's annoying to add, but we need to add coverage for the standard library `database/sql` tests. These are in a different folder and are very similar to these tests. Be aware that if you're testing HTTP with the standard library connector it may need you to set the `session_id` if you're modifying the experimental setting.

I would reference Variant/Dynamic/JSON as well as DateTime/DateTime64 for these. You can copy/paste/rework the stdlib tests from that folder. Make sure we have all the right sanity checks for things like arrays, matching input/output, etc.

SpencerTorres (changes requested):
Overall looks good, left some comments for things I had questions on. I understand this PR was copied from a different author. I appreciate your help ✨  enhancing ✨  it.

Let me know if you have any questions 👌

lib/column/time.go
kavirajk: It's part of `sql` package. It think it is there to handle cases like
```go
var s sql.NullTime
db.QueryRow("SELECT time FROM foo WHERE id=?", id).Scan(&s)

if s.Valid() { // non-null
  // do something
}
```

Similar to how we already handle for [string type in clickhouse-go](https://github.com/ClickHouse/clickhouse-go/blob/f7bc653a8e88798eba41bedaba67b01a8417ee5c/lib/column/string.go?plain=1#L58)

lib/column/time.go
kavirajk: Strange indeed. could have been simple if else. Looks like copied from other places (like [strings.go](https://github.com/ClickHouse/clickhouse-go/blob/f7bc653a8e88798eba41bedaba67b01a8417ee5c/lib/column/string.go?plain=1#L90-L95))

I can fix it on time and time64 in this PR and have a separate PR to clean this up in other places. wdyt?

lib/column/time.go
kavirajk: I think we need `if` or else irrespective of whether the time is NULL or not, we will get zero value `time.Time{}`?

Basically, if v is non-null, then we need actual time value (`v.Time`) not zero value `time.Time{}`. Anything I'm missing?

lib/column/time.go
kavirajk: On the second thought. Sure it can be replaced with `if` and `else` but it just works with switch as it is. I'm fine to leave it as is. But if you think it improves the readability happy to change it :)

TYPES.md
kavirajk: The more I think about this, the more I realize, `time.Time` is not the correct Go type for this `time` and `time64` CH type. Unwanted complexities with `Date` (year, month, day, timezone) aspect of it.

The ideal Go type would be I think is `time.Duration`. Which includes only `time` without `date` component.

1. This exactly align with time and time64[ docs](https://clickhouse.com/docs/sql-reference/data-types/time64)
2. Also Rust library[ uses time::Duration](https://github.com/ClickHouse/clickhouse-rs/blob/main/examples/time_types_example.rs#L83-L116)

also Go definition of time.Duration seems to reflect the same purpose.

```
const (
	Nanosecond  Duration = 1
	Microsecond          = 1000 * Nanosecond
	Millisecond          = 1000 * Microsecond
	Second               = 1000 * Millisecond
	Minute               = 60 * Second
	Hour                 = 60 * Minute
)
```

We only have to make sure that `time` has precision max up to only `milliseconds`. We don't have to fight with default values for `year`, `month`, `day` and `timezone` aspect of it. Which can leads to some subtle logical errors depends on how users use those apis.

I think sticking to "**Make your APIs easy to use correctly, and make it hard to use in-correctly"** philosophy will help here :)

wdyt @SpencerTorres ?

kavirajk: Looks like it doesn't need lots of change(may be i'm wrong 😆 ). In fact may lead to removing lots of code from both ch-go and clickhouse-go. I'm going to try with `time.Duration` type and see what happens.

SpencerTorres: This sounds much more correct, let's get ch-go updated too then. ch-go technically allows for these breaking changes with its versioning.

I'd say take all the code from here and ch-go and make it use `time.Duration`, it is much more fitting as you said.

go.mod
```diff
@@ -5,7 +5,7 @@ go 1.24.0
 toolchain go1.24.1
 
 require (
-	github.com/ClickHouse/ch-go v0.68.0
+	github.com/ClickHouse/ch-go v0.68.1-0.20251003160956-f6cfd63ac24c
```
kavirajk: FYI: This points to this [PR commit in ch-go.](https://github.com/ClickHouse/ch-go/pull/1097) Once it's merged we can point `main` (or after releasing point to specific newer version)

tests/time_test.go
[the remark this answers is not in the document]
kavirajk: Ah thanks for the context. I will do it in follow up PR :) (this PR is already big enough I think)

go.mod
ernado: Merged

SpencerTorres (approved):
Looks usable, in a followup PR we should add tests for the stdlib. They're usually the same, but there's some quirks sometimes

Issue #1685, opened 2025-10-15
bug(performance-memory): Unnecessary memory allocations from reading profile events


erezrokah:
## Observed

ClickHouse Go driver still scans profile events even when a `profileEvents` listener is not set, causing unnecessary memory allocations

## Expected behaviour

Ideally profile events could be disabled via a session setting, but `log_profile_events: 0` doesn't seem to help here. Even if `log_profile_events` is set to `1`, the Go driver shouldn't scan events if there's no listener.

## Code example

Example test

```
func benchmarkProfileEvents(ctx context.Context, conn clickhouse.Conn) error {
	for i := 0; i < 10_000; i++ {
		err := conn.Exec(ctx, fmt.Sprintf(`INSERT INTO test_profile_events VALUES (
			%d, '%s', [1, 2, 3, 4, 5, 6, 7, 8, 9], now()
		)`, i, "Golang SQL database driver"), false)
		if err != nil {
			return err
		}
	}
	return nil
}

func BenchmarkProfileEvents(b *testing.B) {
	conn, err := tests.GetConnectionTCP("issues", nil, nil, nil)
	ctx := context.Background()
	require.NoError(b, err)

	const ddl = `CREATE TABLE test_profile_events (Col1 UInt64, Col2 String, Col3 Array(UInt8), Col4 DateTime) Engine ReplacingMergeTree() ORDER BY Col1`
	err = conn.Exec(ctx, ddl)
	require.NoError(b, err)
	defer func() {
		conn.Exec(ctx, "DROP TABLE IF EXISTS test_profile_events")
	}()

	for k := 0; k < b.N; k++ {
		require.NoError(b, benchmarkProfileEvents(ctx, conn))
	}
}
```

### Environment
* [x] `clickhouse-go` version: v2.40.3
* [x] Interface: ClickHouse API
* [x] Go version: 1.25.1
* [x] Operating system: MacOS 26.0.1
* [x] ClickHouse version: 25.6
* [x] Is it a ClickHouse Cloud? No
* [x] ClickHouse Server non-default settings, if any: settings are based on the tests suite in this repo
* [x] `CREATE TABLE` statements for tables involved: See test above

### Memory profile showing allocations

Benchmark results

```
BenchmarkProfileEvents-12                  1        11126073667 ns/op       276147480 B/op   2349943 allocs/op
```

[...]

PR #1686, opened 2025-10-15
fix: Don't scan profile events if listener is not set

erezrokah:
## Summary

Fixes https://github.com/ClickHouse/clickhouse-go/issues/1685. A better way would be to be able to disable these events altogether from being returned to the client.

Also, if this PR is accepted we might want to do the same for other listeners, though I haven't seen a big issue with allocations in other places (see [issue](https://github.com/ClickHouse/clickhouse-go/issues/1685) for memory profiles).

For the benchmark in the issue this is over a **3x improvement**

## Checklist

### DIFF
--- a/tests/issues/1685_test.go
+++ b/tests/issues/1685_test.go
@@ -2,7 +2,6 @@ package issues
 
 import (
 	"context"
-	"fmt"
 	"testing"
 
 	"github.com/ClickHouse/clickhouse-go/v2"


conn_process.go
```diff
@@ -185,11 +184,14 @@ func (c *connect) handle(ctx context.Context, packet byte, on *onProcess) error
...
-		events, err := c.profileEvents(ctx)
+		scanEvents := on.profileEvents != nil
+		events, err := c.profileEvents(ctx, scanEvents)
 		if err != nil {
 			return err
 		}
-		on.profileEvents(events)
+		if scanEvents {
+			on.profileEvents(events)
+		}
```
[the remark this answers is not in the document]
erezrokah: > this condition will never execute with a nil handler

This is incorrect, if `scanEvents` is `false` `on.profileEvents` is `nil` and the call will panic (double checked by removing the condition and re-running the test)

conn_profile_events.go
```diff
@@ -18,12 +17,16 @@ type ProfileEvent struct {
...
 
-func (c *connect) profileEvents(ctx context.Context) ([]ProfileEvent, error) {
+func (c *connect) profileEvents(ctx context.Context, scanEvents bool) ([]ProfileEvent, error) {
 	block, err := c.readData(ctx, proto.ServerProfileEvents, false)
 	if err != nil {
 		return nil, err
 	}
 	c.debugf("[profile events] rows=%d", block.Rows())
+	if !scanEvents {
+		c.debugf("[profile events] skipping scan")
```
kavirajk: can we add also why we skip? may be something like
`[profile events] skipping scan. No profile-events handler registered`

conn_process.go
kavirajk, line 194:
```suggestion
		if on.profileEvents != nil {
			events, err := c.profileEvents(ctx)
			if err != nil {
				return err
			}
			on.profileEvents(events)
			break
		}
		c.debugf("[skip profile events. no handler registered]")

```

kavirajk: This way we don't have to change the `c.profileEvents` method. We just skip on the consumer side itself. what do you think?

conn_profile_events.go
```diff
@@ -18,12 +17,16 @@ type ProfileEvent struct {
 	Value       int64
 }
 
-func (c *connect) profileEvents(ctx context.Context) ([]ProfileEvent, error) {
+func (c *connect) profileEvents(ctx context.Context, scanEvents bool) ([]ProfileEvent, error) {
```
kavirajk, line 20: We may not need to change this `profileEvents` api at all? . We can skip calling it from the consumer side. See my other comment.

conn_process.go
erezrokah: Thanks for the comment @kavirajk, I tried that initially but it appears we still need to read the block from the server in `c.profileEvents(ctx)` (we still need to call `c.readData(ctx, proto.ServerProfileEvents, false)`).

If we could do something like this in queries

```
....

SETTINGS PROFILE_EVENTS_DISABLED=1
```

That would be the best I think, but this is probably a bigger change and needs to be done in https://github.com/ClickHouse/ClickHouse

conn_profile_events.go
erezrokah: Replied, I did that initially but it appears we have to consume the data from the server otherwise the querying fails

conn_process.go
erezrokah: Regardless of having an option to disable profile events in the session, I think it's good for the client to skip scanning them if the listener is not registered

kavirajk: I agree! thanks for the clarification @erezrokah :+1: I don't think we have to block this PR for the server settings to be available.

PR #1687, opened 2025-10-15
refactor: idle pool and fix non-deterministic teardown of idle cleanup goroutine

GeorgeMac:
## Summary

Closes: #1702 

This PR does a couple things:

1. Addresses a leak of `startAutoCloseIdleConnections` goroutines
2. Refactors the idle pool to maintain connections in a circular queue (buffer)

We observed in the CH operator that `startAutoCloseIdleConnections` goroutines were being leaked.
We found in one long running operator, 35k instances of this goroutine running unchecked.
These get leaked if `conn.Close()` is not called, however, it appears that the client implementation can also leak them in the situation that the receiver on `ch.exit` is not ready when `conn.Close()` is called. Since it does so with a select and a default, which is best effort and needs a receiver to be ready.

A quick test with the new `testing/synctest` package showed that this was very easy to demonstrate by opening a connection and then immediately closing it. Now that the repo has been updated with 1.25 toolchain, I have included these in the tests.

These tests fail deterministically like this:

```
--- FAIL: TestIdleConnectionCleanupLeak (0.00s)
panic: deadlock: main bubble goroutine has exited but blocked goroutines remain [recovered, repanicked]

goroutine 101 [running]:
testing.tRunner.func1.2({0x1024ab740, 0x140001101b0})
        /opt/homebrew/Cellar/go/1.25.2/libexec/src/testing/testing.go:1872 +0x190
testing.tRunner.func1()
        /opt/homebrew/Cellar/go/1.25.2/libexec/src/testing/testing.go:1875 +0x31c
panic({0x1024ab740?, 0x140001101b0?})
        /opt/homebrew/Cellar/go/1.25.2/libexec/src/runtime/panic.go:783 +0x120
internal/synctest.Run(0x140001227a0)
        /opt/homebrew/Cellar/go/1.25.2/libexec/src/runtime/synctest.go:251 +0x2c4
testing/synctest.Test(0x14000132380, 0x1024ebdb0)
        /opt/homebrew/Cellar/go/1.25.2/libexec/src/testing/synctest/synctest.go:282 +0x88
github.com/ClickHouse/clickhouse-go/v2.TestIdleConnectionCleanupLeak(0x14000132380?)
        /Users/georgemacrorie/github/ClickHouse/clickhouse-go/clickhouse_test.go:16 +0x24
testing.tRunner(0x14000132380, 0x1024ebc90)
        /opt/homebrew/Cellar/go/1.25.2/libexec/src/testing/testing.go:1934 +0xc8
created by testing.(*T).Run in goroutine 1
        /opt/homebrew/Cellar/go/1.25.2/libexec/src/testing/testing.go:1997 +0x364

goroutine 104 [select (durable), synctest bubble 1]:
github.com/ClickHouse/clickhouse-go/v2.(*clickhouse).startAutoCloseIdleConnections(0x14000112e40)
        /Users/georgemacrorie/github/ClickHouse/clickhouse-go/clickhouse.go:334 +0x9c
created by github.com/ClickHouse/clickhouse-go/v2.Open in goroutine 103
        /Users/georgemacrorie/github/ClickHouse/clickhouse-go/clickhouse.go:76 +0x158
FAIL    github.com/ClickHouse/clickhouse-go/v2  0.696s
FAIL
```

After this change, this test passes.

## Additional Refactor

> I originally attempted this with a min heap for connection age, but that was both over complicated likely not what we need.
> I have since changed it to use a circular queue instead.
> So if you see our conversation below and it doesn't make sense, this is likely why.

To tackle this, I wanted to revisit the idle pool, which so far has used a channel for holding connections.
I think the channel works OK, however, it has some drawbacks:
1. You can only synchronize adding and removing entries (i.e. additional locking is still required to synchronize closing on teardown)
2. It is awkward to clear out expired connections, as you have to empty and refill it every time.

Instead, I have proposed here a new idle pool and circular queue implementation.

### Circular Queue

The circular queue is a ring with a max capacity and constant memory usage (it avoids resizing).
Pushing and pulling from the queue wraps items around in the slice under the hood.
When the queue is full, any attempts to push new entries are rejected.
https://github.com/GeorgeMac/clickhouse-go/blob/71a7247f7421d7939fd6b9ead2a3808121763f1f/internal/circular/queue.go#L40-L51

The queue supports clearing elements based on a filter function (which is used for clearing expired connections).
This returns an iterator over the items which were cleared (we use this later for closing cleared connections).
The queue does not protect against concurrent access, it is expected that the caller will handle that.
https://github.com/GeorgeMac/clickhouse-go/blob/71a7247f7421d7939fd6b9ead2a3808121763f1f/internal/circular/queue.go#L86-L93

There is a set of unit tests and benchmarks focused on attempting to validate these semantics and memory guarantees.

### Idle Pool

The idle pool focuses on supporting concurrent access to connections in the backing queue, background cleanup of expired connections and safe cleanup of these processes on shutdown (to address the original idle cleanup goroutine problem).

It exposes a Get and a Put method for managing connections in the pool.
Put handles automatically dropping connections which are expired or when capacity in the pool is reached.
Get removes connections from the queue, until it finds a non-expired one or the queue is empty.
For each already expired one it comes across, it closes it.

The cleanup routine periodically attempts to clear connections which have expired.
It blocks until either its ticker tics, or it is signaled to finish, in which case it returns.

Close handles signalling the cleanup routine to finish, then waiting for it to do so.
Once the cleanup routine is finished, it goes through and clears all connections in the pool.

### Changes to `type clickhouse struct {}`

Primarily, this swaps the channel used for idle connections with the new `*idlePool`.
This removes the need for the idle cleanup from this part of the code.
Now both acquire and release interface with the idle pool to attempt to fetch existing idle connections and return them.
There is also a little further simplification such as leaning into context timeouts for some dial timeout handling.

Let me know what you think 🙏 

## Checklist

### DIFF
--- a/.github/workflows/check-pr-links.yml
+++ b/.github/workflows/check-pr-links.yml
@@ -1,0 +1,32 @@
+name: check PR Links
+
+on:
+  pull_request:
+    branches: [main]
+  workflow_dispatch:
+
+jobs:
+  linkChecker:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Check links
+        uses: lycheeverse/lychee-action@v2
+        with:
+          # Remap live URLs to build directory because the links are potentially not live (not yet on master)
+          args: |
+            --verbose 
+            --no-progress
+            --exclude-all-private
+            --retry-wait-time 5
+            ./
+          fail: true
+
+      - name: Suggestions
+        if: failure()
+        run: |
+          echo -e "\nPlease review the links reported in the 'Check links' step above."
+          echo -e "If a link is valid but fails due to a CAPTCHA challenge, IP blocking, login requirements, etc.,"
+          echo -e "consider adding such links to .lycheeignore file to bypass future checks.\n"
+          exit 1
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,25 +9,32 @@ jobs:
   publish:
     runs-on: ubuntu-latest
     steps:
+      - name: Generate Token
+        id: generate-token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ secrets.WORKFLOW_AUTH_PUBLIC_APP_ID }}
+          private-key: ${{ secrets.WORKFLOW_AUTH_PUBLIC_PRIVATE_KEY }}
+
       - uses: actions/checkout@main
         with:
           ref: main
           fetch-depth: 0 # required to generate contributor list
+          token: ${{ steps.generate-token.outputs.token }}
 
       - name: Install Go
         uses: actions/setup-go@v5
         with:
-          stable: false
           go-version: 1.25
 
       - name: Prepare release
         id: release
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ steps.generate-token.outputs.token }}
         run: |
           go run internal/cmd/release/main.go
 
-      - name: Commit and push hanges
+      - name: Commit And Push Changes
         run: |
           git config --local user.email "ch-integrations-robot@clickhouse.com"
           git config --local user.name "GitHub Action"
@@ -35,8 +42,8 @@ jobs:
           git commit -m "Update release notes"
           git push
 
-      - name: Publish draft release
+      - name: Publish Draft Release
         env:
           RELEASE_API_URL: ${{ steps.release.outputs.RELEASE_URL }}
         run: |
-            curl -X PATCH -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" -H "Accept: application/vnd.github.v3+json" $RELEASE_API_URL -d '{"draft": false}'
+            curl -X PATCH -H "Authorization: token ${{ steps.generate-token.outputs.token }}" -H "Accept: application/vnd.github.v3+json" $RELEASE_API_URL -d '{"draft": false}'
--- a/.lycheeignore
+++ b/.lycheeignore
@@ -1,0 +1,1 @@
+https://docs.google.com/document/d/1924Dvy79KXIhfqKpi1EBVY3133pIdoMwgCQtZ-uhEKs/edit#heading=h.ah33hoz5xei2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,52 @@
+# v2.41.0, 2025-11-20 <!-- Release notes generated using configuration in .github/release.yml at main -->
+
+## What's Changed
+
+### Enhancements :tada:  
+* feat: add custom http transport by @r0bobo in https://github.com/ClickHouse/clickhouse-go/pull/1656
+* Support interface typed NamedValue binding by @crisismaple in https://github.com/ClickHouse/clickhouse-go/pull/1633
+* remove deprecated Object('json') column by @kavirajk in https://github.com/ClickHouse/clickhouse-go/pull/1674
+* feat: OpenDB api now sets few configs passed via options by @kavirajk in https://github.com/ClickHouse/clickhouse-go/pull/1675
+* chore: Async inserts docs and example updates by @kavirajk in https://github.com/ClickHouse/clickhouse-go/pull/1690
+* Set HttpUrlPath from DSN to resolve missing path in HTTP requests by @RokibulHasan7 in https://github.com/ClickHouse/clickhouse-go/pull/1692
+* Add support to `time` and `time64` datatypes by @kavirajk @shivanshuraj1333  in https://github.com/ClickHouse/clickhouse-go/pull/1669
+* fix: Don't scan profile events if listener is not set by @erezrokah in https://github.com/ClickHouse/clickhouse-go/pull/1686
+* Allow oneline comment in insert query by @luckyevildev in https://github.com/ClickHouse/clickhouse-go/pull/1693
+* Update cloud version matrix (25.10) by @SpencerTorres in https://github.com/ClickHouse/clickhouse-go/pull/1704
+* Disable default transport layer gzip compression by default by @kavirajk https://github.com/ClickHouse/clickhouse-go/pull/1704/commits/5c8c388b7e19b6be4ca101624d7ae3b02bf15a05 and https://github.com/ClickHouse/clickhouse-go/pull/1704#issuecomment-3502818316
+* Add some examples for ephemeral column by @kavirajk in https://github.com/ClickHouse/clickhouse-go/pull/1710
+
+
+### Bug fixes :bug: 
+* Fix potential Divide by Zero panic by @kavirajk in https://github.com/ClickHouse/clickhouse-go/pull/1665
+* fix: SimpleAggregateFunction(anyLast, JSON) by @bripkens in https://github.com/ClickHouse/clickhouse-go/pull/1645
+* bug: Fix Nullable(JSON) with `nil` map by @kavirajk in https://github.com/ClickHouse/clickhouse-go/pull/1667
+* bug: Fixes some correctness bug with ContextWatchdog by @kavirajk in https://github.com/ClickHouse/clickhouse-go/pull/1688
+
+### Other Changes 🛠
+* chore: Remove unused bash script by @kavirajk in https://github.com/ClickHouse/clickhouse-go/pull/1663
+* chore: Remove license header from each go file by @kavirajk in https://github.com/ClickHouse/clickhouse-go/pull/1662
+* Bump github.com/paulmach/orb from 0.11.1 to 0.12.0 by @dependabot[bot] in https://github.com/ClickHouse/clickhouse-go/pull/1664
+* Bump github.com/testcontainers/testcontainers-go from 0.38.0 to 0.39.0 by @dependabot[bot] in https://github.com/ClickHouse/clickhouse-go/pull/1666
+* Bump github.com/docker/docker from 28.4.0+incompatible to 28.5.0+incompatible by @dependabot[bot] in https://github.com/ClickHouse/clickhouse-go/pull/1676
+* Bump github.com/docker/docker from 28.5.0+incompatible to 28.5.1+incompatible by @dependabot[bot] in https://github.com/ClickHouse/clickhouse-go/pull/1679
+* Bump golang.org/x/net from 0.44.0 to 0.46.0 by @dependabot[bot] in https://github.com/ClickHouse/clickhouse-go/pull/1683
+* Bump github.com/ClickHouse/ch-go from 0.68.0 to 0.69.0 by @dependabot[bot] in https://github.com/ClickHouse/clickhouse-go/pull/1678
+* chore: Update contributing guide and docker compose stack by @erezrokah in https://github.com/ClickHouse/clickhouse-go/pull/1684
+* chore: Upgrade Go toolchain to 1.25.x by @kavirajk in https://github.com/ClickHouse/clickhouse-go/pull/1689
+* Bump github.com/docker/docker from 28.5.1+incompatible to 28.5.2+incompatible by @dependabot[bot] in https://github.com/ClickHouse/clickhouse-go/pull/1706
+* Bump github.com/testcontainers/testcontainers-go from 0.39.0 to 0.40.0 by @dependabot[bot] in https://github.com/ClickHouse/clickhouse-go/pull/1705
+* Bump golang.org/x/net from 0.46.0 to 0.47.0 by @dependabot[bot] in https://github.com/ClickHouse/clickhouse-go/pull/1711
+
+## New Contributors
+* @kavirajk made their first contribution in https://github.com/ClickHouse/clickhouse-go/pull/1663
+* @bripkens made their first contribution in https://github.com/ClickHouse/clickhouse-go/pull/1645
+* @erezrokah made their first contribution in https://github.com/ClickHouse/clickhouse-go/pull/1684
+* @RokibulHasan7 made their first contribution in https://github.com/ClickHouse/clickhouse-go/pull/1692
+* @luckyevildev made their first contribution in https://github.com/ClickHouse/clickhouse-go/pull/1693
+
+**Full Changelog**: https://github.com/ClickHouse/clickhouse-go/compare/v2.40.3...v2.41.0
+
 # v2.40.3, 2025-09-13 <!-- Release notes generated using configuration in .github/release.yml at main -->
 
 ## What's Changed
@@ -291,7 +340,7 @@ HTTP still has some limitations to be aware of for things like batch flushing an
 ## New Contributors
 * @serprex made their first contribution in https://github.com/ClickHouse/clickhouse-go/pull/1493
 
-**Full Changelog**: https://github.com/ClickHouse/clickhouse-go/compare/v2.30.3...2.31.0
+**Full Changelog**: https://github.com/ClickHouse/clickhouse-go/compare/v2.30.3...v2.31.0
 
 # v2.30.3, 2025-02-03 <!-- Release notes generated using configuration in .github/release.yml at main -->
 
@@ -759,7 +808,7 @@ v2.25.0 was released with a breaking change in https://github.com/ClickHouse/cli
 * @alrs made their first contribution in https://github.com/ClickHouse/clickhouse-go/pull/1081
 * @testwill made their first contribution in https://github.com/ClickHouse/clickhouse-go/pull/1080
 
-**Full Changelog**: https://github.com/ClickHouse/clickhouse-go/compare/v2.13.4...v2.14
+**Full Changelog**: https://github.com/ClickHouse/clickhouse-go/compare/v2.13.4...v2.14.0
 
 # v2.13.4, 2023-08-30 <!-- Release notes generated using configuration in .github/release.yml at main -->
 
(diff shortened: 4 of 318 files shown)


GeorgeMac:
@kavirajk we could definitely go after a simpler fix that keeps it more in-line with what is already here if that is preferred 👍 

What you proposed works for ensuring that we stop leaking `startAutoCloseIdleConnections`, which is great 💯 

However, one thing to keep in mind here is that is doesn't guarantee to drain the idle pool on shutdown, which leaves the real connections unclosed.

The `select` there with a `default`, which receives on the idle channel. is also best-effort and if it doesn't have something in it at that time, then the default clause will run at the send to exit will happen. However, as soon as the `exit` channel send happens and it synchronizes with the receiver in `startAutoCloseIdleConnections`, a concurrent call to `release()` could put another connection in the idle pool in this moment.

Now we have released connections which will never be closed, because auto close idle routine is gone.

We could go a step further with your proposal there and attempt to drain the idle channel.
However, we need to also be sure that any concurrent calls which release active connectons will not attempt to put those back in the idle pool. We need that guarantee before we do one last drain of the pool.

This is also something I have addressed in my implementation, and I suspect that the existing idle connections counting tests in the `tests` directory will actually show this is necessary. Because they count open connections on the CH server side, which require all these connections in the idle pool have explicitly had `close` called on them when drained.

You are right, the ordering by age is not necessary to solve the problem, and I am happy to remove it if it is not favourable. However, I would challenge the decision to use a channel as an idle pool, and would instead suggest we would be better served with a slice and a lock for this. Much like how [database/sql.DB](https://cs.opensource.google/go/go/+/refs/tags/go1.25.3:src/database/sql/sql.go;l=517-518) manages its connection pool.

This is effectively all I have done, I am just using `container/heap` to maintain an order based on connection age too. Maybe my mistake was leaning on that as the unique selling point 😅 WDYT? Would you still rather try and make a channel based pool work for all cases? (there is certainly a way to get there) or could we explore this approach?

GeorgeMac:
Sorry @kavirajk in an effort to remove complexity, I made the change bigger 😅

The change I made breaks the logic up into three separate parts:

1. The existing top-level pool for opening new connections and retrieving existing idle ones
2. The idle pool for safe concurrent access to idle connections and cleanup of expired ones
3. A circular queue/buffer implementation with fixed capacity, that rejects inserts when full

I updated the description above to go into details.

While bigger, I hope the implementation is more digestible in some ways (or maybe not, you tell me), and with far more tests than before.
If it is still too much, we can just ignore it all and go ahead make the channel work.
I am appreciate this is me scratching an itch 😅 It's your call 👍

kavirajk:
No worries @GeorgeMac. Thanks for trying to fix this issue. I will take a deep look today :)

internal/circular/queue.go
```diff
@@ -0,0 +1,162 @@
...
+	var zero T
+	q.data[q.head] = zero
+	q.head = q.next(q.head)
+	q.len--
+	return value, true
+}
+
+// All returns an iterator over all elements in the queue in FIFO order.
+// The iterator yields (index, value) pairs where index is 0-based from the head.
+func (q *Queue[T]) All() iter.Seq2[int, T] {
```
kavirajk: Do we really need `All()` api. I see it's only getting used on tests?

internal/circular/queue.go
```diff
@@ -0,0 +1,162 @@
...
+			}
+			current = q.next(current)
+		}
+	}
+}
+
+// Compact removes elements from the queue based on a predicate function.
+// Returns an iterator over the removed elements.
+// Elements for which shouldRemove returns true are removed from the queue.
+func (q *Queue[T]) Compact(shouldRemove func(T) bool) iter.Seq[T] {
```
kavirajk: should we call it `Filter()` (as in `map`, `filter`). That fits more I think. `Compact` seems to be merging two queues which is not actually happening here. WDYT?

pool.go
```diff
@@ -0,0 +1,172 @@
...
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	"github.com/ClickHouse/clickhouse-go/v2/internal/circular"
+)
+
+type idlePool struct {
```
kavirajk: should we call it "connPool" instead? Also rename file from `pool.go` -> `conn_pool.go`?

conn_pool.go
```diff
@@ -0,0 +1,172 @@
...
+func newIdlePool(lifetime time.Duration, capacity int) *idlePool {
+	pool := &idlePool{
+		conns:           circular.New[nativeTransport](capacity),
+		ticker:          time.NewTicker(lifetime),
+		finish:          make(chan struct{}),
+		finished:        make(chan struct{}),
+		maxConnLifetime: lifetime,
+	}
+
+	go pool.runDrainPool()
```
kavirajk, line 34: can we also add a test to make sure we don't leak this goroutine after `pool.Close()` ?

conn_pool.go
```diff
@@ -0,0 +1,172 @@
...
+			// context has been cancelled
+			return nil, context.Cause(ctx)
+		}
+
+		if i.closed() {
+			return nil, ErrConnectionClosed
+		}
+
+		// Try to pull a connection
+		conn, ok := i.conns.Pull()
```
kavirajk, line 74: I think this a bit crutial step, because if queue didn't return false on empty, this loop can go forever. Can we write a test for this to lock the behavior? wdyt?

clickhouse.go
```diff
@@ -286,87 +293,46 @@ func DefaultDialStrategy(ctx context.Context, connID int, opt *Options, dial Dia
...
-			conn.close()
-			if conn, err = ch.dial(ctx); err != nil {
-				select {
-				case <-ch.open:
-				default:
-				}
-				return nil, err
-			}
+
+	conn, err = ch.idle.Get(ctx)
```
kavirajk, line 313: I really like the simplified version of `acquire` :)

clickhouse.go
```diff
@@ -401,28 +367,19 @@ func (ch *clickhouse) release(conn nativeTransport, err error) {
 		conn.freeBuffer()
 	}
 
-	select {
-	case ch.idle <- conn:
-	default:
-		conn.debugf("[close: idle pool full %d/%d]", len(ch.idle), cap(ch.idle))
+	if ch.closed.Load() {
 		conn.close()
+		return
```
kavirajk, line 376: wondering, if we should release the open slot by reading it from `<- ch.open` before `return`. 

Because we are [writing to `ch.open` in line 351](https://github.com/ClickHouse/clickhouse-go/blob/a35898e6c634686253c14adc7f24349994fcb2f2/clickhouse.go?plain=1#L350-L353). But I don't see anywhere we release it. Am I missing anything?

kavirajk: Also I just noticed we don't have any tests to verify the behavior of `acquire/release` lifecycle. may be worth adding now given the pool impacts that? WDYT?

conn_pool.go
```diff
@@ -0,0 +1,172 @@
...
+
+func (i *idlePool) Get(ctx context.Context) (nativeTransport, error) {
+	i.mu.Lock()
+	defer i.mu.Unlock()
+
+	if i.closed() {
+		return nil, ErrConnectionClosed
+	}
+
+	for {
```
kavirajk, line 67: nit: can you put a small comment saying why we have infinite `for` loop here and explain the exit conditions?

clickhouse.go
```diff
@@ -286,87 +293,46 @@ func DefaultDialStrategy(ctx context.Context, connID int, opt *Options, dial Dia
...
 		}
-		conn.setReleased(false)
-		conn.debugf("[acquired from pool]")
-		return conn, nil
-	default:
+
+		conn.close()
 	}
+
 	if conn, err = ch.dial(ctx); err != nil {
```
kavirajk, line 328: should we check for `ctx.Done()` before doing `ch.dial()` may be? Because I notice the passed in context only checked again all the way during `queryHello` to the server.

clickhouse.go
```diff
@@ -101,10 +104,13 @@ type nativeTransportRelease func(nativeTransport, error)
 
 type clickhouse struct {
 	opt    *Options
-	idle   chan nativeTransport
-	open   chan struct{}
-	exit   chan struct{}
 	connID int64
+
+	idle *idlePool
```
kavirajk: Correct me if I'm wrong here. Trying to understand the lifecycle of a `conn` (nativeTransport) between `open` channel and `idlePool`.

1. So `open` channel is to make sure we have a cap on `max` open connections at any time.
2. When we move the connection from open to idle (Put on idlePool) we also release the slot on `open` channel by reading from it `<- ch.open`
3. When we take a connection from idlePool to open again we acquire a slot on `open` channel by writing to it `ch.open <- struct{}{}`

Can we write a test to lock this behavior may be? Rationale being to avoid any regression in the future if anyone optimize the pool. This is also related to other comment where I wanted tests on `acquire/release` life cycle. WDYT?

conn_pool.go
kavirajk: Also nice to have a test to verify `drainPool` is actually cleaning up the connections as per `MaxConnectionLifeTime`.  To avoid keeping the stale connections in the queue for longer period of time.

internal/circular/queue.go
GeorgeMac: So I chose to go with `DeleteFunc` as the new name. This is to be closer inline with `slices.DeleteFunc`.
My only concern now is that the semantics are slightly different to DeleteFunc and whether that might be confused.
The stdlib implementation returns the slice which the remaining entries, whereas, for our needs, we want the deleted entries, so we can close them.

Alternatively, we could remove the return arg altogether and have the caller call close on the connections in the provided function? That might remove that confusion in the name. Or leave it as is.
WDYT?

conn_pool.go
GeorgeMac: Ahh mate, synctest is so good. I managed to get both these cases here https://github.com/ClickHouse/clickhouse-go/pull/1687/commits/155cee1d0dd22545f482f454e0cf0de941426f4b

Because it makes a bubble and mocks out the `time` package under the hood so you can do things like Sleep and Now and it all moves in lock-step without actually sleeping and returns Now precisely with the increments of when you call Sleep. Plus it does a final wait after Close to ensure the drain routine is not leaked.

conn_pool.go
GeorgeMac: Sorry, coming back to this one, I am little confused exactly what to test.

I guess there are a couple theoretical cases here for a misbehaving circular queue.
Either it:
1. returns an infinite stream of expired connections
2. returns a nil connection and `true`

The first is a serious bug in the queue implementation, which the best defense I think we have is context cancellation. Is this something you want to see tested? rogue pool + Get can be defended via context cancellation?

The second was result in an instance panic because we would attempt to access on a nil pointer.
Which is fine because we call unlock with a defer.
Is this a case you want to see tested? That Get panics on nil from pool and that the lock is unlocked?

internal/circular/queue.go
GeorgeMac: I went with unexporting it for now. Updated the tests to use the unexported.
Then it is there for whatever reason if we want it again. WDYT?

clickhouse.go
GeorgeMac: So, upon reflection, the tricky part about acquire / release is that they actually dial out a connection.
And not only that, they perform a handshake in both the tcp and http cases.

This makes it tricky to isolate a unit test since we have the network involved.
I think there is one argument to be made that the integration tests do exercise this code path a lot with real connections.
So, I would at-least want to avoid spinning up something that heavy here for unit tests.
As we already have that going on in those tests.

Which I think leaves a couple options on the table r.e. further unit testing:

1. abstract dialing away internally

It is a little invasive, but we could do some surgery in `*clickhouse.dial` such that we can swap in our own dial implementation in the test suite. That would allow us to mock out a connection in unit tests and assert our dial function is called. Bit of unexported dependency injection.

2. in-process server-side implementation

I expect no one wants me to bring this back 😂 but in my last contribution I actually wrote a Go implementation of the server protocol for clickhouse-go, which allowed for in-process unit style tests. Which meant you could actually perform a real dial and then implement your own handlers in Go for the handshake and responding to queries.

It was ditched for being too much to maintain, which is fair enough.
https://github.com/GeorgeMac/clickhouse-go/blob/5062e567e9ed470694aca5c36147dbc4d7c3a687/lib/testing/server.go#L39-L55

I suspect approach 1 is the best, but thought I would stop and ask your thoughts first.

clickhouse.go
GeorgeMac: Happy to address this once we get a plan for how to go about mocking or swapping in a dial implementation of sorts 👍 (see comment above).

clickhouse.go
GeorgeMac: Yeah we can check again, or we could check in dial before dial / dialHttp. Let me update that.

internal/circular/queue.go
kavirajk: Here is an idea. Looking at the usage of this api. Only real usage currently is to "remove expired connections".
```
	// Remove only expired connections
	for conn := range i.conns.DeleteFunc(func(conn nativeTransport) bool {
		return i.isExpired(conn)
	}) {
		conn.close()
	}

```

Other places are just tests. Can we just expose an api just to *evict* those expired connections? And call the api just `Evict()` with some doc comments? It doesn't have to accept any arguments neither return any.

Basically I'm think this method doesn't have to extensible at all? WDYT?

clickhouse.go
kavirajk: Good points. I would prefer option (1) given that's more easy to mock. In fact I was thinking to do similar things to test the acquire/release lifecycle :) may be it's a good chance to do it.

This is how I envision (feel free to protest :) )

1. Have an explicit `dial DialerFunc` in `clickhouse` type. 
2. Move the current implementation of `clickhouse.dial` as `defaultDialer`
3. Have an api `openWithDialer` (similar to `Open` api) which we can call on tests to pass in our mock implementation of dialer.
4. Change `Open()` api to use `openWithDialer(... defaultDialer)` to make sure both share exact same code. 

Goal is to avoid any explicit mocks generation code for `dial` (not a big fan, makes the error message messy and hard to debug). We don't have to export `openWithDialer` (at least for now). 

WDYT?

clickhouse.go
kavirajk: Agree. Added my thoughts on previous comment :)

conn_pool.go
kavirajk: Sorry. I should have been more clearer.

Basically what I'm thinking is testing each of the exit points (return statements below) from this loop. So that we know this loop terminates correctly with all possibilities.

```
		if err := ctx.Err(); err != nil {
			// context has been cancelled
			return nil, context.Cause(ctx)
		}

		if i.closed() {
			return nil, ErrConnectionClosed
		}

		// Try to pull a connection
		conn, ok := i.conns.Pull()
		if !ok {
			return nil, nil // queue is empty
		}

		if !i.isExpired(conn) {
			return conn, nil
		}
```

Also can we pass `nil, ErrQueueEmpty` instead of `nil, nil`?

GeorgeMac: Totally, I like that plan 👍

clickhouse.go
GeorgeMac: Great plan 👍

GeorgeMac: I just realized that we can already intercept the dial call in `DialStrategy` by simply overriding this and ignoring the provided `Dial` implementation. So we don't actually need to change anything. Abstractions are all there 👍

GeorgeMac: OK I pushed tests. So, circling back real quick to this original ask r.e. adding a recv on `open` before return.

This implementation worked because we actually always took from the open semaphore up on line 355 above, before we did anything else. So down on the referenced line we would've already taken from the open chan. Effectively returning a slot to the max open pool.

clickhouse.go
GeorgeMac: I made sure with these new tests to move them back over the old implementation too and check that they satisfied the old implementation.

The tests all passed except for the test which ensures that a `release` after a `Close` would lead to the underlying connection being closed. Which is actually a bug in the old implementation. Because the old impl would put the released connection back into the idle pool even after close. So another small leak plugged there and a test around it in place.

conn_pool.go
GeorgeMac: OK I updated this implementation in light of all the tests and added some more comments to the body of the function.

So, in retrospect, the function became a little simpler in https://github.com/ClickHouse/clickhouse-go/pull/1687/commits/89111f65c0b0752866e8e5b6e4bd89346c3b6f50 and therefore, a little easier to test (fewer branches). Because, given we hold a lock for the lifetime of `Get`, we only need to check `i.closed()` after acquiring the lock, as in order for `Close()` to close the pool, it also has to hold the same lock. So that case was no longer necessary.

Now we have full line coverage according to -cover in `conn_pool.go` via `TestConPoll_GetEmpty`, `TestConnPool_GetWithCancelledContext` and `TestConnPool_ExpiredConnectionNotReturned` covering the three branches left in this implementation.

Now I will add a non-nil sentinel error for emptiness.

internal/circular/queue.go
kavirajk: @GeorgeMac WDYT about this comment?

GeorgeMac: Ah so the problem is that the `conns` in this situation is a generic circular queue.
It is not nativeTransport type aware, so it has no concept of expiration or that API at all.
It is the idle conn pool that wraps it that knows/handles expiration.
TL;DR I don't think Evict makes sense on the circular queue level.

PR #1688, opened 2025-10-16
bug: Fixes some correctness bug with ContextWatchdog

kavirajk:
## Summary

1. We don't need infinite for loop to call the callback once context timeout
2. Currently callback will be called multiple times even after context is cancelled. This PR fixes and block the behavior with tests to make sure callback is called only once.
3. Make sure the goroutine started by watchdog exits at all cases without any leaking

## Checklist

### DIFF
--- a/context_watchdog_test.go
+++ b/context_watchdog_test.go
@@ -1,3 +1,5 @@
+//go:build go1.25
+
 package clickhouse
 
 import (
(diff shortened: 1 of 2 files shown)


context_watchdog_test.go
```diff
@@ -0,0 +1,89 @@
+//go:build go1.25
```
kavirajk, line 1: Can we removed once we upgrade the Go tool chain to v1.25.
https://github.com/ClickHouse/clickhouse-go/pull/1689

SpencerTorres (approved):
Glad this has test coverage 👌

PR #1689, opened 2025-10-16
chore: Upgrade Go toolchain to 1.25.x

kavirajk:
## Summary

Main goal is to be able to use `synctest` package to capture some goroutine leaks bugs

It is experemental in v1.24 and becomes part of standard library in v1.25.
https://go.dev/blog/synctest

So that we remove build tags from this these tests.
https://github.com/ClickHouse/clickhouse-go/pull/1688/files#r2435808409

## Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,8 +1,6 @@
 module github.com/ClickHouse/clickhouse-go/v2
 
-go 1.24.0
-
-toolchain go1.24.1
+go 1.25.3
 
 require (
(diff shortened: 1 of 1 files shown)


SpencerTorres (approved):
Curious how this may affect users on 1.24. We can revert if needed, but it should be fine

erka:
@kavirajk Is there any particular reason this is set to 1.25.3 rather than 1.25.0?

PR #1694, opened 2025-11-02
feat: Add support for proper exception handling

kavirajk:
## Summary

Related: #1468 

This PR adds support to extract the exception happens during mid-stream. Supports generic exception marker for both CH 24.11 and before.

## Checklist

### DIFF
--- a/conn_http.go
+++ b/conn_http.go
@@ -440,17 +440,22 @@ func (h *httpConnect) readData(reader *chproto.Reader, timezone *time.Location, 
 		// The decode error typically happens because it tries to read the
 		// "__exception__" marker as binary data
 
+		// Exception block size can be up to 16KiB max.
+		// https://clickhouse.com/docs/interfaces/http#http_response_codes_caveats
+		// NOTE: When exception happens, a dedicated block is allocated for exception and only
+		// that exception will be present in the block. So safe to assume whole block size is same
+		// exception block max size 16KiB.
+		maxSize := int64(16 * 1024)
+
 		// Try to read any remaining data
-		lr := &limitedReader{reader: reader, limit: 100 * 1024}
-		buf := make([]byte, 4096)
-		for {
-			n, readErr := lr.Read(buf)
-			if n > 0 && captureBuffer != nil {
-				captureBuffer.Write(buf[:n])
-			}
-			if readErr != nil {
-				break
-			}
+		lr := &limitedReader{reader: reader, limit: 2 * maxSize} // allocating 2 * maxSize for just in case
+		buf := make([]byte, maxSize)
+		n, readErr := lr.Read(buf)
+		if n > 0 && captureBuffer != nil {
+			captureBuffer.Write(buf[:n])
+		}
+		if readErr != nil {
+			h.debugf("[readData]: decoding block failed when parsing exception block: %w", err)
 		}
 
 		// Check if the captured data contains the exception marker
@@ -460,7 +465,7 @@ func (h *httpConnect) readData(reader *chproto.Reader, timezone *time.Location, 
 		}
 
 		// Not an exception, return the original decode error
-		return nil, fmt.Errorf("block decode: %w", err)
+		return nil, fmt.Errorf("block decode for exception: %w", err)
 	}
 	return &block, nil
 }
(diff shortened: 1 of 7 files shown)


conn_http.go
```diff
@@ -433,12 +433,130 @@ func (h *httpConnect) readData(reader *chproto.Reader, timezone *time.Location)
...
+
+		// Try to read any remaining data
+		lr := &limitedReader{reader: reader, limit: 100 * 1024}
+		buf := make([]byte, 4096)
+		for {
+			n, readErr := lr.Read(buf)
+			if n > 0 && captureBuffer != nil {
+				captureBuffer.Write(buf[:n])
+			}
+			if readErr != nil {
```
chernser: I'd add some debug log here - when we failed, error

conn_http.go
```diff
@@ -433,12 +433,130 @@ func (h *httpConnect) readData(reader *chproto.Reader, timezone *time.Location)
...
+
+	// Try to decode the block
 	if err := block.Decode(reader, h.revision); err != nil {
+		// Decode failed - check if captured data contains exception marker
+		// The decode error typically happens because it tries to read the
+		// "__exception__" marker as binary data
+
+		// Try to read any remaining data
+		lr := &limitedReader{reader: reader, limit: 100 * 1024}
+		buf := make([]byte, 4096)
```
chernser: it would require 25 times to call socket IO. 
I think this can be increased to `10024` safely.

conn_http.go
```diff
@@ -433,12 +433,130 @@ func (h *httpConnect) readData(reader *chproto.Reader, timezone *time.Location)
...
 	}
+
+	// Try to decode the block
 	if err := block.Decode(reader, h.revision); err != nil {
+		// Decode failed - check if captured data contains exception marker
+		// The decode error typically happens because it tries to read the
+		// "__exception__" marker as binary data
+
+		// Try to read any remaining data
+		lr := &limitedReader{reader: reader, limit: 100 * 1024}
```
chernser: This relates to the block size. What if block size is 1Mb and exception is at the end?

conn_http.go
```diff
@@ -433,12 +433,130 @@ func (h *httpConnect) readData(reader *chproto.Reader, timezone *time.Location)
...
+//	\n
+//	<message_length> <TAG>
+//	\r\n
+//	__exception__
+//	\r\n
+func parseExceptionFromBytes(data []byte) error {
+	dataStr := string(data)
+
+	// Verify exception marker exists
+	if !strings.Contains(dataStr, "__exception__") {
```
chernser: this is redundant because next `if` checks it again using proper way.

conn_http.go
```diff
@@ -433,12 +433,130 @@ func (h *httpConnect) readData(reader *chproto.Reader, timezone *time.Location)
...
+
+	// Skip past first __exception__\r\n
+	pos := firstMarker + len("__exception__")
+	// Skip \r\n after first marker
+	if pos+2 < len(dataStr) && dataStr[pos:pos+2] == "\r\n" {
+		pos += 2
+	}
+
+	// Skip the exception tag (16 bytes) + \r\n
+	if pos+16+2 < len(dataStr) {
```
chernser: `16` looks very megical - lets create a constant what ideally should be calculated from marker string (what we actually do a few lines above.

conn_http.go
```diff
@@ -433,12 +433,130 @@ func (h *httpConnect) readData(reader *chproto.Reader, timezone *time.Location)
...
+	}
+
+	// Find the first __exception__ marker
+	firstMarker := strings.Index(dataStr, "__exception__")
+	if firstMarker < 0 {
+		return fmt.Errorf("exception marker not found")
+	}
+
+	// Skip past first __exception__\r\n
+	pos := firstMarker + len("__exception__")
```
chernser: we can save `len("__exception__")` for further use.

conn_http.go
kavirajk: done.

conn_http.go
kavirajk: Good catch. Thanks.

I actually made a small change to it. Given the [max exception block is 16KiB ](https://clickhouse.com/docs/interfaces/http#http_response_codes_caveats)and when exception happens at the mid-stream, CH flushes the existing chunks (blocks) [and start new one just for the exception.](https://github.com/ClickHouse/ClickHouse/blob/63501c580e1dddcd6b28c27f69436bb4c87eb62e/src/Server/HTTP/WriteBufferFromHTTPServerResponse.cpp?plain=1#L312-L315) 

So the exception would be the only thing present in this block. So just allocating 16KiB and reading it once should be good enough. Updated accordingly.

conn_http.go
kavirajk: explained it in my [last comment.](https://github.com/ClickHouse/clickhouse-go/pull/1694#discussion_r2578950002)

conn_http.go
kavirajk: good catch. Removed it.

conn_http.go
kavirajk: Make sense. Moved it to local constants :)

conn_http.go
kavirajk: Make sense. Moved it to local constants as well :)

chernser (approved):

PR #1704, opened 2025-11-06
Update cloud version matrix (25.10)

SpencerTorres:
Update versions

### DIFF
--- a/conn_http.go
+++ b/conn_http.go
@@ -1,4 +1,3 @@
-
 package clickhouse
 
 import (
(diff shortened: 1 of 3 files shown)


kavirajk (approved):
LGTM 👍 looks like some settings are failing in 25.10.

conn_http.go
```diff
@@ -253,6 +252,7 @@ func createHTTPRoundTripper(opt *Options) (http.RoundTripper, error) {
 		IdleConnTimeout:       opt.ConnMaxLifetime,
 		ResponseHeaderTimeout: opt.ReadTimeout,
 		TLSClientConfig:       opt.TLS,
+		DisableCompression:    true,
```
kavirajk, line 255: Given `enable_http_compression` is default in CH since 25.10. I would rather Go driver control the HTTP header (`Accept-Encoding`) fully instead of some magic happening behind Go standard library. That's the rationale of it.

PR #1709, opened 2025-11-11
Handshake is setting DialTimeout instead of context deadline

NamanMahor:
## Summary

If a user provides a context.Context with a timeout, it is ignored [here in conn_handshake.go](https://github.com/ClickHouse/clickhouse-go/blob/a6d5a5f853f52348584a89723938df4f509047e6/conn_handshake.go#L18) even though comment is correct to set `context level deadline override any read deadline` but instead of deadline we are setting Dialtimeout.

Made change to handle context deadline as we are already doing in  [here in conn_ping.go](https://github.com/ClickHouse/clickhouse-go/blob/a6d5a5f853f52348584a89723938df4f509047e6/conn_ping.go#L19)


NamanMahor:
@kavirajk sure will add the test. I have one question which could also be bug. we are overriding the ctx with timeout/deadline here https://github.com/ClickHouse/clickhouse-go/blob/main/clickhouse.go#L304 which is being used by `ch.dial(ctx)` down in the code and later in `ch.dial(ctx)` will call `handshake` with same context so effectively we still using the Dial timeout in handshake even after my PR.

NamanMahor:
@kavirajk sorry about late response totally miss this one. I have added the test.

must108 (changes requested):
Hi @NamanMahor thanks for the contribution and effort. Please address the agent's comments reasonably. Thanks!

PR #1718, opened 2025-11-20
chore(release): Workflow Authentication

juliojimenez:
## Summary
Set up release workflow with workflow-authentication app.

https://github.com/ClickHouse/security/issues/6602

https://clickhouse-inc.slack.com/archives/C02H133R3LK/p1763564403660789

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,7 +25,6 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v5
         with:
-          stable: false
           go-version: 1.25
 
       - name: Prepare release
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,10 +9,18 @@ jobs:
   publish:
     runs-on: ubuntu-latest
     steps:
+      - name: Generate Token
+        id: generate-token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ secrets.WORKFLOW_AUTH_PUBLIC_APP_ID }}
+          private-key: ${{ secrets.WORKFLOW_AUTH_PUBLIC_PRIVATE_KEY }}
+
       - uses: actions/checkout@main
         with:
           ref: main
           fetch-depth: 0 # required to generate contributor list
+          token: ${{ steps.generate-token.outputs.token }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -9,34 +9,41 @@ jobs:
...
       - name: Prepare release
         id: release
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ steps.generate-token.outputs.token }}
         run: |
           go run internal/cmd/release/main.go
 
-      - name: Commit and push hanges
+      - name: Commit And Push Changes
```
kavirajk, line 37: @juliojimenez don't we need `env: GITHUB_TOKEN` on this step (Comit and Push Changes) too?

(asking because that's the step exactly failing) 🤔 
https://github.com/ClickHouse/clickhouse-go/actions/runs/19505353474/job/55829961726#step:5:19

.github/workflows/release.yml
```diff
@@ -9,34 +9,41 @@ jobs:
   publish:
     runs-on: ubuntu-latest
     steps:
+      - name: Generate Token
+        id: generate-token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ secrets.WORKFLOW_AUTH_PUBLIC_APP_ID }}
+          private-key: ${{ secrets.WORKFLOW_AUTH_PUBLIC_PRIVATE_KEY }}
```
kavirajk, line 17: just to clarify. `APP_ID` and `KEY` has different prefix.

`*_PUBLIC_API_ID`
`*_PUBLIC_PRIVATE_KEY`

is that intentional? (shouldn't it be `PUBLIC_KEY`?) 🤔

.github/workflows/release.yml
kavirajk: Never mind. I see a commit pushed by the bot in main already :)

.github/workflows/release.yml
kavirajk: Never mind. I see a commit pushed by the bot in main already :)

kavirajk (approved):

PR #1721, opened 2025-11-24
Fix tuple append row handling to gracefully accept empty structs.

jtmolon:
## Summary

AppendRow fails with conversion error if the value of the tuple field is a nil pointer. This is a problem when inserting from protobuf using optional nested messages, when these are not given a value. 

Example error:
```
clickhouse [AppendRow]: updated_ts clickhouse [AppendRow]: converting *timestamppb.Timestamp to Tuple(seconds Int64, nanos Int32) is unsupported
```

This change allows for graceful handling of empty structs.
Full example here: https://gist.github.com/jtmolon/569f5a37bed17620c4c976baee0eb7ab

## Checklist

### DIFF
--- a/lib/column/tuple.go
+++ b/lib/column/tuple.go
@@ -495,10 +495,25 @@ func (col *Tuple) Append(v any) (nulls []uint8, err error) {
 
 func (col *Tuple) AppendRow(v any) error {
 	// allows support of tuples where map or slice is typed and NOT any. Will fail if tuple isn't consistent
-	value := reflect.ValueOf(v)
-	if value.Kind() == reflect.Pointer {
-		value = value.Elem()
+	var value reflect.Value
+	switch v := v.(type) {
+	case reflect.Value:
+		value = v
+	default:
+		value = reflect.ValueOf(v)
 	}
+
+	// Handle nil pointers gracefully
+	if value.Kind() == reflect.Ptr {
+		if value.IsNil() {
+			// Instead of returning an error, create a zero value
+			// for the underlying type
+			value = reflect.New(value.Type().Elem()).Elem()
+		} else {
+			value = value.Elem()
+		}
+	}
+
 	switch value.Kind() {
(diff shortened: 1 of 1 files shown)


lib/column/tuple.go
```diff
@@ -495,10 +495,25 @@ func (col *Tuple) Append(v any) (nulls []uint8, err error) {
...
+	default:
+		value = reflect.ValueOf(v)
 	}
+
+	// Handle nil pointers gracefully
+	if value.Kind() == reflect.Ptr {
+		if value.IsNil() {
+			// Instead of returning an error, create a zero value
+			// for the underlying type
+			value = reflect.New(value.Type().Elem()).Elem()
```
kavirajk, line 511: I'm not sure if we should allow this behavior by default. 

In your example, 
```
var ts *timestamp.Timestamp

r.Append(ts)
```

It makes sense to panic here instead of silently inserting zero value IMHO. Because the later is hard to debug (make it worse in bigger systems).

In your use-case, I see two posibilities.
1. Easier option is to use non-pointer for `timestamp` with `protbuf` tag, so if it's `omitempty` it automatically creates zero value. 
2. If you want to stick with `pointer` type, then doing `nil` check before calling `append` should solve the problem. No?

I would prefer to keep it more explicit and fail early instead of silent insertion of zero values.

jtmolon: Hi @kavirajk,

Sorry for opening the PR and disappearing, and thanks for your feedback. 

Regarding your suggestions, the first is not doable in our use case because the structs we're using to insert data into Clickhouse are auto-generated with `protoc`, so we don't have much control over them. Also given that they're optional fields, I understand that using pointers is the way to go. 

The second one, though, made me realise that my approach was wrong. It definitely makes a lot more sense to treat the input before calling `append`, and it makes sense that using the zero value or not is responsibility of the caller and not the lib. I'll close the PR.

Thanks!

PR #1722, opened 2025-11-24
fix: restore minimal Go version to 1.24

vincentbernat:
## Summary

In #1689, Go minimal version was bumped to 1.25.3, preventing usage by earlier versions of Go, notably 1.24.x. This is not needed as the test using synctest is protected by a build constraint, so it works just fine.

In this PR, the toolchain is set to 1.25.4, such that most people hacking on this package will use Go 1.25. However, for test workflows, we use GOTOOLCHAIN=local to respect the target Go version. This last bit can be removed with actions/setup-go@v6 as it will set GOTOOLCHAIN=local itself.

We could also not set the toolchain at all in go.mod. It would work too.

The alternative to set GOEXPERIMENT=synctest does not work as the API of this package was modified between 1.24 and 1.25.

## Checklist

### DIFF
--- a/clickhouse.go
+++ b/clickhouse.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"math/rand"
+	"sync"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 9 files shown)


vincentbernat:
The failures seem unrelated?

PR #1727, opened 2025-12-04
bug: Fix `formatTime` escaping

### LINKED ISSUE #1713: Input validation issue on bind method
## Observed

In the [format](https://github.com/ClickHouse/clickhouse-go/blob/c463789f1b7d2d1ccbc533efda0c6d79b483f623/bind.go#L264) function, the cases of `time.Time` and `*time.Time` lack the proper sanitization of the provided value which can result in unexpected behavior of the SQL query executed against the database.

The impact and chance of abuse are low since it requires for the user of the library to allow arbitrary input into the time parameter.  

## Expected behaviour
The expected behavior is for the bind function to sanitize the provided input of all types.

## Code example
Using the following snippet you can recreate the issue.

```go
	// Create a test table
	_, err = db.Exec(`
		CREATE TABLE IF NOT EXISTS test_users (
			id UInt32,
			username String,
			email String,
			created_at DateTime
		) ENGINE = Memory
	`)
	if err != nil {
		log.Fatal("Failed to create table:", err)
	}

	// Dummy data
	_, err = db.Exec(`
		INSERT INTO test_users (id, username, email, created_at) VALUES
		(1, 'alice', 'alice@example.com', now()),
		(2, 'bob', 'bob@example.com', now()),
		(3, 'admin', 'admin@example.com', now())
	`)
	if err != nil {
		log.Fatal("Failed to insert test data:", err)
	}

	// The union content would be user provided input.
	namedMaliciousLoc := time.FixedZone("UTC') UNION ALL SELECT id, username, email FROM test_users --", 0)
	namedMaliciousTime := time.Now().In(namedMaliciousLoc)	
	rows, err := db.Query(
		"SELECT id, username, email FROM test_users WHERE created_at > @timestamp",
		clickhouse.Named("timestamp", namedMaliciousTime),
	)
```

kavirajk:
## Summary

Currently formatTime do not properly escape the timezone info passed to it. Meaning timezone info can contain malicious SQL.

This PR makes sure it's properly escaped

Fixes: #1713 

## Checklist

### DIFF
--- a/bind.go
+++ b/bind.go
@@ -252,10 +252,14 @@ func formatTime(tz *time.Location, scale TimeUnit, value time.Time) (string, err
 		}
 		return fmt.Sprintf("toDateTime64('%s', %d)", value.Format(fmt.Sprintf("2006-01-02 15:04:05.%0*d", int(scale*3), 0)), int(scale*3)), nil
 	}
+
+	// Escape the timezone string (timezone may contain malicious SQL query)
+	escapedTimezone := stringQuoteReplacer.Replace(value.Location().String())
+	// escapedTimezone := value.Location().String()
 	if scale == Seconds {
-		return fmt.Sprintf("toDateTime('%s', '%s')", value.Format("2006-01-02 15:04:05"), value.Location().String()), nil
+		return fmt.Sprintf("toDateTime('%s', '%s')", value.Format("2006-01-02 15:04:05"), escapedTimezone), nil
 	}
-	return fmt.Sprintf("toDateTime64('%s', %d, '%s')", value.Format(fmt.Sprintf("2006-01-02 15:04:05.%0*d", int(scale*3), 0)), int(scale*3), value.Location().String()), nil
+	return fmt.Sprintf("toDateTime64('%s', %d, '%s')", value.Format(fmt.Sprintf("2006-01-02 15:04:05.%0*d", int(scale*3), 0)), int(scale*3), escapedTimezone), nil
 }
 
 var stringQuoteReplacer = strings.NewReplacer(`\`, `\\`, `'`, `\'`)
--- a/bind.go
+++ b/bind.go
@@ -2,6 +2,7 @@ package clickhouse
 
 import (
 	std_driver "database/sql/driver"
+	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


bind.go
```diff
@@ -252,10 +252,14 @@ func formatTime(tz *time.Location, scale TimeUnit, value time.Time) (string, err
 		}
 		return fmt.Sprintf("toDateTime64('%s', %d)", value.Format(fmt.Sprintf("2006-01-02 15:04:05.%0*d", int(scale*3), 0)), int(scale*3)), nil
 	}
+
+	// Escape the timezone string (timezone may contain malicious SQL query)
+	escapedTimezone := stringQuoteReplacer.Replace(value.Location().String())
```
chernser: It may be fixes bug but causes sending invalid timezone to server. 
We could fix it with https://pkg.go.dev/time#LoadLocation but it might have horrible performance. :-(

chernser: We should not accept invalid timezone and inform application owner about malicious attempt. 
the most simple way is to compare length of escaped and unescapted TZ and return error say that TZ contains invalid string and print escaped string.

chernser (changes requested):
need changes. see comments.

bind.go
kavirajk: Good catch.  Not only `time.LoadLocation` is slow, there also a chance, the timezone location we look up for may not present in the local node that is running (that not necessarily meaning it's invalid).

I think comparing the length approach is decent one. But I'm bit worried it's not full fool proof. Technically there is no limit on the timezone string.

Let me think about it.

chernser: @kavirajk the idea is not to limit string but check if we escaped something. 
if `UTC` or `ANY_LONG_TIMEZONE/ON_SOME_FAR_FAR_PLANET` is in - then we have nothing to escape and 
len(original) == len(escaped). 

The main thing here is to catch invalid thing on client side. Most cases it should be fine - but weird names will be blocked and users will create issue if they believe it is a valid timezone.

kavirajk: Make sense. I added the length check and updated the tests accordingly. @chernser

chernser (approved):

PR #1736, opened 2025-12-12
Add context option to append more ClientInfo to the system.query_log

SpencerTorres:
## Summary
ClientInfo is appended per-query, but it can only be set upon connection.

This PR allows users to append to it per-query. The original ClientInfo from the initial connection is unaffected, it's append-only.

Changes:
- Add `WithClientInfo` function for usage with `clickhouse.Context`
- Made `Comment` public on `ClientInfo` so users can append to it in their `clickhouse.Options` config
- Updated TCP and HTTP usage. ClientInfo is merged from connection options and query options.
- Updated unit + integration tests

## Usage

```go
// This data will be appended to the existing ClientInfo from clickhouse.Options
queryCtx := clickhouse.Context(ctx, clickhouse.WithClientInfo(clickhouse.ClientInfo{
	Products: []struct {
		Name    string
		Version string
	}{
		{ Name:    "ctxProduct", Version: "1.2.3" },
	},
	Comment: []string{"ctxComment"},
}))

// The system.query_log table will include the additional ClientInfo from the queryCtx
row := conn.QueryRow(queryCtx, "SELECT 1")
```

## Checklist

### DIFF
--- a/client_info.go
+++ b/client_info.go
@@ -2,10 +2,11 @@ package clickhouse
 
 import (
 	"fmt"
-	"github.com/ClickHouse/clickhouse-go/v2/lib/proto"
 	"runtime"
 	"sort"
 	"strings"
+
+	"github.com/ClickHouse/clickhouse-go/v2/lib/proto"
 )
 
 const ClientName = "clickhouse-go"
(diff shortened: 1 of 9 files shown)


client_info.go
```diff
@@ -23,7 +24,34 @@ type ClientInfo struct {
 		Version string
 	}
 
-	comment []string
+	Comment []string
+}
+
+// Append returns a new copy of the combined ClientInfo structs
+func (a ClientInfo) Append(b ClientInfo) ClientInfo {
```
kavirajk, line 31: Q: why not just append with existing products and comments instead of creating new struct all together? any specific rationale for this?

e.g: 
```
a.Products = append(a.Products, b.Products)
a.Comments = append(a.Comments, b.Comments)
return a
```

Any rationale for keeping the original state (a) immutable?

Also can be bit confusing with semantics on the consumer side. No?

```
opts1 := queryOptions(ctx);

cinfo = opts1.ClientInfo.Append(opts2.ClientInfo)

now we have three versions of `ClientInfo`.

1. cinfo
2. opts1.ClientInfo
3. opts2.ClientInfo

```
if anyone uses opts1.ClientInfo, is the intention is to just use the old one?

kavirajk (approved):
LGTM. one minor question.

client_info.go
SpencerTorres: My fear was that the slices in opts1 would remain the same (assuming it came from `*clickhouse.Options`), so each call would keep appending more data to `opts1`. This would result in a memory leak effectively, or maybe I'm wrong and the slices wouldn't accumulate-- either way I didn't want to risk it and instead took an "copy any append" approach.

See https://github.com/ClickHouse/clickhouse-go/pull/1523 for an example of where context copying was an issue

PR #1744, opened 2025-12-24
feat: client-side load balancing

### LINKED ISSUE #1208: ConnOpenStrategy should trigger on every query and between life connect
### Describe the bug
ConnOpenStrategy should trigger on every query and between life connect，this my expect，but now i can only set the ConnMaxLifetime in a smaller interval，such as 3 seconds，and after 3 seconds，the connection idle，and trigger dialStrategy. i wish clickhouse-go should between  the life connections. by th way,  it should classify the connections by the Addr array. for example, my Addr= [127.0.0.1:29100,127.0.0.2:29100], when ConnOpenStrategy=ConnOpenRoundRobin and 127.0.0.1 and 127.0.0.2 is all good, every time i call connection to exec sql, it should round between 127.0.0.1 and 127.0.0.2. 

maybe in the clickhouse-go inside, 127.0.0.1 and 127.0.0.2 has one more connection instance because of MaxOpenConns, but to the caller, he also can round between 127.0.0.1 and 127.0.0.2. 

on the whole，i wish clickhouse-go can matain  good addrs and good connections pool inside, it shoud clean bad or idle connections  inside periodly. ConnOpenStrategy trigger on every query instead of dia.

### Steps to reproduce
1. se=
2.
3.

### Code example
```go
		Addr: []string{
			"127.0.0.1:29100",
			"127.0.0.2:29100",
		},
		ConnOpenStrategy: clickhouse.ConnOpenRoundRobin,
		Debug:            true,
		Auth: clickhouse.Auth{
			Database: "default",
			Username: "xx",
			Password: "xx",
		},
		ConnMaxLifetime: 3 * time.Second,
	})
	if err != nil {
		t.Log(err)
	}
	t.Log("=====================")
	for i := 0; i < 3; i++ {
		v, err := conn.ServerVersion()
		if err != nil {
			t.Log(err)
		}
		t.Log(v.String())
	}
	t.Log("=====================")
	for i := 0; i < 3; i++ {
		v, err := conn.ServerVersion()
		if err != nil {
			t.Log(err)
		}
		t.Log(v.String())
		t.Log("sleep 5 seconds")
		time.Sleep(5 * time.Second)
	}
```
### Error log

    main_test.go:191: =====================
[clickhouse][conn=1][127.0.0.1:29100][handshake] ->  0.0.0
[clickhouse][conn=1][127.0.0.1:29100][handshake] downgrade client proto
[clickhouse][conn=1][127.0.0.1:29100][handshake] <- ClickHouse (127.0.0.1) server version 22.3.3 revision 54455 (timezone Asia/Shanghai)
    main_test.go:197: ClickHouse (127.0.0.1) server version 22.3.3 revision 54455 (timezone Asia/Shanghai)
    main_test.go:197: ClickHouse (127.0.0.1) server version 22.3.3 revision 54455 (timezone Asia/Shanghai)
    main_test.go:197: ClickHouse (127.0.0.1) server version 22.3.3 revision 54455 (timezone Asia/Shanghai)
    main_test.go:199: =====================
    main_test.go:205: ClickHouse (127.0.0.1) server version 22.3.3 revision 54455 (timezone Asia/Shanghai)
    main_test.go:206: sleep 5 seconds
[clickhouse][conn=2][127.0.0.2:29100][handshake] ->  0.0.0
[clickhouse][conn=2][127.0.0.2:29100][handshake] downgrade client proto
[clickhouse][conn=2][127.0.0.2:29100][handshake] <- ClickHouse (127.0.0.2) server version 22.3.3 revision 54455 (timezone Asia/Shanghai)
    main_test.go:205: ClickHouse (127.0.0.2) server version 22.3.3 revision 54455 (timezone Asia/Shanghai)
    main_test.go:206: sleep 5 seconds
[clickhouse][conn=3][127.0.0.1:29100][handshake] ->  0.0.0
[clickhouse][conn=3][127.0.0.1:29100][handshake] downgrade client proto
[clickhouse][conn=3][127.0.0.1:29100][handshake] <- ClickHouse (127.0.0.1) server version 22.3.3 revision 54455 (timezone Asia/Shanghai)
    main_test.go:205: ClickHouse (127.0.0.1) server version 22.3.3 revision 54455 (timezone Asia/Shanghai)
    main_test.go:206: sleep 5 seconds

tank-500m:
## Summary

This PR fixes #1208 by making the client-side load balancing dial strategies (round_robin/random) reliable under high concurrency.
Strategy-driven dialing is now governed by the idle connection cap (MaxIdleConns) and uses an atomic reservation to prevent concurrent acquires from over-dialing beyond the cap.

## Checklist

### DIFF
--- a/lib/proto/block.go
+++ b/lib/proto/block.go
@@ -291,3 +291,7 @@ func (e *BlockError) Error() string {
 	}
 	return fmt.Sprintf("clickhouse [%s]: %s %s", e.Op, e.ColumnName, e.Err)
 }
+
+func (e *BlockError) Unwrap() error {
+	return e.Err
+}


clickhouse.go
```diff
@@ -296,6 +298,32 @@ func DefaultDialStrategy(ctx context.Context, connID int, opt *Options, dial Dia
 	return r, err
 }
 
+func (ch *clickhouse) shouldDialForStrategy(currentConns int) bool {
```
tank-500m, line 301: using a mutex might be safer.

tank-500m:
@kavirajk 

You’re understanding it correctly.

I first implemented it following the approach suggested here:
https://github.com/ClickHouse/clickhouse-go/issues/1208#issuecomment-1971272989

I’m open to changing this behavior if needed.

However, I’m not sure whether it’s correct to decide pooling eligibility based on whether each connection’s destination address is unique (i.e., treating connections as interchangeable only when they target the same resolved endpoint). It’s hard for me to judge what the “right” rule should be here.

I’d appreciate your thoughts. Thanks.

kavirajk:
> My understanding is that you’re essentially proposing following “scene 1” from this comment (https://github.com/ClickHouse/clickhouse-go/issues/1208#issuecomment-1971272989) as the baseline, and then add some kind of mitigation for addresses that fail—e.g., temporarily excluding a bad address (or backing off from it for N minutes) so don’t end up failing every other request in round-robin.
Is that correct?

That's correct :) The connection pooling and robust failovers are something crucial IMO, if we want to go in this direction.  Although I'm not sure how complicated for the refactoring and implementation. Happy to guide to my knowledge if you want to pick that up.

Also my other thought is, given the complexity of this design, for "production" use cases, better to set up proper proxy (like nginx) for even TCP load balancing. [It's not that hard I did in on this PR for my testing. ](https://github.com/ClickHouse/clickhouse-go/pull/1749). 

> our infrastructure team has advised against using an intermediate Load Balancer for this setup, 

May I know what's the rationale behind not going with proxy approach?. Asking because to add it to my list "use-cases" to have nicer client-side load balancing on go client.

tank-500m:
@kavirajk 
sorry I'm late.

our infra team generally recommends using a Load Balancer (preferably DSR) for download traffic.
However, INSERT workload is upload. If we rely on an LB to do per-request load distribution for this kind of heavy ingest traffic, it can put significant pressure on the LB itself, so it’s not recommended on our side.

And If the LB only spreads connections (rather than distributing individual requests), it still seems hard to avoid hotspots unless the client maintains a sufficiently large number of concurrent connections.

PR #1749, opened 2026-01-06
Feat: Add clickhouse cluster mode setup via docker compose

kavirajk:
We need cluster mode setup for testing (both manual testing and automated). Both TCP(native) and HTTP

Long term goal is to run all the tests both in single node and multi-node cluster.

Lot's of cluster mode configurations are copied [from clickhouse-js client.](https://github.com/ClickHouse/clickhouse-js/blob/main/docker-compose.cluster.yml). But that setup doesn't support TCP load-balancing. I added support for TCP on top of it.

I made sure it works with TCP.

Run the cluster
```
docker compose -f docker-compose.cluster.yml up

```

Check 2 nodes cluster with nginx proxy for both HTTP and TCP
```
$ docker ps --format 'table{{.Names}}\t{{.Ports}}'
NAMES                      PORTS
clickhouse-server-node-1   9009/tcp, 0.0.0.0:9181->9181/tcp, [::]:9181->9181/tcp, 0.0.0.0:8124->8123/tcp, [::]:8124->8123/tcp, 0.0.0.0:9002->9000/tcp, [::]:9002->9000/tcp
clickhouse-server-node-2   9009/tcp, 0.0.0.0:8125->8123/tcp, [::]:8125->8123/tcp, 0.0.0.0:9001->9000/tcp, [::]:9001->9000/tcp, 0.0.0.0:9182->9181/tcp, [::]:9182->9181/tcp
clickhouse-go-nginx        0.0.0.0:8123->8123/tcp, [::]:8123->8123/tcp, 80/tcp, 0.0.0.0:9000->9000/tcp, [::]:9000->9000/tcp
```

Connect to TCP via proxy

```
$ clickhouse-client -q 'SELECT cluster,host_name FROM system.clusters FORMAT pretty'
   ┏━━━━━━━━━━━━━━┳━━━━━━━━━━━━━┓
   ┃ cluster      ┃ host_name   ┃
   ┡━━━━━━━━━━━━━━╇━━━━━━━━━━━━━┩
1. │ test_cluster │ clickhouse1 │
   ├──────────────┼─────────────┤
2. │ test_cluster │ clickhouse2 │
   └──────────────┴─────────────┘
$

```

Connect to HTTP via proxy
```
$ curl "http://localhost:8123" --data-binary "SELECT cluster,host_name FROM system.clusters FORMAT pretty"
   ┏━━━━━━━━━━━━━━┳━━━━━━━━━━━━━┓
   ┃ cluster      ┃ host_name   ┃
   ┡━━━━━━━━━━━━━━╇━━━━━━━━━━━━━┩
1. │ test_cluster │ clickhouse1 │
   ├──────────────┼─────────────┤
2. │ test_cluster │ clickhouse2 │
   └──────────────┴─────────────┘

```

## Checklist


.docker/nginx/http.conf
```diff
@@ -0,0 +1,12 @@
+upstream clickhouse_cluster {
+    server clickhouse1:8123;
+    server clickhouse2:8123;
+}
+
+server {
+    listen 8123;
+    client_max_body_size 100M;
```
chernser, line 8: is this limitation documented? I think it may cause someone investigating "strange" issue if testing big payload. 
I would put some readme and text about it. In general it would be good to have a readme about what this docker compose does and what components are present.

docker-compose.yml
```diff
@@ -4,8 +4,9 @@ services:
       CLICKHOUSE_SKIP_USER_SETUP: 1
     networks:
       - clickhouse
+    hostname: clickhouse
     container_name: clickhouse
-    image: clickhouse/clickhouse-server:25.11
+    image: 'clickhouse/clickhouse-server:${CLICKHOUSE_VERSION-25.11-alpine}'
```
chernser: There is a typo in  variable reference : should be `${CLICKHOUSE_VERSION}`

kavirajk: It's not a typo. It's a way to set the `default` value if the ENV is not set (same as in bash)
https://docs.docker.com/compose/how-tos/environment-variables/variable-interpolation/#interpolation-syntax

docker-compose.yml
```diff
@@ -4,8 +4,9 @@ services:
       CLICKHOUSE_SKIP_USER_SETUP: 1
     networks:
       - clickhouse
+    hostname: clickhouse
     container_name: clickhouse
-    image: clickhouse/clickhouse-server:25.11
+    image: 'clickhouse/clickhouse-server:${CLICKHOUSE_VERSION-25.12-alpine}'
```
chernser, line 9: I was aware about `${VAR:-default} -> value of VAR if set and non-empty, otherwise default` 

May be use it as more accurate?

.docker/nginx/http.conf
kavirajk: If any client try to send request more than configured value, nginix will send the proper error message with the status code (413 request entity too large). Tested it on local by setting 10 bytes `client_max_body_size`

```
$ curl -X POST http://localhost:8123 -d "SELECT 1"
1
$ curl -X POST http://localhost:8123 -d "more than 10 bytes of payload"
<html>
<head><title>413 Request Entity Too Large</title></head>
<body>
<center><h1>413 Request Entity Too Large</h1></center>
<hr><center>nginx/1.23.1</center>
</body>
</html>
$

```

docker-compose.yml
kavirajk: Yea thought about it. But it may be bit "surprise" for the user.

e.g:

I by mistake set CLICKHOUSE_VERSION="" (empty). But when I run `docker-compose up` it always picks the `default` value. I may get confuse why it's not picking the value I set.

Where as with current approach, it will take the "empty" value and try to pull the image -> which will fail -> immediate feedback that user made a mistake of setting "wrong" empty value.

I prefer to fail early and fast :) what do you think? @chernser

chernser (approved):

PR #1756, opened 2026-01-21
feat(slog): add support for structured and leveled logger

### LINKED ISSUE #1698: Logging: Support logging levels
Currently we only have `debug=true` to have unstructured log lines on the stdout. We would like to have 

1. Structured logging
2. Basic log levels (INFO, DEBUG, ERROR)

kavirajk:
## Summary
1. Uses standard lib's log/slog as a main logger.
2. Deprecates the old way of logging via Debug flag and DebugF
3. Provide adapter for legacy way of logging with no-op logging
4. Backward compatible with Debug and DebugF apporach of logging

closes: #1698 

Notes (for the reviewers):
1. There are only few significant changes that deserves full review (which I marked as comments)
2. Most of the file changes are updating the caller of log lines.

FAQ:
1. Why deprecate Debug and DebugF - These are unstructured log lines, no labels, no filterable, no much pattern.
2. Why update all the call-sites? - With proper levels and structured labels improve the logging UX
3. Why use slog? why not new `Logger` interface - slog is from standard library, and has all the helpers and log levels already.
4. Is updating the existing Debug logs breaking changes?. No. it's debug logs.

## Checklist

### DIFF
--- a/clickhouse.go
+++ b/clickhouse.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log/slog"
 	"math/rand"
 	"sync"
 	"sync/atomic"
@@ -94,7 +95,7 @@ type nativeTransport interface {
 	connectedAtTime() time.Time
 	isReleased() bool
 	setReleased(released bool)
-	debugf(format string, v ...any)
+	getLogger() *slog.Logger
 	// freeBuffer is called if Options.FreeBufOnConnRelease is set
 	freeBuffer()
 	close() error
@@ -139,7 +140,7 @@ func (ch *clickhouse) Query(ctx context.Context, query string, args ...any) (row
 	if err != nil {
 		return nil, err
 	}
-	conn.debugf("[query] \"%s\"", query)
+	conn.getLogger().Info("executing query", slog.String("sql", query))
 	return conn.query(ctx, ch.release, query, args...)
 }
 
@@ -151,7 +152,7 @@ func (ch *clickhouse) QueryRow(ctx context.Context, query string, args ...any) d
 		}
 	}
 
-	conn.debugf("[query row] \"%s\"", query)
+	conn.getLogger().Info("executing query row", slog.String("sql", query))
 	return conn.queryRow(ctx, ch.release, query, args...)
 }
 
@@ -160,7 +161,7 @@ func (ch *clickhouse) Exec(ctx context.Context, query string, args ...any) error
 	if err != nil {
 		return err
 	}
-	conn.debugf("[exec] \"%s\"", query)
+	conn.getLogger().Info("executing statement", slog.String("sql", query))
 
 	if asyncOpt := queryOptionsAsync(ctx); asyncOpt.ok {
 		err = conn.asyncInsert(ctx, query, asyncOpt.wait, args...)
@@ -182,7 +183,7 @@ func (ch *clickhouse) PrepareBatch(ctx context.Context, query string, opts ...dr
 	if err != nil {
 		return nil, err
 	}
-	conn.debugf("[prepare batch] \"%s\"", query)
+	conn.getLogger().Info("preparing batch", slog.String("sql", query))
 	batch, err := conn.prepareBatch(ctx, ch.release, ch.acquire, query, getPrepareBatchOptions(opts...))
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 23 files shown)


logger.go
```diff
@@ -0,0 +1,125 @@
+package clickhouse
```
kavirajk, line 1: Main logger implementation.

clickhouse.go
```diff
@@ -94,7 +95,7 @@ type nativeTransport interface {
 	connectedAtTime() time.Time
 	isReleased() bool
 	setReleased(released bool)
-	debugf(format string, v ...any)
+	getLogger() *slog.Logger
```
kavirajk, line 98: This is needed to provide connection contexts to connection polling. Without this, the connection pooling doesn't know what's the connection-id, remote-addr, etc. and logs will be very generic like "acquired connection", "releasing connection"

clickhouse_options.go
```diff
@@ -116,13 +117,36 @@ type Options struct {
...
+	//   logger := slog.New(slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{
+	//       Level: slog.LevelDebug,
+	//   }))
+	//   opts := &clickhouse.Options{
+	//       Logger: logger,
+	//   }
+	//
+	// For backward compatibility, if Debug=true and Debugf is set, those will be used
+	// instead of Logger.
+	Logger *slog.Logger
```
kavirajk, line 148: This is the main UX from the user POV. That accepts logger as separate dependency.

clickhouse_options.go
```diff
@@ -398,3 +422,23 @@ func (o Options) setDefaults() *Options {
 	}
 	return &o
 }
+
+// logger returns the appropriate logger based on the Options configuration.
+// Priority order:
+// 1. If Debug=true and Debugf is set, use legacy Debugf (backward compatibility)
+// 2. If Logger is set, use the provided logger
+// 3. Otherwise, use a noop logger (no logging)
+func (o *Options) logger() *slog.Logger {
```
kavirajk, line 431: Provides "sane" default values for logger. Also take into the legacy Debug flag way of doing logging.

clickhouse.go
```diff
@@ -139,7 +140,7 @@ func (ch *clickhouse) Query(ctx context.Context, query string, args ...any) (row
 	if err != nil {
 		return nil, err
 	}
-	conn.debugf("[query] \"%s\"", query)
+	conn.getLogger().Info("executing query", slog.String("sql", query))
```
chernser: - there is a high security risk when logging SQL statements - they may contain sensitive information (SSN, credit card number etc). Usually this is done on debug/trace level only and by dedicated logger (ex.: `clickhouse.client.sql_logger`) 
- Logging with info level is too annoying in most case and can be quite expensive for customers. SQL in this case can be quite large so it will be a problem. Previously it was on debug level. I'd put it on trace level because client seems do very little with SQL itself .

There is more useful bit of information that can be logged- queryID

clickhouse_std.go
```diff
@@ -374,7 +348,7 @@ func (s *stdBatch) Exec(args []driver.Value) (driver.Result, error) {
 		values = append(values, v)
 	}
 	if err := s.batch.Append(values...); err != nil {
-		s.debugf("[batch][exec] append error: %v", err)
+		s.logger.Error("batch append error", slog.Any("error", err))
```
chernser: is it really required to log error on this level? 
User should be able to log it if needed - anyway error is returned as result.
I'd remove logging here completely or leave it on debug level.

clickhouse.go
kavirajk: Agree. logging query shouldn't be info. I'm putting back to Debug (trace level is not a norm in Go).  

> because client seems do very little with SQL itself .
There is more useful bit of information that can be logged- queryID

Sometimes I would like to see the full SQL (in Debug of course) with all the prepare statement get expanded and format applied. Just the actual query sent to the server. Also unfortunately we don't have access to `queryID` at this level.

clickhouse_std.go
kavirajk: I think removing is better. :+1: thanks

chernser (approved):

clickhouse_std.go
```diff
@@ -158,13 +142,6 @@ func OpenDB(opt *Options) *sql.DB {
 	db.SetMaxIdleConns(o.MaxIdleConns)
 	db.SetMaxOpenConns(o.MaxOpenConns)
 	db.SetConnMaxLifetime(o.ConnMaxLifetime)
-	if o.Debug {
-		if o.Debugf != nil {
-			debugf = o.Debugf
-		} else {
-			debugf = log.New(os.Stdout, "[clickhouse-std][opener] ", 0).Printf
-		}
-	}
```
dossy, line 167: FWIW, I think this breaks the ability to enable debug logging by including `debug=1` as a query parameter in the DSN/connection string.

PR #1758, opened 2026-01-22
fix(time/time64): fix the ScanType of both time and time64

### LINKED ISSUE #1757: Time64 ScanType and ScanRow mismatch
## Observed

While working on adding TIME64 support for CDC pipes, we run the following logic to query a table:
```
rows, rowsErr = ch.Query("select <col1>, <col2>, <col3>, ... from <table>")
row := make([]any, 0, numColumns)
for _, ty:= range colTypes {
    row = append(row, reflect.New(ty.ScanType()).Interface())
}
for rows.Next() {
    if err := rows.Scan(row...); err != nil {
        ...
    }
}
```

this is failing with: `clickhouse [ScanRow]: (t) converting Time64 to *time.Time is unsupported`

I then tried the workaround by scanning it into `*time.Duration`.
```
typeName := ty.DatabaseTypeName()
if dbTypeName == "Time64(6)" {
    row = append(row, new(time.Duration))
}
```
It successfully scanned this time, however the precision is in millisecond instead of the expected microseconds. Not sure if this is a bug or if it's an issue in how it's used. 

## Expected behaviour
- ScanType() should match ScanRow() 
- no loss in precision during scan

## Code example

see above

## Error log

```log
 clickhouse [ScanRow]: (t) converting Time64 to *time.Time is unsupported

```

### Environment
* [x] `clickhouse-go` version: 2.42.0
* [x] Interface: ClickHouse API / `database/sql` compatible driver
* [x] Go version: 1.25.5
* [x] Operating system: Linux
* [x] ClickHouse version: 25.11.2

kavirajk:
## Summary
Time and Time64 is of time.Duration type and return correct `ScanType`.

Should fix: #1757 

## Checklist


lib/column/time64.go
```diff
@@ -50,7 +50,7 @@ func (col *Time64) Type() Type {
 }
 
 func (col *Time64) ScanType() reflect.Type {
-	return scanTypeTime
+	return scanTypeDuration
```
chernser, line 53: In java I've defined it as DateTime value. 
Duration in my mind is always positive (the same as in Java) because negative mean something was undone :-) 
I've ended up with DateTime(UTC) because it preserves all information about DB value: 
- it is always possible to calculate number of days, hours, seconds etc. 
- supports negative range.

chernser (approved):

lib/column/time64.go
kavirajk: Thanks for the thoughts @chernser :)

In Go we have dedicated type `time.Duration` (based on int64, so both positive and negative durations are perfectly valid). Plus

1. Duration is just time without any date components (no month, year, days, timezones, etc) - So easy to handle without worrying about user setting some of those.
2. Also supports all the precision (milli, micro, nano) that are supported by ClickHouse. So it's the most appropriate type in Go.

I discussed this with length on these PRs
1. https://github.com/ClickHouse/clickhouse-go/pull/1669#discussion_r2398218310
2. https://github.com/ClickHouse/ch-go/pull/1097

go.mod
```diff
@@ -5,7 +5,7 @@ go 1.24.1
 toolchain go1.25.4
 
 require (
-	github.com/ClickHouse/ch-go v0.70.0
+	github.com/ClickHouse/ch-go v0.70.1-0.20260128190720-c428daf2b5d9
```
kavirajk: I will update this once ch-go is released with patch
https://github.com/ClickHouse/ch-go/pull/1131

kavirajk: Done.

PR #1759, opened 2026-01-22
fix: Fix connection slot leak when context is cancelled during acquire

hermanschaaf:
Fixes a connection pool leak that happens when a query context is cancelled while acquiring a connection.  When `acquire()` gets a connection slot from the `ch.open` channel but then `ch.idle.Get(ctx)` returns a context cancellation error, we were returning immediately without releasing that slot. After enough cancelled queries, the pool would be exhausted even though connections were available.

This fix makes sure the slot gets released. It also adds a test that reproduces the previous issue and shows that is now fixed.

### DIFF
--- a/conn_http.go
+++ b/conn_http.go
@@ -147,20 +147,22 @@ func dialHttp(ctx context.Context, addr string, num int, opt *Options) (*httpCon
 	// Get base logger and enrich with connection-specific context
 	baseLogger := opt.logger()
 	logger := prepareConnLogger(baseLogger, num, addr, "http")
+	scheme := opt.scheme
+	compression := opt.Compression
 
-	if opt.scheme == "" {
+	if scheme == "" {
 		switch opt.Protocol {
 		case HTTP:
-			opt.scheme = opt.Protocol.String()
+			scheme = opt.Protocol.String()
 			if opt.TLS != nil {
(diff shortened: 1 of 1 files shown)


tests/context_cancel_test.go
```diff
@@ -214,3 +215,68 @@ func ExecuteTestContextCancellation(t *testing.T, conn clickhouse.Conn, query st
...
 	assert.Less(t, queryTime-cancelBackoff, time.Second)
 }
+
+// TestContextCancellationNoConnectionSlotLeak verifies that when contexts are cancelled
+// during connection acquisition, connection slots are properly released back to the pool.
+// This test ensures that cancelled queries don't leak connection slots, which would
+// eventually exhaust the connection pool.
+func TestContextCancellationNoConnectionSlotLeak(t *testing.T) {
+	TestProtocols(t, func(t *testing.T, protocol clickhouse.Protocol) {
+		SkipOnHTTP(t, protocol, "context cancel slot leak test")
```
kavirajk: I wonder, why do we skip on this on HTTP? both TCP and HTTP uses pooling via these `acquire()` api.

hermanschaaf: @kavirajk Good point. Updated 👍

PR #1770, opened 2026-02-10
fix: Panic when scanning nullable JSON using a pointer to `clickhouse.JSON`

erezrokah:
## Summary

Without this fix scanning a `Nullable(JSON)` into `*clickhouse.JSON` panics

## Checklist

### DIFF
--- a/tests/json_test.go
+++ b/tests/json_test.go
@@ -3,7 +3,6 @@ package tests
 import (
 	"context"
 	"encoding/json"
-	"reflect"
 	"testing"
 
 	"github.com/ClickHouse/clickhouse-go/v2"
@@ -540,11 +539,11 @@ func TestJSONNullableObjectViaPointer(t *testing.T) {
 		require.Len(t, rowsJson.ColumnTypes(), 1)
 		require.Equal(t, "Nullable(JSON)", rowsJson.ColumnTypes()[0].DatabaseTypeName())
 
-		rowJson := reflect.New(rowsJson.ColumnTypes()[0].ScanType()).Interface()
-		err = rowsJson.Scan(rowJson)
+		var rowJson *clickhouse.JSON
+		err = rowsJson.Scan(&rowJson)
 		require.NoError(t, err)
 
-		xStr, ok := clickhouse.ExtractJSONPathAs[string](*rowJson.(**clickhouse.JSON), "x")
+		xStr, ok := clickhouse.ExtractJSONPathAs[string](rowJson, "x")
 		require.True(t, ok)
 		require.Equal(t, "test", xStr)
 


tests/json_test.go
```diff
@@ -522,6 +522,49 @@ func TestJSONNullableObjectScan(t *testing.T) {
...
+		ctx := context.Background()
+		rows, err := conn.Query(ctx, `SELECT '{"x": "test"}'::Nullable(JSON)`)
+		require.NoError(t, err)
+
+		require.True(t, rows.Next())
+		columnTypes := rows.ColumnTypes()
+		require.Len(t, columnTypes, 1)
+		require.Equal(t, "Nullable(JSON)", columnTypes[0].DatabaseTypeName())
+
+		var row *clickhouse.JSON
```
kavirajk: I don't think passing the `nil` to scan is right way to scan here.

Its should be either
```
row := &clickhouse.JSON{}
```

or
```
row := new(clickhouse.JSON)
```

Then scan will be happy to copy the value into non-nil address.

Basically what you are trying to do here otherwise is asking scan to copy the value to `nil` address. You shouldn't expect to do behind the scene allocation.

It's similar to following simple program
```
package main

func main() {
	var x *int
	*x = 2
}

```

This would cause panic of-course.

lib/column/json.go
```diff
@@ -270,7 +270,7 @@ func (c *JSON) scanRowObject(dest any, row int) error {
 		return nil
 	case **chcol.JSON:
 		obj := c.rowAsJSON(row)
-		**v = *obj
+		*v = obj
```
kavirajk: We cannot change this. This is needed if we were to scan to double pointer memory. This is standard for all the column types.

Your expectation of not panicing when passing `nil` value is false. Seem my other comment.

kavirajk (commented):
Thanks for the PR @erezrokah. Please see my comments. tldr; this expectation is invalid. 

Can you explain your use case? may be gives better idea what you are actually trying to do.

tests/json_test.go
erezrokah: Thanks for clarifying @kavirajk I've pushed another commit to better explain the issue.

> I don't think passing the `nil` to scan is right way to scan here.

That's what is returned when one uses the column `ScanType()` when scanning the row.

For `string` type the code is a bit different and does `new(string)` before setting the value https://github.com/cloudquery/clickhouse-go/blob/ea494b76e00c19342c2417a4db71c26ef19009cf/lib/column/string.go#L55

tests/json_test.go
```diff
@@ -522,6 +523,63 @@ func TestJSONNullableObjectScan(t *testing.T) {
...
+
+		ctx := context.Background()
+
+		rowsString, err := conn.Query(ctx, `SELECT 'test'::Nullable(String)`)
+		require.NoError(t, err)
+
+		require.True(t, rowsString.Next())
+		require.Len(t, rowsString.ColumnTypes(), 1)
+		require.Equal(t, "Nullable(String)", rowsString.ColumnTypes()[0].DatabaseTypeName())
+		rowString := reflect.New(rowsString.ColumnTypes()[0].ScanType()).Interface()
```
erezrokah: Using `ScanType` for string works

tests/json_test.go
```diff
@@ -522,6 +523,63 @@ func TestJSONNullableObjectScan(t *testing.T) {
...
+		require.NoError(t, rowsString.Err())
+
+		rowsJson, err := conn.Query(ctx, `SELECT '{"x": "test"}'::Nullable(JSON)`)
+		require.NoError(t, err)
+
+		require.True(t, rowsJson.Next())
+		require.Len(t, rowsJson.ColumnTypes(), 1)
+		require.Equal(t, "Nullable(JSON)", rowsJson.ColumnTypes()[0].DatabaseTypeName())
+
+		rowJson := reflect.New(rowsJson.ColumnTypes()[0].ScanType()).Interface()
```
erezrokah: Using `ScanType` for JSON panics

I'd expect both to succeed

tests/json_test.go
erezrokah: > You shouldn't expect to do behind the scene allocation.

I think that's what other types do or I'm missing something?

https://github.com/cloudquery/clickhouse-go/blob/ea494b76e00c19342c2417a4db71c26ef19009cf/lib/column/time.go#L56

kavirajk: you are right @erezrokah. I see other types also handles the double pointer by allocation. 

I made those changes. But unable to push because I guess "maintainerCanModify" is set to false for this PR. Can you enable it? or feel free to change yourself.

erezrokah: Thanks for checking @kavirajk appreciate it. Pushed https://github.com/ClickHouse/clickhouse-go/pull/1770/changes/503b51fe1b9917ea19e8a8335e68c0f71673d5f8 with a fix that's more consistent with other types

erezrokah: Hi @kavirajk I don't have an option to allow edits from maintainers (maybe related to the fork being in the `cloudquery` org) but I've made the changes

tests/json_test.go
```diff
@@ -522,6 +523,49 @@ func TestJSONNullableObjectScan(t *testing.T) {
...
+		ctx := context.Background()
+
+		rowsJson, err := conn.Query(ctx, `SELECT '{"x": "test"}'::Nullable(JSON)`)
+		require.NoError(t, err)
+
+		require.True(t, rowsJson.Next())
+		require.Len(t, rowsJson.ColumnTypes(), 1)
+		require.Equal(t, "Nullable(JSON)", rowsJson.ColumnTypes()[0].DatabaseTypeName())
+
+		rowJson := reflect.New(rowsJson.ColumnTypes()[0].ScanType()).Interface()
```
kavirajk: Can you add normal test case without needing `reflect` (like you did before) just try to load result to pointer type directly.

My rationale is we shouldn't make our tests depend on "reflect" package which can be brittle. 

Happy to merge once the tests the fixed. 

Thanks for the fix @erezrokah :)

erezrokah: Sure, I'll do that. Only used reflection to show the use case

erezrokah: Fixed in https://github.com/ClickHouse/clickhouse-go/pull/1770/changes/6c715233765f901d951a118ad050d28f34fed6ef

PR #1771, opened 2026-02-11
fix: inconsistent results for nil in Json.AppendRow

### LINKED ISSUE #1768: Inserting null interface into Nullable(JSON) cause error
## Observed

* Define a ClickHouse table containing a `Nullable(JSON)` column
* Insert a nil interface value (e.g. zero-initialized `any` variable) into this column

Received error is:
`code: 117, message: Cannot parse JSON object here:`

## Expected behaviour

A null value should be inserted into the table

## Code example

```go

func TestNullableJSON(t *testing.T) {
	ctx := context.Background()

		Auth: clickhouse.Auth{
			Database: "default",
			Username: "default",
		},
	})
	require.NoError(t, err, "open clickhouse")

	_ = conn.Exec(ctx, `DROP TABLE IF EXISTS t`)
	err = conn.Exec(ctx, `
		CREATE TABLE t
		(
			data Nullable(JSON)
		)
		ENGINE = Memory
	`)
	require.NoError(t, err, "create table")

	batch, err := conn.PrepareBatch(ctx, `INSERT INTO t (data) VALUES (?)`)
	require.NoError(t, err, "prepare batch")

	var s any
	require.NoError(t, batch.Append(s))
	require.NoError(t, batch.Send(), "batch send")
}
```

## Error log

```log

```

## Details

Similar issue to #1707 

I'll submit a PR fixing this shortly

### Environment
* [x] Go version: go1.25.5
* [x] Operating system: Linux
* [x] ClickHouse version: 25.10.2.
* [ ] Is it a ClickHouse Cloud? no
* [ ] ClickHouse Server non-default settings, if any: none
* [x] `CREATE TABLE` statements for tables involved: (see above)
* [ ] Sample data for all these tables, use [clickhouse-obfuscator](https://github.com/ClickHouse/ClickHouse/blob/master/programs/obfuscator/Obfuscator.cpp#L42-L80) if necessary (n/a)

Laotree:
## Summary

try to fix #1768 
## Checklist

### DIFF
--- a/tests/issues/1708_test.go
+++ b/tests/issues/1708_test.go
@@ -20,10 +20,10 @@ func Test1708(t *testing.T) {
 	}
 
 	t.Run("query without listener succeeds", func(t *testing.T) {
-		var result uint64
+		var result uint8
 		err := conn.QueryRow(context.Background(), "SELECT 1").Scan(&result)
 		require.NoError(t, err)
-		require.Equal(t, uint64(1), result)
+		require.Equal(t, uint8(1), result)
 	})
 
 	t.Run("query with profile events listener receives events", func(t *testing.T) {
@@ -36,10 +36,10 @@ func Test1708(t *testing.T) {
 			}),
 		)
 
-		var result uint64
+		var result uint8
 		err := conn.QueryRow(ctx, "SELECT 1").Scan(&result)
 		require.NoError(t, err)
-		require.Equal(t, uint64(1), result)
+		require.Equal(t, uint8(1), result)
 		// Profile events may or may not arrive for simple queries depending on timing,
 		// but the query must succeed regardless.
 		_ = received
@@ -52,9 +52,9 @@ func Test1708(t *testing.T) {
 			}),
 		)
 
-		var result uint64
+		var result uint8
 		err := conn.QueryRow(ctx, "SELECT 1").Scan(&result)
 		require.NoError(t, err)
-		require.Equal(t, uint64(1), result)
+		require.Equal(t, uint8(1), result)
 	})
 }


lib/column/json.go
```diff
@@ -411,12 +411,12 @@ func (c *JSON) AppendRow(v any) error {
 		}
 
 		var err error
-		if err = c.appendRowString(v); err == nil {
-			c.serializationVersion = JSONStringSerializationVersion
-			return nil
-		} else if err = c.appendRowObject(v); err == nil {
+		if err = c.appendRowObject(v); err == nil {
```
kavirajk, line 433: @Laotree thanks for the fix. Do you think you can a tests to lock this behavior to avoid any regression later?

Laotree: @kavirajk ok, I did tests locally, I would add testcase in PR later

Laotree: Context: When serializationVersion is unset (default), AppendRow auto-detects the serialization mode. The bug (issue
  #1768) was that the old code tried appendRowString before appendRowObject, so a leading nil would lock in String mode,
   causing subsequent object-type rows to fail.

  Test cases — all expect JSONObjectSerializationVersion with no error:
  #: 1
  Case: nil then struct
  Purpose: Core bug: nil should not lock in String mode and block a subsequent struct
  ────────────────────────────────────────
  #: 2
  Case: struct then nil
  Purpose: Reverse order should also succeed consistently
  ────────────────────────────────────────
  #: 3
  Case: nil then map
  Purpose: Same bug pattern with map type
  ────────────────────────────────────────
  #: 4
  Case: map then nil
  Purpose: Reverse order with map
  ────────────────────────────────────────
  #: 5
  Case: nil then *chcol.JSON
  Purpose: Same pattern with the native JSON type
  ────────────────────────────────────────
  #: 6
  Case: *chcol.JSON then nil
  Purpose: Reverse order with native JSON type
  ────────────────────────────────────────
  #: 7
  Case: nil then *struct
  Purpose: Same pattern with struct pointer
  ────────────────────────────────────────
  #: 8
  Case: nil only
  Purpose: Single nil should still resolve to Object mode
  ────────────────────────────────────────
  #: 9
  Case: nil, nil then struct
  Purpose: Multiple leading nils should not cause issues
  ────────────────────────────────────────
  #: 10
  Case: struct, nil, struct
  Purpose: nil sandwiched between valid objects
  ────────────────────────────────────────
  #: 11
  Case: nil then &nil (var s any; s, &s)
  Purpose: Exact reproduction of issue #1768 comment case 1
  ────────────────────────────────────────
  #: 12
  Case: &nil then nil (var s any; &s, s)
  Purpose: Exact reproduction of issue #1768 comment case 2
  Cases 11 and 12 are the exact scenarios from the issue comments, where insertion order of var s any (nil interface)
  and &s (*interface{} pointing to nil) produced inconsistent results before the fix.

lib/column/json.go
WGH-: This comment should be updated to match new behaviour

A better question is why exactly string is used to be preferred? Is it even safe to change this?

WGH-: There is similar code here, in `Append`. Don't we need to change it as well?

Laotree: Yes, I did the change.

kavirajk:
@Laotree @WGH- thanks for the raising the issue and for this PR.

The more I think about the behavior we expect, it boils down to "what kind of nil values" are acceptable on Go side for `Nullable(JSON)` type on the server side.

I think we have a answer to that, it's `nil` value with type `chcol.JSON` (hence `Nullable(JSON)`). I'm bit worries to allow "any nil value types (like string, any, etc)" as allowed `NULL` types here.

Rationale being, 
1. By definition when look at `Nullable(JSON)` it should be nullable "JSON" type not any other types like string, int, any, etc. It's always good to have some "correct type" associated with `nil` in Go.
2. It also makes it "hard to use JSON api in-correctly". Meaning if any types (like string, int, any) by mistake has nil value (usually bug or logic error).

The following program inserts proper NULL on the server side by using `chcol.JSON` type `nil`.

```
func main() {
	ctx := context.Background()

		Addr: []string{
			"127.0.0.1:9000",
		},
		Debug: true,
		// Protocol: clickhouse.HTTP,
	})

	err = conn.Exec(ctx, `
		CREATE OR REPLACE TABLE t
		(
			data Nullable(JSON)
		)
		ENGINE = Memory
	`)

	if err != nil {
		panic(err)
	}

	batch, err := conn.PrepareBatch(ctx, `INSERT INTO t (data) VALUES (?)`)
	if err != nil {
		panic(err)
	}

	var s *chcol.JSON
	if err := batch.Append(s); err != nil {
		panic(err)
	}
	if err := batch.Send(); err != nil {
		panic(err)
	}
}

```

But at the same time, I see the opportunity to improve the error message here.

```
code: 117, message: Cannot parse JSON object here:
```

is confusing and not acceptable. I think we should make it clear, that `nil` value that is getting passed to the api is not of type `chcol.JSON`. 

Ideally it should say something similar to
```
code: 117, message: Cannot parse JSON object as passed `nil` value is not the correct type JSON
```

what do you think @Laotree @WGH- . May be I'm missing some use cases you folks trying to accomplish.

PR #1781, opened 2026-02-27
Fix #1775: omit absent JSON keys when scanning into map[string]any

### LINKED ISSUE #1775: Scanning JSON into map[string]any adds non-present keys from other rows
## Observed

1. Insert `{"a": "foo"}` and `{"b": "bar"}` as two separate rows into JSON data type
2. Scan them into `map[string]any`
3. Get back `map[string]any{"a":"foo", "b": nil}` and  `map[string]any{"a": nil, "b":"bar"}`

## Expected behaviour

`map[string]any{"a":"foo"}` and  `map[string]any{"b":"bar"}`

Note that running the same query through `clickhouse-client` gives back the correct results:

```
bc513d468f09 :) select * from t;

SELECT *
FROM t

Query id: 5fed6c6e-ef1c-4804-865f-3558883bb941

   ┌─data────────┐
1. │ {"a":"foo"} │
2. │ {"b":"bar"} │
   └─────────────┘

2 rows in set. Elapsed: 0.001 sec.
```

## Code example

```go
package code

func TestClickHouseJSONWTF(t *testing.T) {
	ctx := context.Background()

		Auth: clickhouse.Auth{
			Database: "default",
			Username: "default",
		},
	})

        err = conn.Exec(ctx, `
                CREATE TABLE t
                (
                        data JSON
                )
                ENGINE = Memory
        `)
        require.NoError(t, err, "create table")

        err = conn.Exec(ctx, `
                INSERT INTO t VALUES
                        ('{"a": "foo"}'::JSON),
                        ('{"b": "bar"}'::JSON);
        `)
        require.NoError(t, err, "insert data")

        rows, err := conn.Query(ctx, "SELECT data FROM t")
        require.NoError(t, err)

        for rows.Next() {
                var m map[string]any

                err := rows.Scan(&m)
                require.NoError(t, err)

                t.Logf("%#v", m)

                if _, ok := m["a"]; ok {
                        assert.NotContains(t, m, "b")
                }
                if _, ok := m["b"]; ok {
                        assert.NotContains(t, m, "a")
                }
        }
}
```

## Error log

```log
=== RUN   TestClickHouseJSONWTF
    gssdgs_test.go:345: map[string]interface {}{"a":"foo", "b":interface {}(nil)}
    gssdgs_test.go:348:
                Error Trace:    gssdgs_test.go:348
                Error:          map[string]interface {}{"a":"foo", "b":interface {}(nil)} should not contain "b"
                Test:           TestClickHouseJSONWTF
    gssdgs_test.go:351:
                Error Trace:    gssdgs_test.go:351
                Error:          map[string]interface {}{"a":"foo", "b":interface {}(nil)} should not contain "a"
                Test:           TestClickHouseJSONWTF
    gssdgs_test.go:345: map[string]interface {}{"a":interface {}(nil), "b":"bar"}
    gssdgs_test.go:348:
                Error Trace:    gssdgs_test.go:348
                Error:          map[string]interface {}{"a":interface {}(nil), "b":"bar"} should not contain "b"
                Test:           TestClickHouseJSONWTF
    gssdgs_test.go:351:
                Error Trace:    gssdgs_test.go:351
                Error:          map[string]interface {}{"a":interface {}(nil), "b":"bar"} should not contain "a"
                Test:           TestClickHouseJSONWTF
```

### Environment
* [x] Go version: 1.25.5
* [x] Operating system: Linux
* [x] ClickHouse version: 25.10.2.
* [x] ClickHouse Server non-default settings, if any: none
* [x] `CREATE TABLE` statements for tables involved: (see above)
* [x] Sample data for all these tables: (see above)

varundeepsaini:
## Summary
Fixes #1775.

When scanning ClickHouse `JSON` into `map[string]any`, the driver was adding keys with `nil` values even when those keys were not present in the current row.

This change updates JSON map scan behavior to:
- skip missing leaf keys (do not materialize them as `nil`)
- skip empty nested maps that only come from missing child paths

Added an integration regression test for this issue, including nested JSON cases.

## Changelog note
```
Fix JSON scan into `map[string]any` to omit keys that are absent in the current row, instead of returning them as `nil`.
```

## Checklist

Test were written by Claude

### DIFF
--- a/tests/issues/1775_test.go
+++ b/tests/issues/1775_test.go
@@ -39,6 +39,7 @@ func TestIssue1775_JSONMapScanOmitsAbsentKeys(t *testing.T) {
 	}()
 
 	testCases := []struct {
+		name              string
 		insertTuple       string
 		expected          map[string]any
 		notExpected       []string
@@ -46,22 +47,26 @@ func TestIssue1775_JSONMapScanOmitsAbsentKeys(t *testing.T) {
 		notExpectedNested map[string][]string
 	}{
 		{
+			name:        "simple non-nested",
 			insertTuple: "(1, '{\"a\":\"foo\"}'::JSON)",
 			expected:    map[string]any{"a": "foo"},
 			notExpected: []string{"b", "x"},
 		},
 		{
+			name:        "simple non-nested2",
 			insertTuple: "(2, '{\"b\":\"bar\"}'::JSON)",
 			expected:    map[string]any{"b": "bar"},
 			notExpected: []string{"a", "x"},
 		},
 		{
-			insertTuple: "(2, '{\"b\":\"bar\", \"d\": null}'::JSON)", // explicit null
-			expected:    map[string]any{"b": "bar", "d": nil},
+			name:        "simple non-nested with explicit null",
+			insertTuple: "(3, '{\"b\":\"bar\", \"d\": null}'::JSON)", // explicit null
+			expected:    map[string]any{"b": "bar"},                  // but don't expct key `d`. because clickhouse server doesn't even store the keys with null values
 			notExpected: []string{"a", "x"},
 		},
 		{
-			insertTuple: "(3, '{\"x\":{\"a\":1}}'::JSON)",
+			name:        "simple nested",
+			insertTuple: "(4, '{\"x\":{\"a\":1}}'::JSON)",
 			notExpected: []string{"a", "b"},
 			expectedNested: map[string]map[string]any{
 				"x": {"a": 1},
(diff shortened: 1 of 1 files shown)


tests/issues/1775_test.go
```diff
@@ -0,0 +1,88 @@
...
+	for rows.Next() {
+		var (
+			id   uint8
+			data map[string]any
+		)
+
+		require.NoError(t, rows.Scan(&id, &data))
+
+		switch id {
+		case 1:
```
chernser: Would you please make a dataset so JSON object will be near expected (not expected ) data. 
``` 
    "(1, '{"a":"foo"}'::JSON)", { "a" }, {"b"}    
```

chernser (approved):

PR #1788, opened 2026-03-07
Upgrade from math/rand to math/rand/v2

### LINKED ISSUE #1653: Upgrade to `math/rand/v2`

Onyx2406:
## Summary

- Migrate all `math/rand` usages to `math/rand/v2` across the entire codebase (19 files)
- Fix `ConnOpenRandom` strategy to pick random offset once outside the loop
- Re-enable skipped random failover tests

## Motivation

Go 1.25 broke `rand.Seed` which caused `ConnOpenRandom` failover tests to be skipped (#1652). The `math/rand` package is deprecated in favor of `math/rand/v2` which provides automatic seeding and better APIs.

As discussed in #1653, the `ConnOpenRandom` strategy also had a bug where `rand.Int()` was called on every loop iteration, meaning a connection attempt could pick the same (broken) address multiple times and fail to try all addresses.

## Changes

**Core library (`clickhouse.go`, `clickhouse_std.go`):**
- Replace `math/rand` with `math/rand/v2`
- Fix `ConnOpenRandom`: pick random starting offset once via `rand.IntN(len(addrs))` before the loop, then iterate through all addresses deterministically from that offset. This ensures every address is tried exactly once before failing.

**Tests (11 files):**
- Replace `rand.NewSource`/`rand.New` with `rand.NewPCG`/`rand.New`
- Replace `rand.Intn(n)` with `rand.IntN(n)` (v2 naming)
- Replace `src.Int63()` with `rng.Int64()` in string generators
- Remove `rand.Seed` calls (v2 auto-seeds)
- Re-enable `TestConnFailoverRandom` and `TestStdConnFailoverRandom`

**Examples (5 files):**
- Remove `rand.Seed` calls from multi-host and test setup
- Update `ResetRandSeed()` to no-op (v2 has no `rand.Seed`)
- Remove unused imports (`time`, `math/rand`)

## Test Plan

- [x] `go build ./...` passes cleanly
- [x] No remaining `math/rand` v1 imports (`grep` confirms zero matches)
- [x] `TestConnFailoverRandom` and `TestStdConnFailoverRandom` re-enabled (previously skipped)
- [ ] CI integration tests (require ClickHouse Docker container)

Fixes #1653

### DIFF
--- a/clickhouse.go
+++ b/clickhouse.go
@@ -5,7 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
-	"math/rand"
+	"math/rand/v2"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -283,6 +283,7 @@ func (ch *clickhouse) dial(ctx context.Context) (conn nativeTransport, err error
 }
 
 func DefaultDialStrategy(ctx context.Context, connID int, opt *Options, dial Dial) (r DialResult, err error) {
+	random := rand.IntN(len(opt.Addr))
 	for i := range opt.Addr {
 		var num int
 		switch opt.ConnOpenStrategy {
@@ -291,7 +292,6 @@ func DefaultDialStrategy(ctx context.Context, connID int, opt *Options, dial Dia
 		case ConnOpenRoundRobin:
 			num = (connID + i) % len(opt.Addr)
 		case ConnOpenRandom:
-			random := rand.Int()
 			num = (random + i) % len(opt.Addr)
 		}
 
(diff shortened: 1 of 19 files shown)


must108 (commented):
Closing as this change has been superseded by #1810.

PR #1790, opened 2026-03-08
Intern LowCardinality Row values to reduce string allocations

### LINKED ISSUE #1762: LowCardinality(String) is not as efficient as it could be
When you query a `LowCardinality(String)` column, it is returned in an interned format as two things:

* An index (mapping of integers to unique strings)
* A row list (the indices only)

https://github.com/ClickHouse/clickhouse-go/blob/308e8faec3811f1d201cd276d429bf2b15ec0539/lib/column/lowcardinality.go#L186-L194

This is all fine and great, but when it comes to reading the rows, each string is reallocated fresh, largely because of the the byte buffer to string conversion [here](https://github.com/ClickHouse/ch-go/blob/c428daf2b5d9c9e3bf415a436270902636902796/proto/col_str.go#L110-L114):

```go
// Row returns row with number i.
func (c ColStr) Row(i int) string {
	p := c.Pos[i]
	return string(c.Buf[p.Start:p.End])
}
```

It seems to me that it would be a lot more memory efficient not produce new strings.

Perhaps `index` should not be a `ColStr`, but something that does not make a new string for every call.

I think it would also help with https://github.com/ClickHouse/clickhouse-go/issues/1747.

Onyx2406:
## Summary

Optimize `LowCardinality(String)` scanning by caching `Row()` results per dictionary index, reducing string allocations from O(N rows) to O(K unique values).

Fixes #1762

## The problem

For `LowCardinality(String)` columns, `Row()` creates a new Go string for every row via byte-to-string conversion, even when many rows share the same dictionary value. For a column with 1M rows and 100 unique values, this means 1M string allocations instead of 100.

Benchmark from @bobrik in the issue:
- **Baseline:** 7406ms, 9282 MiB RAM (1M allocations)
- **Optimized:** 429ms, 1341 MiB RAM (K allocations)
- **Speedup: 17x, Memory: 7x reduction**

## Fix (20 insertions, 1 file)

Add a `rowCache map[int]any` to `LowCardinality` that caches `index.Row(idx)` results by dictionary index. On the first access for a given index, the value is created and cached. Subsequent rows with the same index reuse the cached value.

- Cache only used for non-pointer access (`ptr == false`) — the common path
- Cache cleared on `Reset()` (between blocks)
- Lazy initialization — no allocation until first `Row()` call

### Why this works

LowCardinality columns store data as:
- **Dictionary** (index): K unique values
- **Keys**: N row-to-dictionary mappings

Without caching: `Row(i)` → `indexRowNum(i)` → `index.Row(idx)` → new string every time
With caching: `Row(i)` → `indexRowNum(i)` → cache lookup → return cached string

## Test plan

- [x] `go build ./lib/column/...` passes
- [ ] CI (full test suite)
- Existing `lowcardinality_test.go` tests cover scanning behavior


lib/column/lowcardinality.go
```diff
@@ -45,6 +45,11 @@ type LowCardinality struct {
 	keys32 UInt32
 	keys64 UInt64
 
+	// rowCache caches index.Row() results by dictionary index to avoid
+	// redundant string allocations. For LowCardinality(String) with N rows
+	// and K unique values, this reduces allocations from N to K.
+	rowCache map[int]any
```
bobrik: Would it be better to have it as a linear slice that is eagerly populated?

Onyx2406: yes, the dictionary size is known upfront so a []any slice indexed by dictionary position is simpler and faster than a map.

must108 (changes requested):
Hi @Onyx2406, thanks for the contribution. Please reasonably address everything flagged by the agents. Thanks!

PR #1797, opened 2026-03-14
fix: serialize nil query parameters to \N instead of NULL (#1760)

### LINKED ISSUE #1760: Query Parameter Nullable(T) does not accept nils
## Observed

If I have a parameter to a query defined as `{filter_by_market_cap:Nullable(Decimal128(8))}` and I pass in `clickhouse.Named("filter_by_market_cap", nil)`, I get an error that NULL can't be parsed as Decimal.

I get this error:
```
 Cannot parse boolean value here: 'NULL', should be 'true' or 'false' controlled by setting bool_true_representation and bool_false_representation or one of True/False/T/F/Y/N/Yes/No/On/Off/Enable/Disable/Enabled/Disabled/1/0: value NULL cannot be parsed as Bool for query parameter 'allow_etf'
```

The same happens with UUIDs, getting the following error:

```
Cannot parse uuid NULL: value NULL cannot be parsed as Nullable(UUID) for query parameter 'after'
```

Same for other types.

## Expected behaviour

A query that contains `Nullable(T)` input parameters parses correctly interprets `nil`s (NULLs) as the specific Nullable type.

## Code example

```
CREATE TABLE ticker
(
    symbol     String,
    exchange   String,
    currency   String,
    beta       Decimal32(8),
    is_etf     Bool,
    created_at DateTime64(3) DEFAULT now64(3),
    updated_at DateTime64(3) DEFAULT now64(3)
) ENGINE = ReplacingMergeTree(updated_at) PRIMARY KEY (symbol)
      ORDER BY (symbol)
      SETTINGS index_granularity = 8192;

SELECT * FROM ticker WHERE ({is_etf:Nullable(Bool)} IS NULL OR is_etf = {is_etf:Nullable(Bool)});
```

I am initializing the DB connection as such:
```go
func Open(ctx context.Context, dsn string, opts ...DatabaseOption) (Database, error) {
	options, err := clickhouse.ParseDSN(dsn)
	if err != nil {
		return Database{}, err
	}

	for _, opt := range opts {
		opt(options)
	}

	// Enable JSON type support
	options.Settings["allow_experimental_json_type"] = true

	conn, err := clickhouse.Open(options)
	if err != nil {
		return Database{}, err
	}

	ctx, cancel := context.WithTimeoutCause(ctx, ConnectionTimeout, ErrNoConn)
	defer cancel()

	if err = conn.Ping(ctx); err != nil {
		return Database{}, err
	}
	return Database{
		conn: conn,
	}, nil
}
```

```go
	if err := querier.Select(ctx, &result, query, clickhouse.Named("is_etf", nil)); err != nil {
		return nil, err
	}
```

## Error log

```log
 Cannot parse boolean value here: 'NULL', should be 'true' or 'false' controlled by setting bool_true_representation and bool_false_representation or one of True/False/T/F/Y/N/Yes/No/On/Off/Enable/Disable/Enabled/Disabled/1/0: value NULL cannot be parsed as Bool for query parameter 'allow_etf'
```

### Environment
* [x] `clickhouse-go` version: github.com/ClickHouse/clickhouse-go/v2 v2.42.0
* [x] Interface: ClickHouse
* [ ] Go version:
* [ ] Operating system:
* [ ] ClickHouse version:
* [ ] Is it a ClickHouse Cloud? Yes

DobryySoul:
Resolves #1760.

### What changes did you make?
When passing `nil` to a query parameter (e.g. `{is_etf:Nullable(Bool)}`), the driver previously serialized it as the string literal `"NULL"`. Because native Query Parameters in ClickHouse are passed in the `TabSeparated` format over HTTP/TCP, ClickHouse expects `\N` for SQL NULLs instead. Sending `"NULL"` caused parsing errors like `Cannot parse boolean value here: 'NULL'`. 

This PR fixes the formatting inside `bindQueryOrAppendParameters()` so that it correctly serializes `nil` as `\N` for query parameters, while leaving the legacy string replacement behavior intact.

### Checklist:
- [x] Added unit tests for query parameters binding to ensure `nil` translates to `\N` and other types remain unaffected
- [x] Added integration tests for both Native and HTTP protocols (`tests` & `tests/std`) to verify `Nullable` parameters on a real server
- [x] Verified tests pass locally using `make test`

### DIFF
--- a/query_parameters.go
+++ b/query_parameters.go
@@ -39,6 +39,9 @@ func bindQueryOrAppendParameters(paramsProtocolSupport bool, options *QueryOptio
 				if err != nil {
 					return "", err
 				}
+				if strVal == "NULL" {
+					strVal = "\\N"
+				}
 				options.parameters[p.Name] = strVal
 
 			case driver.NamedDateValue:
--- a/query_parameters_test.go
+++ b/query_parameters_test.go
@@ -1,0 +1,148 @@
+package clickhouse
+
+import (
+	"testing"
+	"time"
+
+	"github.com/ClickHouse/clickhouse-go/v2/lib/driver"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestBindQueryOrAppendParameters(t *testing.T) {
+	testTime := time.Date(2023, 1, 1, 12, 0, 0, 0, time.UTC)
+
+	tests := []struct {
+		name          string
+		param         any
+		expectedValue string
+		expectError   bool
+	}{
+		// Nil / NULL case (The fixed bug)
+		{
+			name:          "nil translates to \\N",
+			param:         Named("param", nil),
+			expectedValue: "\\N",
+		},
+		// Basic types
+		{
+			name:          "boolean true",
+			param:         Named("param", true),
+			expectedValue: "1",
+		},
+		{
+			name:          "boolean false",
+			param:         Named("param", false),
+			expectedValue: "0",
+		},
+		{
+			name:          "string direct bypass",
+			param:         Named("param", "hello_world"),
+			expectedValue: "hello_world",
+		},
+		{
+			name:          "string with quotes bypass",
+			param:         Named("param", "hello 'world'"),
+			expectedValue: "hello 'world'", // String bypasses format(), so it shouldn't have extra quotes added
+		},
+		{
+			name:          "integer",
+			param:         Named("param", 42),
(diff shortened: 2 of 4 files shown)


must108 (changes requested):
Hi @DobryySoul thanks for the contribution. Please make sure you address the agent's comments reasonably. Thanks!

PR #1802, opened 2026-03-19
fix: preserve original case for DSN query parameter values

### LINKED ISSUE #1800: DSN parser lowercases setting values unexpectedly
## Observed

  1. String-type settings passed via DSN query parameters are silently lowercased
  2. `strings.ToLower` is applied for `true`/`false` comparison, but the lowered value leaks into
  the `default` branch where the original case should be preserved
  3. Affects any string setting where case matters (e.g., `log_comment`, `query_id`)

  ## Expected behaviour

  Setting values passed via DSN should preserve their original case. Only the boolean comparison
  (`true`/`false`) should be case-insensitive.

  ## Code example

  ```go
  package main

  import (
        "fmt"
  )

  func main() {
        opts, _ := clickhouse.ParseDSN("clickhouse://127.0.0.1/default?log_comment=REQUEST_TYPE:proxy")
        fmt.Println(opts.Settings["log_comment"])
        // actual:   "request_type:proxy"
        // expected: "REQUEST_TYPE:proxy"
  }
```
  ## Error log

  No error is raised — the value is silently lowercased.

  ## Details

  The issue is in clickhouse_options.go, fromDSN method:

  // current code
  ```go
  switch p := strings.ToLower(params.Get(v)); p {
  case "true":
      o.Settings[v] = int(1)
  case "false":
      o.Settings[v] = int(0)
  default:
      if n, err := strconv.Atoi(p); err == nil {   // ← lowercased p
          o.Settings[v] = n
      } else {
          o.Settings[v] = p                         // ← lowercased p
      }
  }
```

### Environment
* [ ] `clickhouse-go` version: latest main
* [ ] Interface: ClickHouse API / `database/sql` compatible driver
* [ ] Go version: 1.24
* [ ] Operating system:
* [ ] ClickHouse version:

tank-500m:
## Summary
Fix https://github.com/ClickHouse/clickhouse-go/issues/1800

`strings.ToLower` was applied to query parameter values for `true`/`false` boolean comparison, but the lowered value leaked into the `default` branch, causing settings like `log_comment` to be unintentionally lowercased.

## Checklist


SpencerTorres (approved):
Nice, I wonder if this has any breaking changes

must108 (approved):
Ship it!

PR #1803, opened 2026-03-23
chore: make PR from forks run cloud tests correctly

kavirajk:
## Summary

This enables any PR from forks to origin (main branch) runs the cloud tests without any problem with secrets/envs needed to run the tests.

Basically to avoid cases like this #1796 PR.

It's also secure as we use `pull_request_target` [trigger](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target) which applies only if PR is raised against the origin (this repo) and maintainers approve the workflow gate.

## Checklist

### DIFF
--- a/.github/workflows/run-tests-cloud.yml
+++ b/.github/workflows/run-tests-cloud.yml
@@ -26,6 +26,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event_name == 'pull_request_target' && github.event.pull_request.head.sha || github.ref }}
 
       - name: Install Go ${{ matrix.go }}
         uses: actions/setup-go@v5
--- a/.github/workflows/run-tests-cloud.yml
+++ b/.github/workflows/run-tests-cloud.yml
@@ -14,7 +14,7 @@ on:
 
 jobs:
   test:
-    if: github.event_name == 'push' || github.event_name == 'pull_request'
+    if: github.event_name == 'push' || github.event_name == 'pull_request_target'
     runs-on: [self-hosted, style-checker-aarch64]
     defaults:
       run:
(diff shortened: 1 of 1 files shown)


.github/workflows/run-tests-cloud.yml
koletzilla, line 28: We need to checkout to the PR's commit as `pull_request_target` will checkout to main: https://github.com/ClickHouse/spark-clickhouse-connector/blob/a1d8b7b32cae27e2fe38133c10ce10613b824013/.github/workflows/cloud.yml#L62-L63

.github/workflows/run-tests-cloud.yml
```diff
@@ -7,7 +7,9 @@ on:
     branches:
       - main
   workflow_dispatch:
-  pull_request:
+  pull_request_target:
```
koletzilla, line 10: This will cause the flow to execute automatically so any bad actor can add bad code that may impact our secrets. We need to add some logic, like the label one, to make sure it doesn't run by default but after we "approved" it https://github.com/ClickHouse/spark-clickhouse-connector/blob/a1d8b7b32cae27e2fe38133c10ce10613b824013/.github/workflows/cloud.yml

koletzilla (changes requested):

.github/workflows/run-tests-cloud.yml
kavirajk: good catch. Thanks.

updated.

.github/workflows/run-tests-cloud.yml
kavirajk: Good point. We already have PR gate which runs these workflows after maintainer approving this workflow (different from PR approve). Without this no external actors can run any workflows on their PRs (from forks)

This is something configured on this repo already. e.g

screenshot4

koletzilla: awesome!

koletzilla (approved):

PR #1804, opened 2026-03-23
Fixed handler leak.

Aleksei-Badyaev:
Handle 'pipeReader' is created at conn_http_batch.go:249 by calling function 'io.Pipe' and lost at conn_http_batch.go:274.


must108 (changes requested):
Hi @Aleksei-Badyaev thanks for the contribution. Please just ensure to address the agent's comments. Thanks!

PR #1806, opened 2026-03-24
chore: make cloud test runs on-demand only with labeling

kavirajk:
## Summary

Follow up to #1803 

This is needed for two reasons
1. `pull_request_target` cannot be included in the workflow approve gate.
2. without approval gate, we cannot avoid bad actors to trigger cloud run just by creating PR to origin.
3. I changed the trigger to `labeled` only because this can only run when maintainer explicitly label it.
4. After success/fail run, it automatically remove the label thus no additional commit would trigger it automatically unless maintainer lable it again

## Checklist

### DIFF
--- a/.github/workflows/run-tests-cloud.yml
+++ b/.github/workflows/run-tests-cloud.yml
@@ -10,11 +10,11 @@ on:
   pull_request_target:
     branches:
       - main
-    types: [opened, labeled, synchronize, reopened]
+    types: [labeled]
 
 jobs:
   test:
-    if: github.event_name == 'push' || github.event_name == 'pull_request_target'
+    if: github.event_name == 'push' || github.event_name == 'workflow_dispatch' || (github.event_name == 'pull_request_target' && github.event.label.name == 'tests:run-cloud')
     runs-on: [self-hosted, style-checker-aarch64]
     defaults:
       run:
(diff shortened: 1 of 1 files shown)


.github/workflows/run-tests-cloud.yml
```diff
@@ -10,11 +10,11 @@ on:
   pull_request_target:
     branches:
       - main
-    types: [opened, labeled, synchronize, reopened]
+    types: [labeled]
```
kavirajk, line 13: NOTE: I made it only `labeled` as source of this trigger is only maintainer labeling it explicitly.

I could have kept it like before because in the below check we make sure we run only if labeled with particular label. But then every time it just start the work-flow and skip.

.github/workflows/run-tests-cloud.yml
```diff
@@ -45,3 +45,10 @@ jobs:
           GOTOOLCHAIN: "local"
         run: |
           CLICKHOUSE_DIAL_TIMEOUT=20 CLICKHOUSE_TEST_TIMEOUT=600s CLICKHOUSE_QUORUM_INSERT=2 make test
+
+      - name: Remove tests:run-cloud label
```
kavirajk, line 49: This is a safe measure. Say remove the label after success/fail run of the cloud tests and only be trigger if re-labled again.

koletzilla (approved):

PR #1811, opened 2026-03-30
workflows: integrate claude code Github action for code-review

kavirajk:
## Summary

Changes
1. Use `pull_request_targe`t to trigger claude agent review based on
labelling.
2. Fixes remarks on cloud skills and instructions about tests, docs and errors

## Checklist

### DIFF
--- a/.claude/skills/review-pr/SKILL.md
+++ b/.claude/skills/review-pr/SKILL.md
@@ -1,3 +1,10 @@
+---
+name: review-pr
+description: Review a pull request and post structured feedback as a comment.
+argument-hint: "<PR-number>"
+allowed-tools: Read, Glob, Bash(grep:*), Bash(gh pr view:*), Bash(gh pr diff:*), Bash(gh pr comment:*)
+---
+
 Review the pull request and provide structured feedback.
 
 ## How to fetch the PR
--- a/.claude/skills/review-pr/SKILL.md
+++ b/.claude/skills/review-pr/SKILL.md
@@ -52,7 +52,9 @@ Work through each category before writing the review.
 - [ ] New ClickHouse type support: column implementation + round-trip test + example?
 - [ ] Do test failure messages say what was wrong, what input triggered it, and what was expected vs. got?
 - [ ] Are tests table-driven where there are multiple cases?
-- [ ] No test mocking of `driver.Conn` or `driver.Rows` — use a real ClickHouse via testcontainers.
+- [ ] No test mocking of `driver.Conn` or `driver.Rows` for integrations tests inside root `/tests/` — use a real ClickHouse via testcontainers.
+- [ ] Make sure tests are covered for both TCP and HTTP protocol cases.
+- [ ] Make sure tests are covered for both `clickhouse_native` (Open() api returns) api and `std` api (OpenDB() api returns).
 
 ### Performance & protocol
 - [ ] Does the change avoid unnecessary allocations in hot paths (encoding/decoding columns)?
@@ -62,3 +64,7 @@ Work through each category before writing the review.
 ### Documentation
 - [ ] Are new exported symbols documented with a full-sentence doc comment beginning with the symbol name?
 - [ ] ClickHouse SQL types and function names wrapped in backticks in any prose.
+- [ ] Make sure existing docs and examples are updated.
+- [ ] Make sure new docs and examples are added if needed for new features or bug fixes.
+- [ ] Make sure the docs are covered for both TCP and HTTP protocol cases.
+- [ ] Make sure the docs are covered for both `clickhouse_native` (Open() api returns) api and `std` api (OpenDB() api returns).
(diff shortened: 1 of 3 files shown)


.github/workflows/claude-code-review.yml
```diff
@@ -0,0 +1,48 @@
...
+        uses: actions/checkout@v4
+        # SECURITY: pull_request_target checks out the base branch by default.
+        # Do NOT add `ref: ${{ github.event.pull_request.head.sha }}` here —
+        # that would run untrusted fork code with access to repository secrets.
+        # that's why labeling check (usually by maintainer)
+
+      - name: Review PR with Claude
+        uses: anthropics/claude-code-action@v1
+        with:
+          anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
```
kavirajk, line 28: this secret is already added to the repo level secrets

.github/workflows/claude-code-review.yml
```diff
@@ -0,0 +1,48 @@
...
+          prompt: |
```
kavirajk, line 31: Not sure if this prompt is required. Took from [the examples.](https://github.com/anthropics/claude-code-action/blob/main/docs/usage.md#automation-workflows) Just want to make it work first after merging to main. We can polish it once the integration is working fine.

.github/workflows/claude-code-review.yml
```diff
@@ -0,0 +1,48 @@
...
+            Review pull request #${{ github.event.pull_request.number }} using the checklist in `.claude/skills/review-pr/SKILL.md`.
+
+            Steps:
+            1. Read `.claude/skills/review-pr/SKILL.md` to load the review checklist and output format.
+            2. Fetch the PR description with `gh pr view ${{ github.event.pull_request.number }}`.
+            3. Fetch the diff with `gh pr diff ${{ github.event.pull_request.number }}`.
+            4. Work through every checklist category in SKILL.md: Correctness, API design, Go idioms, Tests, Performance & protocol, Documentation.
+            5. Post a single review comment on the PR using `gh pr comment ${{ github.event.pull_request.number }} --body "..."` with the sections defined in SKILL.md: Summary, Must fix, Should fix, Nits, Verdict.
```
alex-clickhouse, line 39: Will `gh` work without any token here? Does it use something from the anthropic side?

.github/workflows/claude-code-review.yml
```diff
@@ -0,0 +1,48 @@
...
+    steps:
+      - name: Checkout base branch
+        uses: actions/checkout@v4
+        # SECURITY: pull_request_target checks out the base branch by default.
+        # Do NOT add `ref: ${{ github.event.pull_request.head.sha }}` here —
+        # that would run untrusted fork code with access to repository secrets.
+        # that's why labeling check (usually by maintainer)
+
+      - name: Review PR with Claude
+        uses: anthropics/claude-code-action@v1
```
alex-clickhouse, line 26: I think this will need --allowedTools to let it use what it needs for the review and comment posting, but maybe they could go in the SKILL.md instead. Something like:

```
claude_args: |
            --allowedTools "Read,Bash(grep:*),Bash(gh pr comment:*),Bash(gh pr diff:*)"
```

.github/workflows/claude-code-review.yml
kavirajk: I think if it can checkout just fine should be able to read it as well?

.claude/skills/review-pr/SKILL.md
alex-clickhouse, line 8: Should've caught this in the first PR, but the SKILL.md file should have a preamble with the allowed tools etc. Something along the lines of:


```
---
name: review-pr
description: Review a pull request. Use when the user wants to review a PR or diff.
argument-hint: "[PR-number or branch-name or diff-spec]"
allowed-tools: Task, Read, Glob, Grep, WebFetch <more?>
---
```

mshustov (approved):

.github/workflows/claude-code-review.yml
kavirajk: good point. do you think it's more robust to put these in say something settings.local.json inside `.claude` directory? instead of in workflow files?

alex-clickhouse: 1. I'm not sure if .claude/settings.json will apply to the github action. (I see there's a [settings parameter that takes json though](https://code.claude.com/docs/en/github-actions#breaking-changes-reference))
2. There's a case for making the github action permissions more restrictive than the ones used for typical development, eg you probably don't want it to run tests on its own for the PR review on github, since they already run in the CI.

kavirajk: added

.claude/skills/review-pr/SKILL.md
kavirajk: added

PR #1826, opened 2026-04-08
fix: set req.Host for Host header in HTTP transport

### LINKED ISSUE #1825: HTTP transport: Host header set via HttpHeaders is silently ignored by Go's net/http
## Description

When using the HTTP transport, setting a `Host` header via `HttpHeaders` in connection options (or DSN) has no effect on the actual HTTP request. The header is silently dropped.

This is because Go's `net/http` ignores `req.Header.Set("Host", ...)` — the only way to set the wire `Host` header is via `req.Host`. The driver currently sets all custom headers uniformly via `req.Header.Set()` at [`conn_http.go:139-141`](https://github.com/ClickHouse/clickhouse-go/blob/v2.43.0/conn_http.go#L139-L141):

```go
for key, value := range h.headers {
    req.Header.Set(key, value)
}
```

From the [Go `net/http` documentation](https://pkg.go.dev/net/http#Request):

> For incoming requests, the Host header is promoted to the Request.Host field and removed from the Header map.

## Impact

This prevents routing through HTTP proxies (e.g. Envoy) that rely on the `Host` header for routing decisions. The `Host` header appears to be set from the caller's perspective, but never reaches the proxy on the wire.

## Suggested Fix

Special-case the `Host` key when iterating custom headers — set `req.Host` instead of `req.Header.Set()`:

```go
for key, value := range h.headers {
    if strings.EqualFold(key, "Host") {
        req.Host = value
    } else {
        req.Header.Set(key, value)
    }
}
```

## Environment

- clickhouse-go version: v2.43.0 (and all prior versions with `HttpHeaders` support)
- Go version: 1.22+ (behavior exists in all Go versions)

binger-li-dd:
## Summary

Go's `net/http` silently ignores `req.Header.Set("Host", ...)` — the wire `Host` header can only be set via `req.Host` ([Go docs](https://pkg.go.dev/net/http#Request)). The driver currently sets all custom headers uniformly via `req.Header.Set()` in `conn_http.go`, so a `Host` header specified in `HttpHeaders` has no effect on the outgoing request.

This one-line fix special-cases the `Host` key (case-insensitive) to set `req.Host` instead, ensuring the header reaches the wire.

## Problem

```go
    Addr:     []string{"localhost:8123"},
    Protocol: clickhouse.HTTP,
    HTTPHeaders: map[string]string{
        "Host": "my-service.example.com",
    },
})
```

The `Host` header is silently dropped. This breaks routing through HTTP proxies (e.g. Envoy) that rely on the `Host` header for routing decisions.

## Fix

```go
for k, v := range opt.HttpHeaders {
    if strings.EqualFold(k, "Host") {
        req.Host = v
    } else {
        req.Header.Set(k, v)
    }
}
```

## Test plan

- Added unit tests in `conn_http_test.go` covering:
  - `Host` header sets `req.Host`
  - Case-insensitive matching (`host`, `HOST`)
  - `Host` header coexists with other custom headers
  - Non-Host headers are unaffected

Fixes #1825

### DIFF
--- a/conn_http.go
+++ b/conn_http.go
@@ -137,7 +137,11 @@ func applyOptionsToRequest(ctx context.Context, req *http.Request, opt *Options)
 	req.Header.Set("User-Agent", opt.ClientInfo.Append(queryOpt.clientInfo).String())
 
 	for k, v := range opt.HttpHeaders {
-		req.Header.Set(k, v)
+		if strings.EqualFold(k, "Host") {
+			req.Host = v
+		} else {
+			req.Header.Set(k, v)
+		}
 	}
 
 	return nil
--- a/conn_http_test.go
+++ b/conn_http_test.go
@@ -1,6 +1,7 @@
 package clickhouse
 
 import (
+	"context"
 	"net/http"
 	"testing"
 )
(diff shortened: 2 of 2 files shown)


SpencerTorres (changes requested):
Can we assume the `Host` header will always be exactly `"Host"`? if so we can change this to a direct key lookup outside the loop. Feels like a better use for the existing map structure.

Other than that, this PR seems like a good change, as long as there's no potential breaking changes or security concerns with proxying as you said. Thanks!

binger-li-dd:
> Can we assume the `Host` header will always be exactly `"Host"`? if so we can change this to a direct key lookup outside the loop. Feels like a better use for the existing map structure.
> 
> Other than that, this PR seems like a good change, as long as there's no potential breaking changes or security concerns with proxying as you said. Thanks!

Yeah that makes sense -- `Host` is usually the convention and it makes sense to assume the header will be exactly it. I just made a change to adopt this suggestion. Thank you!

kavirajk:
@binger-li-dd @SpencerTorres I actually would prefer to have case in-sensitive "host".  To avoid any confusion and better UX IMO. Just someone passed "host" and not having the expected behavior is bad experience even to debug. May I know if I'm missing any rationale for keeping it exactly "Host"?

Also with latest change, I think the header "Host" can be duplicated in both `req.Host` as well as in `req.Headers`.

binger-li-dd:
> @binger-li-dd @SpencerTorres I actually would prefer to have case in-sensitive "host". To avoid any confusion and better UX IMO. Just someone passed "host" and not having the expected behavior is bad experience even to debug. May I know if I'm missing any rationale for keeping it exactly "Host"?
> 
> Also with latest change, I think the header "Host" can be duplicated in both `req.Host` as well as in `req.Headers`.

Fair point — restoring case-insensitive matching with strings.EqualFold. Someone passing "host" and not getting the expected behavior would be frustrating to debug.

On the duplication: the latest change have addressed the duplication as well. Now only `req.Host` contains the host information and `req.Headers` won't have that.

kavirajk (approved):

SpencerTorres (approved):

Issue #1801, opened 2026-03-19
Panic on inserting any[] on a batch with LowCardinality Command


JulesPoulain-SloClap:
Hi!
I have a very strange problem with the Clickhouse Go Library. I'm trying to push an any[] slice to a batch (with the ... operator). But sometime, arround line 300 everytime (but not at the same line/data), I've a nil map panic from lowcardinality.go:148, with col.append.index becoming nil.

Do you already have seen a problem like this?

## Expected behaviour
Don't cause a panic on append

## Code example
```go
batchChannel := make(chan []any)

// Do something to insert the values into the channel.
// The value have correct types and order but are inserted as any

batch, err := ctx.DBConn.PrepareBatch(context.Background(), "INSERT INTO "+ctx.TableName)
if err != nil {
	return itemCount, err
}

go func() {
	go func() {
		for value := range batchChannel {
			if err := batch.Append(value...); err != nil { // Panic Here
				log.Printf("batch append error: %v", err)
			}
		}
	}()
}()
```

## Error log

```log
panic: assignment to entry in nil map

goroutine 32107 [running]:
github.com/ClickHouse/clickhouse-go/v2/lib/column.(*LowCardinality).AppendRow(0xc0010557a0, {0x810f00, 0xc00491dc08})
        C:/Users/xxxx/go/pkg/mod/github.com/!click!house/clickhouse-go/v2@v2.43.0/lib/column/lowcardinality.go:148 +0x367
github.com/ClickHouse/clickhouse-go/v2/lib/proto.(*Block).Append(0xc0021bdf40, {0xc004e07908?, 0x9f, 0xc001076000?})
        C:/Users/xxxx/go/pkg/mod/github.com/!click!house/clickhouse-go/v2@v2.43.0/lib/proto/block.go:49 +0x186
github.com/ClickHouse/clickhouse-go/v2.(*batch).Append(0xc0021c59e0, {0xc004e07908?, 0x0?, 0x0?})
        C:/Users/xxxx/go/pkg/mod/github.com/!click!house/clickhouse-go/v2@v2.43.0/conn_batch.go:127 +0x65
```

## Details

Create Table:
```SQL
CREATE TABLE table_name
(
    build_config LowCardinality(String),
    changelist LowCardinality(String),
    environement LowCardinality(String),
    allocation_id String,
    game_mode LowCardinality(String),
    players_count UInt8,
    spectators_count UInt8,
    hardware_os_version LowCardinality(String),
[...]

PR #1830, opened 2026-04-14
fix: lowcardinality nil map panic

Laotree:
## Summary

fix: prevent nil map panic in LowCardinality.AppendRow (#1801)
When inserting data into a batch with LowCardinality columns, a panic 
"assignment to entry in nil map" occurs around line 300. This happens 
because:
1. LowCardinality column's append.index is not initialized when decoded 
   from server response
2. After Encode (during Send/Flush), append.index is set to nil
3. Subsequent AppendRow calls try to write to a nil map
The fix adds a nil check at the start of AppendRow to lazily initialize 
the map, matching the defensive pattern already used in Reset() and parse().

## Checklist


kavirajk (approved):

kavirajk:
thanks @Laotree for the PR :heart:

Issue #1831, opened 2026-04-14
Bug: Underlying TCP connection leak when using HTTP protocol


kavirajk:
I heard from internal team that when closing HTTP connection, the underlying TCP connection may not get closed consistently.

Did some investigation and creating this issue to track it.

This behavior is even when `client.Close()` is explicitly called and noticed underlying TCP connections are not getting closed.

Here is my theory

#### Problem 1 - Race condition
When `client.Close()` is explicitly invoked, [according to this code,](https://github.com/ClickHouse/clickhouse-go/blob/1422cc5ec4aa145f80908f8f8374c380c90c0545/clickhouse.go?plain=1#L410-L415) we do two things 
```go
     ch.closeOnce.Do(func() {
		err = ch.idle.Close()
		ch.closed.Store(true)
	})
```
(a). drain the existing idle connections from pool. 
(b) mark the connection pool is closed. Note these two are not atomic operation. Some other actors may try to put the connection back in the pool meanwhile (say "after" draining the idle pool and "before" connection pool is marked as closed). 

You can confirm this behavior by [looking at connection pool's release code](https://github.com/ClickHouse/clickhouse-go/blob/1422cc5ec4aa145f80908f8f8374c380c90c0545/clickhouse.go?plain=1#L412-L415)
```go
if ch.closed.Load() {
		conn.close()
		return
	}

	ch.idle.Put(conn)
```
So there a race condition. The connection that was still in-flight and tried to put it back to pool may not get drained (so not closed correctly)

#### Problem 2 - Connection pool's `release` doesn't close the connection in some case.
Even without the race condition, The `idle.Put()` api just [return without closing the connection](https://github.com/ClickHouse/clickhouse-go/blob/1422cc5ec4aa145f80908f8f8374c380c90c0545/conn_pool.go?plain=1#L96-L98) when `connectionPool.closed()` is set
```go
	if i.closed() {
		return
	}
```
We should close that connection at any cost to avoid such connection leak.

#### Proposals
1. Fix the order in `Close()`. First mark the idle pool as closed then drain it. So that no more connection cannot be added while draining.
2. `Put()` api should close the connection even when pool is closed.

must108:
Addressed by #1833

PR #1833, opened 2026-04-17
fix: close connections leaked during pool shutdown (#1831)

Laotree:
Two bugs allowed TCP connections to escape cleanup when Close() was called:

1. clickhouse.Close() drained the idle pool before marking it closed, leaving a window where release() could Put() connections back. Fixed by storing closed=true first, then draining.

2. connPool.Put() silently dropped connections when the pool was already closed instead of closing them, leaking the underlying TCP connection. Fixed by calling conn.close() before returning.

## Checklist


kavirajk:
addresses this issue https://github.com/ClickHouse/clickhouse-go/issues/1831

Laotree:
@must108 Could you review this one?

Fixes two TCP connection leaks during pool shutdown (#1831):

- **`clickhouse.Close()` ordering** — `ch.closed` is now stored *before* draining the idle pool, so a concurrent `release()` sees the closed flag and shuts the connection down itself instead of returning it to an already-drained pool.
- **`connPool.Put()`** — connections handed to a closed pool are now closed rather than silently dropped, which was the actual leak.

Coverage:
- `TestConnPool_PutCloseRace` (unit, passes with `-race`) drives 50 concurrent `Put` vs `Close` and asserts every connection ends up closed.
- `tests/issues/1831_test.go` (integration) races 20 workers against `Close()` on a real server.
- Both fixes complement each other with no double-close path (`release()` is exclusive with `Put`, and `connect.close()` is idempotent); lock ordering is unchanged (close under the pool mutex was already the norm in `drainPool`).

Build, `go vet`, and the `-race` conn-pool tests pass. Branch is `MERGEABLE` with no conflicts.

must108 (changes requested):
@Laotree LGTM, just fix the nits identified by agent and I will approve. Thanks!

must108 (approved):
LGTM! Ship it

PR #1837, opened 2026-04-20
docs: clarify HTTP compression modes

### LINKED ISSUE #1766: Doc: Clarify HTTP web compression vs HTTP block compression
Since CH 25.11, the `enable_http_compression` setting is default in ClickHouse,

Which leads to two different types of compression when using HTTP.

1. HTTP web compression - based on HTTP headers (whole response)
2. ClickHouse block compression

(1) is what general HTTP client understand (e.g: curl)
(2) ibetter compression at the block level.

The problem is sometimes user end up using both. But (2) needs custom code to decompress (via `compress=1`) query param.

The goal for this issue to document this behavior public.

must108:
## Summary
This PR adds documentation that clarifies how compression works in clickhouse-go, especially for the HTTP protocol where there are two independent compression layers:
- HTTP web compression via `Accept-Encoding`/`Content-Encoding`
- ClickHouse native block compression over HTTP via `compress=1`/`decompress=1`

This change specifically updates the README to explain which `compress=` values map to which layer, shows that `lz4hc` is native-only, and adds guidance to avoid double compressing.

Fixes #1766

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,7 +171,7 @@ conn.SetConnMaxLifetime(time.Hour)
     * round_robin - choose a round-robin server from the set
     * in_order    - first live server is chosen in specified order
 * debug - enable debug output (boolean value)
-* compress - specify the compression algorithm: `none` (default), `zstd`, `lz4`, `lz4hc`, `gzip`, `deflate`, `br`. If set to `true`, `lz4` will be used.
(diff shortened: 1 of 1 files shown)


kavirajk (approved):
Thanks @must108 for documenting this. LGTM

PR #1838, opened 2026-04-20
docs: Add documentation for driver.Batch

### LINKED ISSUE #1774: driver.Batch documentation is missing
`driver.Batch` interface is defined like this:

```go
type Batch interface {
	Abort() error
	Append(v ...any) error
	AppendStruct(v any) error
	Column(int) BatchColumn
	Flush() error
	Send() error
	IsSent() bool
	Rows() int
	Columns() []column.Interface
	Close() error
}
```

And there is zero documentation how `Flush`, `Send`, and `Close` are related to each other. Which of those you have to call to send the data? Can you reuse the batch object after `Send` or `Flush`?

There is an [example](https://github.com/ClickHouse/clickhouse-go/blob/v2.43.0/examples/clickhouse_api/batch.go) which implies `Close` should be called through defer, and `Send` actually sends the currently pending data, but examples are poor substitutes for docs, and `Flush` purpose still remains unclear.

`WithCloseOnFlush` and `WithReleaseConnection` are not properly explained either.

There is at least [some documentation for `batch.Close` ](https://github.com/ClickHouse/clickhouse-go/blob/v2.43.0/conn_batch.go#L319-L324), but with `batch` being unexported type, it isn't viewable in go doc.

must108:
## Summary
This PR improves driver.Batch documentation to clearly show how `Flush`, `Send`, and `Close` relate, and shows how to use batches safely, and what to expect across protocols.

Fixes #1774

### DIFF
--- a/.claude/skills/review-pr/SKILL.md
+++ b/.claude/skills/review-pr/SKILL.md
@@ -75,3 +75,4 @@ Work through each category before writing the review.
 - [ ] Make sure new docs and examples are added if needed for new features or bug fixes.
 - [ ] Make sure the docs are covered for both TCP and HTTP protocol cases.
 - [ ] Make sure the docs are covered for both `clickhouse_native` (Open() api returns) api and `std` api (OpenDB() api returns).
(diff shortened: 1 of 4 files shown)


kavirajk (approved):
Thanks @must108 👍 LGTM

PR #1840, opened 2026-04-20
fix: honor CompressionZSTD for native block compression

### LINKED ISSUE #1772: Setting `CompressionZSTD` has no effect
## Observed

Both LZ4 and ZSTD compression methods simply set `compress=1`:

https://github.com/ClickHouse/clickhouse-go/blob/aebb65f56bab7308e4359d70c04ca44535b937d4/conn_http_query.go#L42-L43

There's no indication whether to use LZ4 or ZSTD, so the response is always LZ4.

## Expected behaviour

ZSTD is actually used.

### Environment
* [x] `clickhouse-go` version: v2.41.0
* [x] Interface: ClickHouse API, HTTP
* [x] Go version: go1.25.5
* [x] Operating system: Linux
* [x] ClickHouse version: 24.8.15.33588

must108:
## Summary
This PR fixes HTTP native block compression settings so `CompressionZSTD` and `CompressionLZ4` no longer behave the same. 

Fixes #1772 

## Checklist


must108:
@chernser @kavirajk bumping, would love for you guys to take a look!

must108:
@chernser @kavirajk bumping, would love for you guys to take a look!

must108:
@mshustov good to check again!

kavirajk:
@must108 do you think we should tackle TCP side also in this PR? Also works if you can create a an issue to tack the TCP.

conn_http_query.go
```diff
@@ -50,7 +50,11 @@ func (h *httpConnect) query(ctx context.Context, release nativeTransportRelease,
 	headers := make(map[string]string)
 	switch h.compression {
 	case CompressionZSTD, CompressionLZ4:
-		options.settings["compress"] = "1"
+		compressionLevel := 0
+		if h.opt != nil && h.opt.Compression != nil {
+			compressionLevel = h.opt.Compression.Level
+		}
+		applyHTTPNativeCompressionSettings(options.settings, h.compression, compressionLevel, false)
```
kavirajk, line 57: I see we pass the `network_compression_method` in HTTP requests correctly now 👍 

But I don't see we fix the places where we choose compression reader/writer to correctly decode the blocks? e.g: now we send [non-default compression](https://clickhouse.com/docs/reference/settings/session-settings/network#network_compression_method) say zstd, and blocks are now compressed with `zstd` instead of `lz4` (default). how does our query path decodes it correctly? I don't see relevant code is touched in this PR. 

Just wondering how it was working and how it will work now?

kavirajk (approved):

must108:
> @must108 do you think we should tackle TCP side also in this PR? Also works if you can create a an issue to tack the TCP.

I think we can create a new issue if you have not already @kavirajk

conn_http_query.go
must108: https://github.com/must108/clickhouse-go/blob/86977961a553c012ac04a7dace7278f651b530d1/conn_http.go#L455-L458

`readData()` should enable native compression whenever the method is either `lz4` or `zstd`. So I think this concern is already handled

must108 (commented):
Responding to comments

PR #1842, opened 2026-04-21
Create cross-repo-bug-relay.yml

alex-clickhouse:
Sets up new issue relay for cross-repo investigation.

If you'd rather trigger this on labeling an issue a bug we can do this instead:

```
  on:
    issues:
      types: [labeled]

  permissions: {}

  jobs:
    relay:
      if: github.event.label.name == 'bug'
      uses: ClickHouse/integrations-shared-workflows/.github/workflows/cross-repo-bug-relay.yml@main
      secrets: inherit
```

But there's a triage before the cross-repo investigation that filters out irrelevant issues.


kavirajk (approved):
LGTM 👍 thanks

PR #1844, opened 2026-04-21
fix: correct escaped single quote parsing in Enum type strings

### LINKED ISSUE #1839: Bug: parsing escaped enum names with multiple escape sequences
## Summary

Enum names that contain escaped single quotes (e.g. `DateTime(\'UTC\')`) are not parsed correctly when:
1. A single enum name contains **more than one** escape sequence, or
2. Any earlier enum value contains escape sequences (positions bleed into the next value).

## Root Cause

In `lib/column/enum.go`, `extractEnumNamedValues` has two bugs:

### Bug 1 — `skippedValueTokens` is never reset between values

The slice that records backslash positions is declared before the main loop and is never cleared when a new string starts. Positions from a previous enum value are re-applied to the next value's `foundName`, corrupting it.

### Bug 2 — stale positions when removing multiple backslashes

The removal loop (lines 152-154) applies the original source positions without adjusting for prior removals. Each removal shortens the slice by 1, so subsequent positions are off-by-one per preceding removal.

## Reproducing Query

```sql
SELECT variantType(CAST(toDateTime('2024-01-15 10:30:00', 'UTC'), 'Variant(String, DateTime(\'UTC\'))'))
```

ClickHouse returns a value of type `Enum8('String' = 0, 'DateTime(\'UTC\')' = 1)`. The Go client parses the enum name incorrectly, yielding a corrupted string instead of `DateTime('UTC')`.

## Related

Central tracking issue: ClickHouse/integrations-ai-playground#17

chaewonkong:
## Summary

 Fix incorrect parsing of escaped single quotes in `Enum8`/`Enum16` type strings. 

  Two bugs existed in `extractEnumNamedValues` in `lib/column/enum.go`:

  1. `skippedValueTokens` was never reset between values, causing escape positions from a previous enum
   value to bleed into the next one.
  2. When removing multiple backslashes from a single value, positions were not adjusted for prior
  removals, causing off-by-one corruption.

- closes:  #1839 

## Checklist


chaewonkong:
The CI failure in test (1.25, 25.10) seems unrelated to this PR. Test1229 is flaky and fails intermittently on main.

must108 (changes requested):
Hi @chaewonkong, thanks for the contributions. Please address the nits from agent, LGTM otherwise.

Issue #1841, opened 2026-04-21
Bug: a non-nil `*string` holding valid JSON stored at `{}`


kavirajk:
`AppendRow` unset-version [switch matches](https://github.com/ClickHouse/clickhouse-go/blob/27f6ead3480350e25094d12b567f556efc2c561c/lib/column/json.go?plain=1#L422-L424) `string` and `[]byte`, but not `*string`. [The reflect-based](https://github.com/ClickHouse/clickhouse-go/blob/27f6ead3480350e25094d12b567f556efc2c561c/lib/column/json.go?plain=1#L428-L431) `[]byte` check only catches slice-of-uint8, so `*string` [falls through to](https://github.com/ClickHouse/clickhouse-go/blob/27f6ead3480350e25094d12b567f556efc2c561c/lib/column/json.go?plain=1#L434-L436) `appendRowObject`, where [no branch converts a pointer-to-string into JSON.](https://github.com/ClickHouse/clickhouse-go/blob/27f6ead3480350e25094d12b567f556efc2c561c/lib/column/json.go?plain=1#L446) 

`obj` stays nil [and is replaced with an empty](https://github.com/ClickHouse/clickhouse-go/blob/27f6ead3480350e25094d12b567f556efc2c561c/lib/column/json.go?plain=1#L481-L483) `chcol.NewJSON()`. Then `AppendRow` returns nil, the caller sees no error, and the column is stored as `{}`.

I think this bug is made visible [after this PR](https://github.com/ClickHouse/clickhouse-go/pull/1771) (which changed the order of append. first try object then string. If like before the PR first trying string would have worked)

Below is a simple example to reproduce on tip of the main branch
```go
func bug(ctx context.Context, conn driver.Conn) error {
	if err := conn.Exec(ctx, "DROP TABLE IF EXISTS go_json_repro_issue1"); err != nil {
		return err
	}
	if err := conn.Exec(ctx, `CREATE TABLE go_json_repro_issue1 (product JSON) ENGINE=Memory`); err != nil {
		return err
	}

	jsonText := `{"id":1234,"name":"Book","tags":["a","b"]}`
	ptr := &jsonText

	batch, err := conn.PrepareBatch(ctx, "INSERT INTO go_json_repro_issue1 (product)")
	if err != nil {
		return err
	}
	if err := batch.Append(ptr); err != nil {
		return fmt.Errorf("append returned an error (would be preferable): %w", err)
	}
[...]

PR #1850, opened 2026-04-22
fix: choosing serialization version correctly in JSON column type

kavirajk:
## Summary

JSON is a complex type in ClickHouse. In the native protocol, choosing "how to serialize" (server call it `serialization version`) is a subtle problem. Two things needs to be understood.

#### 1. Serialization version in JSON

ClickHouse server supports multiple ways to serialize it. Two types are explicitly available at the Native format way. String or Object. [In server terms it's STRING and FLATTEN](https://github.com/ClickHouse/ClickHouse/blob/644c366011ddcd97122dcf43511967701463e43f/src/DataTypes/Serializations/SerializationObject.h?plain=1#L46-L57). In the client these are called String and Object serialization version.

With String - the whole JSON is represented as plain string. Each row you send to or receive from the server is a String type.

With Object - the whole response is more complex Dynamic/Variant type of JSON that is represented as Blocks.

Go client doesn't make user to "choose" which one to use for serialization, instead it chooses based on the value.

This change refactored to make this choosing more explicit.

#### 2. Serialization version per block per column.

Each block is a collection of column of same type in Native format. So [ClickHouse have to choose single "way" to serialize the whole block columns.](https://github.com/ClickHouse/ClickHouse/blob/644c366011ddcd97122dcf43511967701463e43f/src/DataTypes/Serializations/SerializationObject.cpp?plain=1#L265-L275) For types like JSON, this is tricky because both String based and Object JSON are valid values for each row in same column in same block, but ClickHouse have to "choose" one serialization version for it.
That's why it's enforced during Append.

fixes: #https://github.com/ClickHouse/clickhouse-go/issues/1841

#### 1. *string pointer usage

Say we have code like following
```go
	jsonText := `{"id":1234,"name":"Book","tags":["a","b"]}`
	ptr := &jsonText

	batch, err := conn.PrepareBatch(ctx, "INSERT INTO go_json_repro_issue1 (product)")
	if err != nil {
		return err
	}
	if err := batch.Append(ptr); err != nil {
		return fmt.Errorf("append returned an error (would be preferable): %w", err)
	}
	if err := batch.Send(); err != nil {
		return err
	}

	var stored string
	if err := conn.QueryRow(ctx,
		"SELECT toString(product) FROM go_json_repro_issue1").Scan(&stored); err != nil {
		return err
	}

	fmt.Printf("  input (*string):  %s\n", jsonText)
	fmt.Printf("  stored value:     %s\n", stored)
	fmt.Printf("  data lost:        %t\n", stored != jsonText)
```

Before the patch (`*string` data is lost silently)
```bash
$ go run examples/json_repro_issues/main.go
Issue 1: *string holding JSON is silently stored as {}
  input (*string):  {"id":1234,"name":"Book","tags":["a","b"]}
  stored value:     {}
  data lost:        true

```

After the patch (data is not lost)
```bash
$ go run examples/json_repro_issues/main.go
Issue 1: *string holding JSON is silently stored as {}
  input (*string):  {"id":1234,"name":"Book","tags":["a","b"]}
  stored value:     {"id":1234,"name":"Book","tags":["a","b"]}
  data lost:        false

```

#### 2. nil usage with both string and object value

Say you have code like following
```go
	jsonText := `{"id":2,"name":"Book"}`

	// Scenario A: nil first, then a JSON string.
	// Fixed behavior: nil defers, string latches String mode, nil flushes as
	// the JSON literal "null" — both rows round-trip correctly.
	fmt.Println("  Scenario A: Append(nil) then Append(jsonString)")
	batch, err := conn.PrepareBatch(ctx, "INSERT INTO go_json_repro_issue2 (id, product)")
	if err != nil {
		return err
	}
	if err := batch.Append(uint32(1), nil); err != nil {
		return fmt.Errorf("append nil: %w", err)
	}
	if err := batch.Append(uint32(2), jsonText); err != nil {
		return fmt.Errorf("append string: %w", err)
	}
	if err := batch.Send(); err != nil {
		return fmt.Errorf("send: %w", err)
	}
	if err := dumpRows(ctx, conn, "go_json_repro_issue2"); err != nil {
		return err
	}

	// Scenario B: same rows, reordered (string first, then nil).
	fmt.Println("  Scenario B: Append(jsonString) then Append(nil) — reordered")
	if err := conn.Exec(ctx, "TRUNCATE TABLE go_json_repro_issue2"); err != nil {
		return err
	}
	batch, err = conn.PrepareBatch(ctx, "INSERT INTO go_json_repro_issue2 (id, product)")
	if err != nil {
		return err
	}
	if err := batch.Append(uint32(2), jsonText); err != nil {
		return err
	}
	if err := batch.Append(uint32(1), nil); err != nil {
		return err
	}
	if err := batch.Send(); err != nil {
		return fmt.Errorf("send (reordered): %w", err)
	}
	if err := dumpRows(ctx, conn, "go_json_repro_issue2"); err != nil {
		return err
	}

```

Before the patch 
1. appending `nil` first choose `object` serialization version and appending string next lost data silently
2. appending `string` first choose `string` serialization version and appending `nil` next cause error because `nil` cannot be serialized as string before.
```bash
$ go run examples/json_repro_issues/main.go
Issue 2: nil-first on Nullable(JSON) — previously locked object mode
  Scenario A: Append(nil) then Append(jsonString)
    id=1  product=
    id=2  product={}
  Scenario B: Append(jsonString) then Append(nil) — reordered
2026/04/22 22:32:58 issue 2 repro failed: send (reordered): code: 117, message: Cannot parse JSON object here:
```

After the patch
1. `nil` doesn't have any serialization version. It's deferred to the time of flushing based on first non-null value.
2. so both scenarios (`nil` then `string`, `string` then `nil` both works)
```bash
$ go run examples/json_repro_issues/main.go
Issue 2: nil-first on Nullable(JSON) — previously locked object mode
  Scenario A: Append(nil) then Append(jsonString)
    id=1  product=
    id=2  product={"id":2,"name":"Book"}
  Scenario B: Append(jsonString) then Append(nil) — reordered
    id=1  product=
    id=2  product={"id":2,"name":"Book"}

```

#### 3. Mixing serialization mode in same block

Say you have following code
```go
	jsonText := `{"id":2,"name":"Book"}`

	jsonObj1 := struct {
		Name string
		Id   int32
	}{"kavi", 3}

	batch, err := conn.PrepareBatch(ctx, "INSERT INTO go_json_repro_issue2 (id, product)")
	if err != nil {
		return err
	}
	if err := batch.Append(uint32(1), jsonObj1); err != nil {
		return fmt.Errorf("append object: %w", err)
	}
	if err := batch.Append(uint32(2), jsonText); err != nil {
		return fmt.Errorf("append string: %w", err)
	}
	if err := batch.Send(); err != nil {
		return fmt.Errorf("send: %w", err)
	}
	if err := dumpRows(ctx, conn, "go_json_repro_issue2"); err != nil {
		return err
	}
	return nil

```

Before the patch
 It picks the serializaiton mode whatever inserted first and second append silently lost data
```bash
$ go run examples/json_repro_issues/main.go
Mixed serialization mode: use String mode and Object mode in same block
    id=1  product={"Id":3,"Name":"kavi"}
    id=2  product={} # data lost
```

After the patch
It throws error saying you cannot have different serialization version per single block.
```bash
$ go run examples/json_repro_issues/main.go
Mixed serialization mode: use String mode and Object mode in same block
2026/04/22 22:44:26 mixed mode error: append string: clickhouse [AppendRow]: product clickhouse: JSON column is already using object serialization after row 1; cannot append a value that requires string serialization. The wire format allows only one serialization version per column per block — every row in a batch must use the same mode

```

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -420,6 +420,33 @@ For `clickhouse.Conn.PrepareBatch` (native interface):
 - Use `Send` to flush any remaining rows and finalize the INSERT. After `Send`, the batch is considered sent and should not be reused.
 - Use `defer batch.Close()` to ensure resources are released if `Send` is not reached.
 
+## JSON columns: append contract
+
+The ClickHouse Native protocol requires **one serialization version per `JSON` column per block** — a column cannot mix `object` rows and `string` rows on the wire. The driver enforces this at append time.
+
+**Two modes, one per batch:**
+
+- `object` — the driver decomposes a value into typed/dynamic paths. Accepts: `struct`, `map[string]any`, `*struct`, `*map`, `*clickhouse.JSON`, and any type implementing `clickhouse.JSONSerializer`.
+- `string` — the driver stores raw JSON text. Accepts: `string`, `*string`, `[]byte`, `*[]byte`, `json.RawMessage`, `*json.RawMessage`, `sql.NullString`, `*sql.NullString`, and types implementing `driver.Valuer` or `fmt.Stringer`.
+
+**Null rows are mode-agnostic.** `nil`, typed-nil pointers (`(*string)(nil)`, `(*clickhouse.JSON)(nil)`), `*interface{}` holding nil, and `sql.NullString{Valid: false}` do **not** latch a mode. They are buffered until a non-null row chooses the mode, and then flushed into the chosen backing column. `Nullable(JSON)` works the same way — the null mask lives on the Nullable wrapper; the inner JSON column still needs to emit something that parses server-side.
+
+**The first non-null row picks the mode.** Subsequent rows must match:
+
+```go
+batch.Append(struct{ Name string }{"Alice"}) // latches "object"
+batch.Append(`{"x":1}`)                      // error: string in an object-mode column
+```
+
+**Mixed-mode appends return an error**, identifying the type of the rejected row. There is no silent `{}` fallback.
+
+**All-null batches** default to `string` mode at send time and encode each null row as the JSON literal `"null"` (smaller on the wire than an empty object, and valid JSON so the server accepts the payload in `Nullable(JSON)` String mode).
+
+**Columnar bulk inserts (`batch.Column(i).Append(slice)`)** follow the same rules:
+- `[]string`, `[]*string`, `[][]byte`, `[]*[]byte`, `[]json.RawMessage`, `[]*json.RawMessage`, `[]sql.NullString`, `[]*sql.NullString` → `string` mode.
+- `[]struct{...}`, `[]map[string]any`, `[]clickhouse.JSON`, `[]*clickhouse.JSON`, `[]clickhouse.JSONSerializer` → `object` mode.
+- `Append` expects a slice — passing a single scalar returns an error. Use `AppendRow` for per-row inserts.
+
 ## Benchmark
 
 Indicative numbers measured on: Linux 6.19.6-arch1-1 · Intel Core Ultra 7 258V (8 cores) · 30 GiB RAM · NVMe SSD. Run the linked programs directly to get numbers on your hardware, e.g. `go run benchmark/v2/read/main.go`. Go benchmark tests can be run with `go test -bench=. ./benchmark/...`.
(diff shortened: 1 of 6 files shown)


chernser (approved):

chernser:
@kavirajk 
tests looks correct
main question was about how nulls deferred: 
1. flushed when non null value appended 
2. flushed when batch is sent 

So fix I think is correct.

PR #1852, opened 2026-04-29
fix: restore debug logging when debug=1 is set via DSN

### LINKED ISSUE #1835: Debug logs missing in v2.43.0
## Observed

I came from https://github.com/amacneil/dbmate/issues/786

1. added `debug=1` to DSN. For example: `clickhouse://default:default@172.17.0.1:9440/test?secure=1&skip_verify=1&debug=1` - there is no debug log in stdout

### Environment
* [x] Interface: `github.com/ClickHouse/ch-go` v0.71.0
* [x] Go version: 1.26.1
* [x] Operating system: Linux amd64
* [x] ClickHouse version: 26.3.3.20 (official build)

wucm667:
Fixes #1835

## Summary

When `debug=1` is set in the DSN (e.g., `clickhouse://host/db?debug=1`), no debug logs appeared in stdout since v2.43.0.

## Root Cause

The `logger()` method in `clickhouse_options.go` required both `Debug=true` AND `Debugf != nil` to produce a working logger. When only `debug=1` is set via DSN, `Debug` is set to true but `Debugf` remains nil, so the method fell through to the noop logger.

## Fix

Added a fallback in the `logger()` method: when `Debug=true` but `Debugf` is nil, create a default `slog.Logger` that outputs to stdout at Debug level.

## Changes

- `clickhouse_options.go`: Add default stdout logger fallback when `Debug=true` without `Debugf`
- `clickhouse_options_test.go`: Add `TestLogger` covering 4 scenarios:
  - `debug=1` via DSN produces non-noop logger
  - No debug flag produces noop logger
  - `Debug=true` with `Debugf` uses legacy handler (backward compat)
  - Custom `Logger` option takes precedence

## Testing

```bash
go test -v -count=1 .
```

All tests pass, including existing `TestParseDSN`, `TestLegacyDebugfBackwardCompatibility`, and new `TestLogger`.

### DIFF
--- a/clickhouse_options.go
+++ b/clickhouse_options.go
@@ -9,6 +9,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"os"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


wucm667:
@github-actions nice catch on the priority order — good thing you flagged it. Pushed a new commit that moves `Logger` ahead of the `Debug` fallback, and added a test for `Debug=true` + `Logger=customLogger` to make sure the custom logger isn't silently dropped.

Also cleaned up the duplicate subtests (already covered in `logger_test.go`), renamed the noop one, and pulled the stdout logger into a `newStdoutDebugLogger()` helper.

kavirajk (approved):
LGTM. thanks @wucm667 

One failing test case is a legit problem at the head of the ClickHouse latest master branch.

PR #1855, opened 2026-05-05
Add cluster interserver-secret authentication

BorisTyshkevich:
Add cluster interserver-secret authentication

A trusted internal service can authenticate as a ClickHouse cluster peer by
sharing the cluster secret instead of a user password, and execute each query
as an arbitrary `initial_user` which the server accepts via
`AlwaysAllowCredentials`. Mirrors the existing distributed-query protocol
documented in `src/Server/TCPHandler.cpp`.

API
---
- `Options.Cluster{Name, Secret}` configures interserver mode.
- `WithInitialUser` sets the impersonated user per query when interserver mode
  is active (falls back to `Auth.Username`). It is a no-op otherwise.
- `ClusterCredentials.String/GoString` redact `Secret` so accidental logging
  of the Options struct cannot leak it.
- Hello sends `" INTERSERVER SECRET "`, empty password, cluster name, and a
  32-byte random salt when `Cluster.Secret` is set.
- Each query encodes `SHA256(salt + secret + body + id + initial_user)` into
  the interserver-secret slot and flips `ClientQueryInitial` to
  `ClientQuerySecondary`. When `Cluster.Secret` is empty the slot stays the
  legacy empty string, so existing callers are unaffected.
- Driver advertises protocol revision 54460, so nonce and externally-granted
  roles fields from `DBMS_MIN_REVISION_WITH_INTERSERVER_SECRET_V2` (54462) and
  later are intentionally omitted.

Validation
----------
Fail-closed validation runs on both the native and `database/sql` surfaces:
- `ErrClusterSecretRequiresName` — `Cluster.Secret` without `Cluster.Name`.
- `ErrClusterSecretNeedsNative` — `Cluster.Secret` with `Protocol: HTTP`.
- `ErrClusterSecretRequiresUsername` — `Cluster.Secret` with empty
  `Auth.Username` (otherwise `setDefaults` rewrites it to `"default"` and a
  forgotten `WithInitialUser` would silently run as the cluster superuser).
- `ErrClusterSecretWithJWT` — `Cluster.Secret` combined with `GetJWT`.

A Warn line is emitted when interserver mode is active, and a second Warn when
it is active without TLS. The V1 protocol the driver speaks has no nonce, so an
on-path attacker on a plaintext link could replay a captured signed query frame
on the same connection.

DSN: deliberately not supported
-------------------------------
`Cluster.Secret` is sensitive cluster-wide credential material. DSNs end up in
startup logs, error messages, config files, and stack traces; the secret must
not. The driver intentionally has no DSN parameter for the cluster secret—keep
it in memory (env var → `Options{}`).

Tests
-----
- `lib/proto/query_interserver_test.go` pins a known hash vector, decodes the
  exact interserver-hash slot, checks its 32-byte frame delta, and verifies
  `query_kind` flips Initial→Secondary in actual encoded output.
- `tests/interserver_test.go` covers the end-to-end flow against a real server:
  creates a passwordless user, opens a connection with `Cluster.Secret`, runs a
  query under `WithInitialUser`, and asserts `system.query_log` shows
  `is_initial_query=0` with `user=initial_user`. Negative paths cover
  wrong-secret rejection by the server, validation errors, and redaction.
- `tests/std/interserver_test.go` exercises validation through both `OpenDB`
  and `Connector`.
- Fixture-dependent e2e tests skip ClickHouse Cloud and remote servers which do
  not mount `tests/resources/custom.xml`.
- `tests/resources/custom.xml` adds a `<test_cluster_secret>` entry so the
  single-node test container accepts incoming `" INTERSERVER SECRET "`
  connections naming that cluster.

Docs
----
- README gains "Cluster interserver-secret authentication" with Security
  model, Why-not-DSN rationale, and a comparison with the new `EXECUTE AS` SQL
  statement (server version coverage, connection-pool friction,
  identity-in-query-text surface, audit-trail differences).
- `examples/clickhouse_api/cluster_secret.go` runnable example sources the
  secret from `CLICKHOUSE_CLUSTER_SECRET`.


must108 (changes requested):
Hi @BorisTyshkevich, thanks for the contributions. Please address the agent review where it is reasonable. Thanks!

clickhouse.go
```diff
@@ -67,7 +70,32 @@ func Open(opt *Options) (driver.Conn, error) {
...
 	}
+	// Inspect Auth.Username before setDefaults rewrites it to "default": if
+	// the caller enabled interserver-secret mode, we force them to name the
+	// fallback user explicitly so a forgotten WithInitialUser cannot silently
+	// run as the cluster superuser.
+	if opt.Cluster.Secret != "" && opt.Auth.Username == "" {
+		return nil, ErrClusterSecretRequiresUsername
+	}
 	o := opt.setDefaults()
+	if err := o.validate(); err != nil {
```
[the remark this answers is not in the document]
BorisTyshkevich: Fixed in 70482e05. Validation now runs before setDefaults on Open, Connector, OpenDB, and stdDriver.Open. Connector/OpenDB retain the validation error in stdConnOpener so it is returned from Connect/Ping, and tests/std/interserver_test.go covers the username, cluster-name, protocol, and JWT cases through both stdlib entry points.

clickhouse_options.go
```diff
@@ -435,6 +468,21 @@ func (o Options) setDefaults() *Options {
...
+// touching the network. Validation is invoked from Open after setDefaults so
+// the caller never reaches the dial path with mutually inconsistent fields.
+func (o *Options) validate() error {
+	if o.Cluster.Secret != "" {
+		if o.Cluster.Name == "" {
+			return ErrClusterSecretRequiresName
+		}
+		if o.Protocol != Native {
+			return ErrClusterSecretNeedsNative
+		}
```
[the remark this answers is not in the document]
BorisTyshkevich: Fixed in 70482e05. Cluster.Secret combined with GetJWT now returns ErrClusterSecretWithJWT before dialing, with coverage through both Connector and OpenDB.

tests/interserver_test.go
```diff
@@ -0,0 +1,220 @@
...
+
+// TestInterserverSecretAuthenticatesAsInitialUser verifies the end-to-end
+// interserver-secret flow: a connection that authenticates with the cluster
+// secret can run a query as a non-default user without supplying that user's
+// password, and the query is logged in `system.query_log` as Secondary.
+func TestInterserverSecretAuthenticatesAsInitialUser(t *testing.T) {
+	env, err := GetNativeTestEnvironment()
+	require.NoError(t, err)
+
+	if !CheckMinClickHouseVersion(t, env, 23, 3, 0) {
```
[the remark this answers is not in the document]
BorisTyshkevich: Fixed in 70482e05. Both fixture-dependent e2e tests now call SkipOnCloud and also skip RemoteClickHouse, since tests/resources/custom.xml is only mounted for the local test container.

clickhouse_options.go
```diff
@@ -68,6 +68,35 @@ type Auth struct { // has_control_character
 	Password string
 }
 
+// ClusterCredentials configures client-side interserver authentication.
+// When Secret is non-empty, the client authenticates as a trusted cluster
+// peer using the shared cluster secret instead of a user password, and the
+// server executes queries as the InitialUser set on the connection's Auth
```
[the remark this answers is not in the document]
BorisTyshkevich: Fixed in 70482e05. The GoDoc now names Auth.Username as the fallback, states that the target user must exist, and the marker text in tests/resources/custom.xml and the PR description now matches the single-spaced protocol constant.

context.go
```diff
@@ -98,6 +99,18 @@ func WithJWT(jwt string) QueryOption {
 	}
 }
 
+// WithInitialUser sets the `initial_user` sent with the query. When the
+// connection is configured with a cluster interserver secret, the server
+// executes the query as this user without a password check. Without an
+// interserver secret this only sets the `initial_user` ClientInfo field
```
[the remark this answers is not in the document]
BorisTyshkevich: Fixed in 70482e05. WithInitialUser is now documented as interserver-only, and effectiveInitialUser returns an empty value for ordinary connections so their legacy wire output remains unchanged. A unit test covers ordinary, override, and fallback behavior.

lib/proto/query_interserver_test.go
```diff
@@ -0,0 +1,160 @@
...
+		InitialUser:   "alice",
+	}
+
+	h := sha256.New()
+	h.Write([]byte(q.ClusterSalt))
+	h.Write([]byte(q.ClusterSecret))
+	h.Write([]byte(q.Body))
+	h.Write([]byte(q.ID))
+	h.Write([]byte(q.InitialUser))
+	want := string(h.Sum(nil))
```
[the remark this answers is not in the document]
BorisTyshkevich: Addressed both test improvements in 70482e05: the hash expectation is now a pinned hex vector, and the loose byte scan is replaced by deterministic decoding of the exact interserver-hash field, verification of the following StateComplete byte, and the 32-byte signed/unsigned frame-length delta.

One clarification on the rationale: an implementation-only field-order regression would already have failed the old test, because the expected sequence was written separately in the test and would have retained the original order. The pinned vector is still stronger because it also protects against accidentally changing the implementation and expectation together. Likewise, a 32-byte length delta alone proves payload size, not field position; that is why the updated test checks both the delta and the exact decoded slot.

Slach:
@must108 all review comments fixed in https://github.com/ClickHouse/clickhouse-go/pull/1855/commits/70482e055f33cdf2234f92497ad6395f2b40b083, could you trigger CI/CD?

examples/clickhouse_api/cluster_secret.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/ClickHouse/clickhouse-go/v2"
+)
+
+// ClusterSecretAuth shows how to authenticate as a trusted cluster peer
```
must108: Please reduce the size of these comments. They should be short and effective

lib/proto/const.go
```diff
@@ -31,6 +31,12 @@ const (
 	ClientPing   = 4
 )
 
+// UserInterserverMarker is the special username clients send to
```
must108: Reduce size

lib/proto/query.go
```diff
@@ -28,6 +29,13 @@ type Query struct {
 	Compression              bool
 	InitialUser              string
 	InitialAddress           string
+	// ClusterSecret enables interserver-secret query signing. When non-empty
```
must108: Reduce size a little bit

lib/proto/query.go
```diff
@@ -68,8 +76,34 @@ func swap64(b []byte) {
 	}
 }
 
+// interserverHash computes the query signature expected by the ClickHouse
```
must108: Reduce size

lib/proto/query_interserver_test.go
```diff
@@ -0,0 +1,175 @@
...
+	a := base
+	a.Body = "SELECT 1"
+	b := base
+	b.Body = "SELECT 2"
+	if a.interserverHash() == b.interserverHash() {
+		t.Fatal("hash must differ when body differs")
+	}
+}
+
+// TestEncodeClientInfoQueryKind verifies that the query_kind byte flips to
```
must108: Reduce size

lib/proto/query_interserver_test.go
```diff
@@ -0,0 +1,175 @@
...
+				Body:          "SELECT 1",
+				InitialUser:   "alice",
+				ClusterSecret: tc.secret,
+				ClusterSalt:   "salt",
+			}
+			buf := &chproto.Buffer{}
+			if err := q.Encode(buf, DBMS_TCP_PROTOCOL_VERSION); err != nil {
+				t.Fatalf("Encode failed: %v", err)
+			}
+			// Skip the leading query ID string: var-len header + body.
```
must108: Reduce size, this can definitely be one line

lib/proto/query_interserver_test.go
```diff
@@ -0,0 +1,175 @@
...
+				t.Fatalf("read query_kind: %v", err)
+			}
+			if gotKind != tc.want {
+				t.Fatalf("query_kind = %d, want %d", gotKind, tc.want)
+			}
+		})
+	}
+}
+
+// TestEncodeEmptySecretUsesLegacyHashSlot verifies the exact encoded hash
```
must108: Reduce

tests/resources/custom.xml
```diff
@@ -34,6 +34,22 @@
     <display_name>clickhouse</display_name>
     <custom_settings_prefixes>custom_</custom_settings_prefixes>
 
+    <!-- Used by interserver-secret integration tests: a single-shard cluster
```
must108: Reduce

tests/interserver_test.go
```diff
@@ -0,0 +1,231 @@
...
+		t.Skip("interserver-secret negotiation is exercised against >= 23.3 servers")
+	}
+
+	admin, err := TestClientWithDefaultSettings(env)
+	require.NoError(t, err)
+	defer admin.Close()
+
+	const initialUser = "interserver_test_user"
+	createUser(t, admin, initialUser)
+	defer dropUser(t, admin, initialUser)
```
must108: Address failing tests here

tests/interserver_test.go
```diff
@@ -0,0 +1,231 @@
...
+
+// testClusterSecret matches the <secret> configured for <test_cluster_secret>
+// in tests/resources/custom.xml. Hard-coded because the cluster name and the
+// secret are server-side configuration; the test only needs to mirror them.
+const (
+	testClusterName   = "test_cluster_secret"
+	testClusterSecret = "test_interserver_secret"
+)
+
+// TestInterserverSecretAuthenticatesAsInitialUser verifies the end-to-end
```
must108: Reduce size here and in line 15

tests/interserver_test.go
```diff
@@ -0,0 +1,231 @@
...
+		LIMIT 1
+	`, queryID).Scan(&loggedUser, &loggedInitialUser, &isInitialQuery)
+	require.NoError(t, err, "expected query_log row for query_id=%s", queryID)
+
+	require.Equal(t, initialUser, loggedUser, "server must run query as initial_user")
+	require.Equal(t, initialUser, loggedInitialUser)
+	require.Equal(t, uint8(0), isInitialQuery, "interserver-secret query must be Secondary (is_initial_query=0)")
+}
+
+// TestInterserverSecretWrongSecretRejected verifies the server refuses a
```
must108: Reduce

tests/interserver_test.go
```diff
@@ -0,0 +1,231 @@
...
+		Auth:     clickhouse.Auth{Username: "default"},
+		Cluster: clickhouse.ClusterCredentials{
+			Name:   testClusterName,
+			Secret: testClusterSecret,
+		},
+	})
+	require.ErrorIs(t, err, clickhouse.ErrClusterSecretNeedsNative)
+}
+
+// TestInterserverSecretRequiresExplicitUsername closes the implicit-default
```
must108: Reduce here, and line 169, 187, 203

clickhouse.go
```diff
@@ -70,7 +74,25 @@ func Open(opt *Options) (driver.Conn, error) {
 	if opt == nil {
 		opt = &Options{}
 	}
+	if err := opt.validate(); err != nil {
+		return nil, err
+	}
 	o := opt.setDefaults()
+	if o.Cluster.Secret != "" {
+		// Surfacing this at Warn level (not Debug) so accidental enablement
```
must108: Reduce size here, and 89

clickhouse_options.go
```diff
@@ -68,6 +68,35 @@ type Auth struct { // has_control_character
 	Password string
 }
 
+// ClusterCredentials configures client-side interserver authentication.
```
must108: Reduce size here, and line 86, 154, 487

conn.go
```diff
@@ -137,12 +137,30 @@ type connect struct {
 	maxCompressionBuffer int
 	readerMutex          sync.Mutex
 	closeMutex           sync.Mutex
+	// clusterSalt is the 32-byte salt sent during an interserver handshake
```
must108: Reduce size here, 150

context.go
```diff
@@ -98,6 +99,17 @@ func WithJWT(jwt string) QueryOption {
 	}
 }
 
+// WithInitialUser sets the `initial_user` sent with the query when the
```
must108: Reduce size here, an slightly on line 113

must108 (changes requested):
Functionality wise, LGTM. Please address the comment lengths and the failing test cases

tests/interserver_test.go
BorisTyshkevich: Fixed in dd4f3971. The new test duplicated the existing `dropUser` helper. It now reuses that helper, checks cleanup errors, and both integration-test packages compile.

lib/proto/query.go
BorisTyshkevich: Shortened in dd4f3971.

tests/interserver_test.go
BorisTyshkevich: Shortened in dd4f3971.

lib/proto/query_interserver_test.go
BorisTyshkevich: Shortened in dd4f3971.

lib/proto/query.go
BorisTyshkevich: Shortened in dd4f3971.

context.go
BorisTyshkevich: Shortened in dd4f3971.

tests/resources/custom.xml
BorisTyshkevich: Shortened in dd4f3971.

conn.go
BorisTyshkevich: Shortened in dd4f3971.

clickhouse_options.go
BorisTyshkevich: Shortened in dd4f3971.

tests/interserver_test.go
BorisTyshkevich: Shortened in dd4f3971.

tests/interserver_test.go
BorisTyshkevich: Shortened in dd4f3971.

lib/proto/query_interserver_test.go
BorisTyshkevich: Shortened in dd4f3971.

lib/proto/const.go
BorisTyshkevich: Shortened in dd4f3971.

examples/clickhouse_api/cluster_secret.go
BorisTyshkevich: Shortened in dd4f3971.

clickhouse.go
BorisTyshkevich: Shortened in dd4f3971.

lib/proto/query_interserver_test.go
BorisTyshkevich: Shortened in dd4f3971.

must108 (approved):
LGTM. Great work!

PR #1865, opened 2026-05-30
Add type safety for settings.

YlanzinhoY:
## Summary

I've opened this PR to improve the DX. This prevents us from having to constantly check the documentation to see the available Settings options and type them manually. We now have proper typing for Settings options, without breaking existing systems that currently rely on strings.

I also ran the tests and no errors were found.

✅ Type-safety: Known configurations are now typed constants
✅ Safety: Reduces the risk of typos in configuration names
✅ Documentation: Constants serve as inline documentation

## how it is done today
```go
  
    Settings: clickhouse.Settings{
        "max_execution_time": 60,
    }
```

## What would happen if the PR is approved?

```go
    Settings: clickhouse.Settings{
      clickhouse.MaxExecutionTime: 60,
    }
```

If it gets approved, I can update the documentation.

## Checklist


YlanzinhoY (commented):
OK, I don't think that turned out so well.

YlanzinhoY (commented):
I removed the Copilot modification and simply left the constants at the package level; I preferred to keep it simple.

YlanzinhoY (commented):
I removed all redundant conversions and conversions from SettingsType (since we don't use it anymore).
It's practically back to how it was before I opened the PR.

context.go
```diff
@@ -17,6 +17,20 @@ var _contextOptionKey = &QueryOptions{
 	},
 }
 
+const (
```
kavirajk: I like the idea. Can we move it to different package? (`serversettings` for example).

Then at the usage point, it makes it more clear.

```
  
    Settings: clickhouse.Settings{
        serversettings.MaxExecutionTime: 60,
    }
```

In addition to clear usage, it also
1. avoids conflict with may other local variables in the client context.
2. one package to keep all the server related settings. We can rename or add more easily.

Of course having these settings as constants have one caveat though, if anything changes on server side (say settings renamed, deprecated or removed, etc) we need to handle it on client side. That's an additional work for maintainers. 

But I think having at least common settings as constants really helps.

kavirajk (commented):
thanks @YlanzinhoY for the PR. I like the idea. Proposed few changes.

context.go
YlanzinhoY: I liked the idea; we'll use a package that contains the constants for the server settings.

And it really does add extra work for the maintainers, because whenever a new option is added here, we'll have to have this constant, but I think that's fine.

I'll make that change.

kavirajk: But you made those settings part of `internal` package. In Go anything [internal package](https://go.dev/doc/modules/layout#package-or-command-with-supporting-packages) will not be seen by external consumers it's only for project's internal usage.

Which makes this feature almost not helpful for users? I don't know if that's intentional.

Also would be better to actually "use" those settings everywhere in the tests, examples, etc (once you moved that values to say `serversettings` package)

YlanzinhoY: Can I create a folder called `serversettings` in the project root, removing it from the internal directory?

internal/settingsType/settings_types_constants.go
```diff
@@ -0,0 +1,15 @@
+package settingsType
```
rbroggi, line 1: Go convention typically don't use cammelCase for package names. Could probably move into a `settings` or `chsettings` package in the root of the project to allow for visibility outside the repo. `internal` will not be visible.

kavirajk:
After some thinking. I guess having separate package for settings constants may not scale, given the [surface area of thousands of server settings.](https://clickhouse.com/docs/operations/server-configuration-parameters/settings)

If we start supporting few "most" used settings. We may have few problems
1. "most" is not objective. And strictly it can different depend on the uses cases. Which low-level library like clickhouse-go should not have any opinion on IMHO.
2. We may end up in the situation, where we have to support both for indefinite period of time. And any consumer needs to be aware of this cognitive knowledge that weather they have to use the constants or just strings. they may end up using both every time. Not a good UX IMHO.

Even if anyone wants to have a settings constants, they can do it completely via additional package or helpers that is independent of this clickhouse-go library.

PR #1869, opened 2026-06-01
fix: evict connections on wrapped net.OpErrors

### LINKED ISSUE #1868: Connection write network errors do not evict connection from the pool
## Observed

1. A write to ClickHouse fails with a `*net.OpError` (e.g. write deadline exceeded).
2. `sendData` returns an error but does not call `setClosed()` — the connection is returned to the pool as healthy.
3. The next query on that connection fails immediately. The pool does not retry on a fresh connection.

We experienced this in production when a ClickHouse cluster was temporarily unable to serve requests after accepting connections. After the cluster recovered the application servers continued to fail with "write tcp ... i/o timeout" errors. Suggesting those connections were returned to the pool and checked out again, but could not be used successfully. We've observed this 3 or 4 times, and recovery from the error always takes less time than the max connection lifetime we have configured, which also supports the theory connections are being reused despite being corrupted.

## Expected behaviour

A connection that fails with a `*net.OpError` during `sendData` should be marked closed and evicted from the pool. Subsequent queries should transparently retry on a new connection.

## Code example

The bug can be reproduced without network conditions using a mock `net.Conn`:

```go
package clickhouse

import (
    "net"
    "os"
    "testing"
)

type writeErrConn struct {
    net.Conn
    err error
}

func (c *writeErrConn) Write([]byte) (int, error) { return 0, c.err }
func (c *writeErrConn) RemoteAddr() net.Addr      { return &net.TCPAddr{} }

func TestSendDataLeavesConnectionOpen(t *testing.T) {
    opErr := &net.OpError{Op: "write", Net: "tcp", Err: os.ErrDeadlineExceeded}
    conn := &connect{conn: &writeErrConn{err: opErr}, /* ... */}

    conn.sendData(proto.NewBlock(), "")

    // Fails: connection is NOT marked closed, but should be
    if !conn.isClosed() {
        t.Fatal("expected connection to be closed after *net.OpError write failure")
    }
}
```

There are two code-level causes:

**1. `sendData` only calls `setClosed()` for `syscall.EPIPE` and `io.EOF`** (`conn.go`). The `default` branch handles all other errors — including `*net.OpError` — without marking the connection bad. Any write `*net.OpError` leaves the ClickHouse protocol framing in an unrecoverable state, so the connection cannot be safely reused.

**2. `isConnBrokenError` uses a direct type assertion** (`clickhouse_std.go`):
```go
if _, ok := err.(*net.OpError); ok { ... }
```
`flush()` wraps the error as `fmt.Errorf("write: %w", err)`, and `sendData` wraps it again. By the time `isConnBrokenError` sees the error in the `database/sql` path, the direct assertion always fails. `errors.As` is needed to walk the chain.

## Error log

The write failure is returned to the caller with this message (same in both buggy and fixed versions — the bug is silent):

```log
send data: write error to 127.0.0.1:9000 (conn_id=1, block_cols=1, block_rows=0): write: write tcp: i/o timeout
```

The symptom is the caller's *next operation* failing with a server exception, rather than the pool transparently retrying on a fresh connection. A write timeout can leave partial bytes on the wire; the server is waiting for the rest of that incomplete message. When the connection is reused, the next query's bytes are misinterpreted as a continuation of the previous one, and the server returns a parse error.

### Environment
* [x] `clickhouse-go` version: v2.46.0 (present on `main`)
* [x] Interface: both ClickHouse API and `database/sql`
* [x] Go version: 1.25
* [ ] Operating system: any
* [ ] ClickHouse version: any
* [ ] Is it a ClickHouse Cloud: N/A
* [ ] ClickHouse Server non-default settings: N/A
* [ ] `CREATE TABLE` statements: N/A

Grundlefleck:
## Summary
Fixes: https://github.com/ClickHouse/clickhouse-go/issues/1868

`isConnBrokenError` now uses errors.As() to detect when an error returned by `sendData()` has wrapped a net.OpError. This allows the database/sql connection pooling to evict the connection.

It also changes `sendData()` to explicitly close a connection on net.OpError. So that the native client can evict corrupted connections as well.

## Checklist


conn.go
```diff
@@ -339,6 +342,15 @@ func (c *connect) sendData(block *proto.Block, name string) error {
 			c.setClosed()
 			return fmt.Errorf("send data: unexpected EOF to %s (conn_id=%d, block_cols=%d, block_rows=%d): %w",
 				c.conn.RemoteAddr(), c.id, len(block.Columns), block.Rows(), err)
+		case isOpErr:
+			// *net.OpError not already caught by more specific EPIPE/EOF check
+			c.logger.Error("connection broken: write error",
+				slog.Any("error", err),
+				slog.Int("block_columns", len(block.Columns)),
+				slog.Int("block_rows", block.Rows()))
+			c.setClosed()
```
Grundlefleck, line 351: Note the `default:` branch continue to avoid calling `c.setClosed()`. This is in keeping with the spirit of [PR 871](https://github.com/ClickHouse/clickhouse-go/pull/871)  and trying to resort to close only when there is a high probability the connection is irretrievably broken. In this case, although we don't have the definitive signal of EPIPE or EOF to say it's broken, AIUI any *net.OpError returned from Write() means the binary protocol stream is in an indeterminate state. Server might have received partial data with no way to cancel or recover, so the connection cannot safely be reused.

jmorton-clerk (approved):
LGTM 🐤

kavirajk (commented):
Thanks @Grundlefleck for the PR. LGTM 👍 

Also thanks for catching connection not being closed correctly on the default branch as well.

kavirajk (approved):

Grundlefleck:
Thanks @kavirajk!

PR #1873, opened 2026-06-03
Update to not use deprecated `github.com/docker/docker`

tempoz:
## Summary
Update to use `github.com/moby/moby/api` instead of `github.com/docker/docker`, as the latter has been deprecated and is now subject to multiple CVEs. See [here](https://github.com/moby/moby#go-modules).

Note that these dependencies, afaik, are only used in tests.

## Checklist

_None relevant._


kavirajk (approved):
Thanks @tempoz LGTM

PR #1875, opened 2026-06-07
feat: add struct column extraction helper

### LINKED ISSUE #1874: Expose struct column extraction helper for AppendStruct inserts
## Observed

1. Existing issue #587 documents that `AppendStruct` fails when a table contains columns that are absent from the Go struct, even when those columns have server-side `DEFAULT` expressions.
2. The current workaround is to explicitly list the inserted columns in `PrepareBatch`, for example `INSERT INTO events (id, name, timestamp)`, so ClickHouse can apply defaults to omitted columns.
3. That workaround is effective, but it duplicates the struct's `ch` tags in raw SQL. In wide or frequently evolving ingestion structs, the column list can drift from the struct definition.

## Expected behaviour

The driver should expose a small helper that extracts the ordered ClickHouse column names represented by a struct, so callers can build explicit insert column lists from the same source of truth used by `AppendStruct`.

This does not need to change `AppendStruct` semantics or add an implicit `DESCRIBE` round trip. It would make the documented workaround for #587 easier to use correctly.

## Code example

```go
package main

import (
    "context"
    "fmt"
    "strings"

)

type Event struct {
    ID   uint64 `ch:"id"`
    Name string `ch:"name"`
}

func insert(ctx context.Context, conn clickhouse.Conn, events []Event) error {
    columns, err := clickhouse.StructColumns(Event{})
    if err != nil {
        return err
    }

    batch, err := conn.PrepareBatch(ctx, fmt.Sprintf(
        "INSERT INTO events (%s)",
        strings.Join(columns, ", "),
    ))
    if err != nil {
        return err
    }
    defer batch.Close()

    for i := range events {
        if err := batch.AppendStruct(&events[i]); err != nil {
            return err
        }
    }
    return batch.Send()
}
```

With a table like this, ClickHouse can apply `source` server-side because the generated insert statement only names `id` and `name`:

```sql
CREATE TABLE events (
    id UInt64,
    name String,
    source String DEFAULT 'server default'
) ENGINE = MergeTree()
ORDER BY id;
```

## Error log

Current open-ended insert behaviour still fails as described in #587:

```log
clickhouse [AppendStruct]: missing destination name "source" in *main.Event
```

### Environment
* [x] `clickhouse-go` version: current `main` and released v2 versions with `AppendStruct`
* [x] Go version: not version-specific
* [x] Operating system: not OS-specific
* [x] ClickHouse version: any version where the table has `DEFAULT` columns
* [x] `CREATE TABLE` statements for tables involved: see code example above

Related: #587

rbroggi:
## Summary

- add `clickhouse.StructColumns` to extract deterministic insert column lists from structs using the same mapping rules as `AppendStruct`/`ScanStruct`
- support `ch` tag options in struct mapping by using the name before the first comma
- document using explicit column lists with `AppendStruct` when tables contain server-side `DEFAULT` columns
- add a ClickHouse API example that uses `StructColumns` so omitted columns can use their `DEFAULT` values
- harden `StructColumns` by de-duplicating returned column names and ignoring anonymous non-struct embeds without panicking

Fixes #1874
Related to #587

## Testing

- `go test -race -count=1 -v . -run 'TestStructColumns|TestStructIdx|TestMapper'`
- `make up && go test -race -count=1 -v ./tests -run TestAppendStruct && go test -race -count=1 -v ./examples/clickhouse_api -run TestStructColumns && make down`
- `go test -count=1 ./...` was also run earlier; it failed in existing JSON integration tests against `clickhouse/clickhouse-server:latest` (`26.5.1`) with timestamp parsing errors in `TestJSONString`, unrelated to this change.


rbroggi:
Thanks for the review. I pushed `b48cc74` to address the recommended robustness fixes:

- de-duplicated `StructColumns` output in first-occurrence order, so generated `INSERT` column lists do not contain duplicate identifiers
- guarded anonymous embed traversal with `Kind() == reflect.Struct` in both `StructColumns` and `structIdx`, avoiding panics for embedded named scalar/interface types
- removed the redundant cache-miss return copy and added a comment explaining why the `StructColumns` cache is package-level
- expanded tests for duplicate columns and anonymous non-struct embeds

Re-ran:

- `go test -race -count=1 -v . -run 'TestStructColumns|TestStructIdx|TestMapper'`
- `make up && go test -race -count=1 -v ./tests -run TestAppendStruct && go test -race -count=1 -v ./examples/clickhouse_api -run TestStructColumns && make down`

must108 (changes requested):
@rbroggi please address agent review. Thanks!

struct_map.go
```diff
@@ -3,13 +3,58 @@ package clickhouse
...
 	"reflect"
+	"strings"
 	"sync"
 )
 
 type structMap struct {
 	cache sync.Map
 }
 
+// structColumnsCache is package-level because StructColumns is a package-level
```
must108, line 14: Reduce size of comment

struct_map.go
```diff
@@ -3,13 +3,58 @@ package clickhouse
...
+	return copyStrings(columns.([]string)), true
+}
+
+func (c *structColumnCache) Store(t reflect.Type, columns []string) {
+	c.cache.Store(t, copyStrings(columns))
+}
+
+// StructColumns returns the ClickHouse column names represented by v.
+//
+// StructColumns follows the same field mapping rules as AppendStruct and
```
must108, line 37: Reduce size of comment

must108 (changes requested):

PR #1879, opened 2026-06-16
fix: ignore placeholders in quoted contexts

### LINKED ISSUE #1860: bindPositional treats '?' inside backtick identifiers and string literals as a placeholder
## Description

`bindPositional` in `bind.go` (lines 103-161) scans the query byte-by-byte and replaces every `?` with the next positional argument. The only escape mechanism it recognizes is a preceding backslash (`\?`). It does **not** track whether the `?` appears inside:

- a backtick-quoted identifier (`` `field?` ``),
- a single-quoted string literal (`'foo?bar'`), or
- a double-quoted identifier (`"field?"`).

As a result, any user query that legitimately contains a `?` inside a quoted name or string is corrupted on the client side before being sent to ClickHouse.

The earlier tracker ClickHouse/clickhouse-go#878 was closed-completed in 2023 after adding the `\?` escape, but the loop was never taught to enter/exit quote state. The same class of bug was recently reported against the JDBC v2 driver in ClickHouse/clickhouse-java#2290.

The detection regexes used to decide which binding mode to use are also context-unaware:

```go
bindNumericRe    = regexp.MustCompile(`\$[0-9]+`)
bindPositionalRe = regexp.MustCompile(`[^\\][?]`)
```

so the `ErrBindMixedParamsFormats` heuristic in `bind()` can also false-positive when `$N` or `?` simply appears inside a literal/identifier.

## ClickHouse server version

Code analysis only; not verified against a running server. The defect is purely in the client-side query preprocessing inside `bind.go` on `main`.

## Reproduction

Minimal Go test that demonstrates the corruption. Add to `bind_test.go` and run with `go test -run TestBindPositional_QuestionMarkInQuotedContexts ./...`:

```go
func TestBindPositional_QuestionMarkInQuotedContexts(t *testing.T) {
    cases := []struct {
        name  string
        query string
        args  []any
        want  string
    }{
        {
            name:  "question mark inside backtick identifier",
            query: "SELECT * FROM (SELECT '1' AS `field?`) WHERE `field?` = ?",
            args:  []any{"1"},
            want:  "SELECT * FROM (SELECT '1' AS `field?`) WHERE `field?` = '1'",
        },
        {
            name:  "question mark inside single-quoted string literal",
            query: "SELECT * FROM t WHERE name = 'foo?bar' AND id = ?",
            args:  []any{42},
            want:  "SELECT * FROM t WHERE name = 'foo?bar' AND id = 42",
        },
    }

    for _, tc := range cases {
        t.Run(tc.name, func(t *testing.T) {
            got, err := bind(time.UTC, tc.query, tc.args...)
            if err != nil {
                t.Fatalf("bind returned error: %v", err)
            }
            if got != tc.want {
                t.Errorf("\nwant: %q\n got: %q", tc.want, got)
            }
        })
    }
}
```

**Actual output (current `main`):**

- backtick case → `` SELECT * FROM (SELECT '1' AS `field1`) WHERE `field?` = ? `` (first `?` consumed by the argument; trailing `?` left dangling). It also returns an error because there is one arg but two `?` characters were matched.
- single-quote case → `SELECT * FROM t WHERE name = 'foo42bar' AND id = ?` (string literal corrupted, trailing `?` not bound, error returned).

**Expected output:** only the `?` outside any quoted context should be substituted.

## Suggested fix

Replace the single-state byte loop in `bindPositional` with a small state machine that tracks quote/backtick state. Suggested rules:

- toggle `inBacktick` on an unescaped `` ` ``,
- toggle `inSingle` on `'` (respecting `\'` and the `''` doubling convention used by ClickHouse string literals),
- toggle `inDouble` on `"` (respecting `\"`),
- only substitute `?` when none of those flags are set,
- keep the existing `\?` escape behaviour for backwards compatibility.

The same parser should ideally be used to drive the `haveNumeric` / `havePositional` detection in `bind()` (currently `bindNumericRe` / `bindPositionalRe`), so the `MixedParamsFormats` check does not false-positive on `$N` or `?` characters that appear inside literals or identifiers.

Relevant locations:
- `bind.go:50-53` — context-unaware detection regexes
- `bind.go:103-161` — `bindPositional`, the buggy scanner

## Link

Same class of bug reported on the JDBC v2 driver: ClickHouse/clickhouse-java#2290. Prior incomplete fix in this repo: ClickHouse/clickhouse-go#878.

wucm667:
## Summary
Closes #1860.

- replace context-unaware bind mode regex checks with a quote-aware scanner for backticks, single-quoted strings, and double-quoted identifiers
- make positional and numeric binding ignore placeholders inside quoted contexts while preserving the existing `\?` literal escape behavior
- add regression coverage for quoted `?` / `$N` markers and mixed-format false positives

## Checklist

## Testing
- `go test -run 'TestBind(Numeric|Positional|MixedParamsFormats)' .`
- `golangci-lint run`
- `CLICKHOUSE_VERSION=26.4 go test ./...`

Note: I also tried `CLICKHOUSE_VERSION=26.4 make test`; the local race run timed out in existing `tests/issues/Test1421BatchFlushBrokenConn` after 4m, while the non-race supported-version suite above passed.


bind.go
```diff
@@ -162,13 +252,16 @@ func bindPositional(tz *time.Location, query string, args ...any) (_ string, err
...
+		unbind         = make(map[string]struct{})
+		params         = make(map[string]string)
+		buf            = make([]byte, 0, len(query))
+		state          bindQuoteState
 	)
 	for i, v := range args {
 		if fn, ok := v.(std_driver.Valuer); ok {
 			if v, err = fn.Value(); err != nil {
-				return "", nil
+				return "", err
```
kavirajk, line 264: thanks for fixing this previous bug as well 👍

kavirajk (approved):
thanks @wucm667 for the PR.

I think we may still have to fix the similar error in `bindNamed`.  But I will handle that in separate PR.

PR #1894, opened 2026-07-01
fix: bind float params as typed Float literals (#1862)

### LINKED ISSUE #1862: bind: float64(1.0) serialised as integer "1" — CH narrows to UInt8 and breaks float64 scans
## Repro

When binding a Go `float64(1.0)` (or any integer-valued `float64`) as a parameter to `clickhouse-go/v2`, `lib/bind/format.go::format()` — the `format(tz, scale, v any)` function — has no `case float64:` (nor `case float32:`) branch. The value falls through every type switch and hits the trailing `return fmt.Sprint(v), nil` on the bottom path. `fmt.Sprint(float64(1.0))` yields the bare string `"1"` (no decimal), so the SQL the driver assembles carries an integer literal.

Confirmed in `v2.46.0` (`bind.go` lines 277-391):

```go
func format(tz *time.Location, scale TimeUnit, v any) (string, error) {
    quote := func(v string) string { ... }
    switch v := v.(type) {
    case nil:        return "NULL", nil
    case string:     return quote(v), nil
    case time.Time:  return formatTime(tz, scale, v)
    case *time.Time: ...
    case bool:       ...
    case GroupSet, []GroupSet, ArraySet, fmt.Stringer, column.OrderedMap, column.IterableOrderedMap: ...
    }
    switch v := reflect.ValueOf(v); v.Kind() {
    case reflect.String:               ...
    case reflect.Slice, reflect.Array: ...
    case reflect.Map:                  ...
    case reflect.Ptr:                  ...
    }
    return fmt.Sprint(v), nil  // <-- float64(1.0) ends up here
}
```

Reproducer (Go):

```go
package main

import "fmt"

func main() {
    var x float64 = 1.0
    var y float64 = 1.5
    fmt.Printf("%q %q\n", fmt.Sprint(x), fmt.Sprint(y))
    // prints: "1" "1.5"
}
```

## Expected

`1.0` (or `cast(1, 'Float64')`, or `toFloat64(1)`) — anything that preserves the column type as `Float64` on the wire so downstream typed scans see a Float64 column.

## Actual

`1` — a bare integer literal in the rendered SQL. ClickHouse infers `UInt8` for that literal (smallest unsigned integer type that fits), and the resulting expression carries `UInt8` (or a promoted `UInt16` / `UInt32` after arithmetic) downstream.

## Impact

When the bound parameter is used as a value column projected back to a typed Go scan target (`*float64`), `clickhouse-go/v2` refuses the conversion at `Scan` time:

```
converting UInt8 to *float64 is unsupported
```

(or `UInt16 -> *float64` after a `* 2` style binop on the bound literal). This surfaces in downstream consumers as a 502 / 500 on otherwise valid queries.

## Workaround used downstream

In [cerberus](https://github.com/tsouza/cerberus) — a Prometheus / Loki / Tempo HTTP gateway that emits ClickHouse SQL via `clickhouse-go/v2` — every bound float literal is wrapped at the SQL-emission layer in `toFloat64(?)` so CH projects `Float64` on the wire regardless of the driver's narrowing. The same wrap is applied to integer-valued literals used as float sentinels (e.g. `any(toFloat64(1))` for `group(<vector>)`, `toFloat64(1)` for `absent_over_time`). Examples:

- `tsouza/cerberus#634` — synthetic-scalar binop, wrapped the scalar literal in `toFloat64(?)`.
- `tsouza/cerberus#644` — `group(<vector>)` lowered to `any(toFloat64(1))` rather than `any(1)` to survive `UInt8 -> *float64` Scan.
- `tsouza/cerberus#197` series — analogous wraps for the `count(...)` / `absent_over_time(...)` lowerings.

The wrap is correct but costs an extra `toFloat64(...)` per bound float across the emitter surface, and the same trap will catch any other consumer that binds `float64` and scans the result as `*float64`.

## Suggested fix

Add an explicit `case float64:` (and ideally `case float32:`) branch to `format()` that emits the value with a forced decimal:

```go
case float64:
    // strconv.FormatFloat with 'g' and -1 precision keeps the value
    // exact and round-trippable; 1.0 -> "1" (loses the type signal)
    // is the bug we are avoiding, so use a form that pins the
    // wire type as Float64 on the CH side:
    return fmt.Sprintf("cast(%s, 'Float64')", strconv.FormatFloat(v, 'g', -1, 64)), nil
case float32:
    return fmt.Sprintf("cast(%s, 'Float32')", strconv.FormatFloat(float64(v), 'g', -1, 32)), nil
```

Alternatively, `strconv.FormatFloat(v, 'f', -1, 64)` and ensure a decimal is always present (post-process to append `.0` when the rendered form has no `.` and no `e`). The `cast(..., 'Float64')` form is the most defensive because it pins the type at the CH parser level regardless of the literal's textual shape.

The `NaN` / `±Inf` corner is worth handling in the same branch — CH parses `inf` / `-inf` / `nan` (lowercase) but `fmt.Sprint(math.Inf(1))` emits `"+Inf"` (mixed case), so the same fall-through path produces a syntax error on non-finite floats today. Downstream cerberus has a hand-rolled `1.0/0` / `-1.0/0` / `0.0/0` workaround for the same root cause; a first-class `case float64:` branch should fold that in too.

## Versions

- `github.com/ClickHouse/clickhouse-go/v2` `v2.46.0` (latest at time of filing — bug reproduces on the line numbers cited above).

arpitjain099:
Binding a float parameter renders it wrong today. A float64/float32 falls through the switch in `format()` to `fmt.Sprint`, so `1.0` comes out as the bare literal `1`. ClickHouse then reads that as an integer, and a later typed scan into a `*float64` fails with "converting UInt8 to *float64 is unsupported". Non-finite values are broken too, since Go prints `NaN` and `+Inf` but ClickHouse only parses the lowercase `nan`/`inf` spellings, so binding one produces a SQL syntax error.

This adds float32/float64 cases (plus the matching reflect kinds) that render the value as `cast(<value>, 'Float32'|'Float64')`, which pins the type and uses the spellings ClickHouse accepts for non-finite values. Closes #1862.

I added a `TestFormatFloat` regression test in `bind_test.go` covering integer-valued, fractional, and non-finite floats for both widths. Thanks for taking a look.

### DIFF
--- a/bind.go
+++ b/bind.go
@@ -4,7 +4,9 @@ import (
 	std_driver "database/sql/driver"
 	"errors"
 	"fmt"
+	"math"
 	"reflect"
+	"strconv"
 	"strings"
 	"time"
 
@@ -487,6 +489,10 @@ func format(tz *time.Location, scale TimeUnit, v any) (string, error) {
 			return "1", nil
 		}
 		return "0", nil
+	case float32:
+		return formatFloat(float64(v), 32), nil
+	case float64:
+		return formatFloat(v, 64), nil
 	case GroupSet:
 		val, err := join(tz, scale, v.Value)
 		if err != nil {
@@ -573,6 +579,10 @@ func format(tz *time.Location, scale TimeUnit, v any) (string, error) {
 			values = append(values, fmt.Sprintf("%s, %s", name, val))
 		}
 		return "map(" + strings.Join(values, ", ") + ")", nil
+	case reflect.Float32:
+		return formatFloat(v.Float(), 32), nil
+	case reflect.Float64:
+		return formatFloat(v.Float(), 64), nil
 	case reflect.Ptr:
 		if v.IsNil() {
 			return "NULL", nil
@@ -580,6 +590,27 @@ func format(tz *time.Location, scale TimeUnit, v any) (string, error) {
 		return format(tz, scale, v.Elem().Interface())
 	}
 	return fmt.Sprint(v), nil
+}
+
+// formatFloat renders a float as a CAST to the matching ClickHouse Float type.
+// Without the cast, integer-valued floats like 1.0 render as the bare literal
+// "1", which ClickHouse infers as an integer and later narrows (breaking typed
+// float scans). NaN and infinities are quoted in the lowercase form ClickHouse
+// accepts, since Go's default formatting ("NaN", "+Inf") is not valid SQL.
+func formatFloat(f float64, bitSize int) string {
+	chType := "Float64"
+	if bitSize == 32 {
+		chType = "Float32"
+	}
+	switch {
+	case math.IsNaN(f):
+		return fmt.Sprintf("cast('nan', '%s')", chType)
+	case math.IsInf(f, 1):
+		return fmt.Sprintf("cast('inf', '%s')", chType)
+	case math.IsInf(f, -1):
+		return fmt.Sprintf("cast('-inf', '%s')", chType)
+	}
+	return fmt.Sprintf("cast(%s, '%s')", strconv.FormatFloat(f, 'g', -1, bitSize), chType)
 }
 
 func join[E any](tz *time.Location, scale TimeUnit, values []E) (string, error) {
--- a/bind_test.go
+++ b/bind_test.go
@@ -1,6 +1,7 @@
 package clickhouse
 
 import (
+	"math"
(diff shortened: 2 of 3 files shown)


kavirajk (approved):
LGTM 👍 thanks @arpitjain099

PR #1897, opened 2026-07-03
fix: escaping for string keys in Map parameter binding

destinyoooo:
Changes:

- Updated reflect.Map parameter formatting so map keys reuse the shared format() escaping path.
- Prevent invalid SQL generation when map[string]... keys contain single quotes or backslashes.
- Added a regression test for map string keys containing ' and \.


kavirajk (approved):
LGTM. thanks @destinyoooo

PR #1905, opened 2026-07-10
fix: use atomic.Int64 for clickhouse.connID (386 alignment)

forkwhileant:
# Problem
On 32-bit platforms (GOARCH=386), `atomic.AddInt64(&ch.connID, 1)` in `dial()` panics with "unaligned 64-bit atomic operation" on the very first connection attempt. This is because the `*Options` field preceding `connID int64` in `type clickhouse struct` is 4 bytes, so `connID` lands at struct offset 4.

# To reproduce
Run `GOOS=linux GOARCH=386 CGO_ENABLED=0 go build`, run the binary on a 32-bit host, call `clickhouse.Open()`, and then call any method that dials.

# Fix
Switch to `sync/atomic.Int64`, which embeds an `align64` marker so the compiler 8-byte-aligns the field regardless of what precedes it. See the note at the top of the `sync/atomic` package docs for more info.


forkwhileant:
I'm gonna be honest y'all, it's not worth my time to file a ticket with legal to get them to review your CLA. Feel free to close this PR but I encourage you to fix this bug.

kavirajk:
Hi @forkwhileant thanks for your contribution. I would really love to merge this changes as your contribution.
is there anything I can do to make the process of signing CLA more friendly? 

Unfortunately the CLA is required by external contributors of any ClickHouse OSS projects and difficult to make any exceptions.

forkwhileant:
Not really. maybe just make this same change with your own commit?

kavirajk (approved):
LGTM. thanks @forkwhileant . We decided to merge as it given the minimal change.

Q: I'm curious where is it you using the library for 32-bit arch, if it's ok to share.

Issue #1670, opened 2025-09-26
Inconsistent error types returned between the `Native` and `HTTP` protocols.


r0bobo:
## Observed

The returned errors are of different types when switching between the `Native` and `HTTP` protocols.
We have a large codebase that uses the `Native` protocol but are trying to switch to the `HTTP` protocol
to be able to use HTTP load balancers in front of ClickHouse.
After switching to `HTTP` protocol error assertions don't catch errors any more since the underlying error type has changed.

## Expected behaviour

Same error types returned from both `HTTP` and `Native` protocol.

## Code example

```go
package main

import (
	"errors"
	"testing"

	"github.com/ClickHouse/clickhouse-go/v2"
)

func TestErrors(t *testing.T) {
	for _, tt := range []struct {
		port     string
		protocol clickhouse.Protocol
	}{
		{
			protocol: clickhouse.Native,
			port:     "9000",
		},
		{
			protocol: clickhouse.HTTP,
			port:     "8123",
		},
	} {
		t.Run(tt.protocol.String(), func(t *testing.T) {
			db := clickhouse.OpenDB(&clickhouse.Options{
				Addr:     []string{"localhost:" + tt.port},
				Protocol: tt.protocol,
			})
			_, err := db.Query("SELECT * FROM non_existent_table")
			var chErr *clickhouse.Exception
			if errors.As(err, &chErr) {
				if chErr.Code != 60 {
					t.Errorf("unknown error: %s", err.Error())

				}
			} else if err != nil {
				t.Errorf("unknown error: %s", err.Error())
			}
		})
	}
}
```

## Error log

```log
❯ go test ./test/...
--- FAIL: TestErrors (0.00s)
    --- FAIL: TestErrors/http (0.00s)
        foo_test.go:37: unknown error: sendQuery: [HTTP 404] response body: "Code: 60. DB::Exception: Unknown table expression identifier 'non_existent_table' in scope SELECT * FROM non_existent_table. (UNKNOWN_TABLE) (version 25.1.5.31 (official build))
            "
FAIL
FAIL    github.com/formulatehq/ingest/test      0.009s
FAIL
```

### Environment
* [x] `clickhouse-go` version: v2.40.3
* [x] Interface: `database/sql` compatible driver
* [x] Go version: go1.25.0 linux/amd64
* [x] Operating system: Fedora 42
* [x] ClickHouse version: 25.1.5.31
* [x] Is it a ClickHouse Cloud? No
[...]

PR #1910, opened 2026-07-13
feat: support structured Exception for HTTP protocol

### LINKED ISSUE #1877: Structured `*clickhouse.Exception` (or `clickhouse.Error` interface) on the HTTP protocol
### Discussed in [https://github.com/ClickHouse/clickhouse-go/discussions/1870](<https://github.com/ClickHouse/clickhouse-go/discussions/1870>)

Originally posted by **swvooda** June 2, 2026 When the HTTP protocol returns an error originating from a ClickHouse `DB::Exception`, the driver surfaces it as an ad-hoc `fmt.Errorf` string. Callers that need to branch on the ClickHouse error code (e.g. `60 UNKNOWN_TABLE`, `159 TIMEOUT_EXCEEDED`, `241 MEMORY_LIMIT_EXCEEDED`) are forced to regex-parse path on HTTP.

### Current behavior (HTTP)

On a query that fails server-side, callers receive an error whose only carrier of structured data is the message string:

```
sendQuery: [HTTP 404] response body: "Code: 60. DB::Exception: Unknown table expression identifier 'non_existent_table' in scope SELECT * FROM non_existent_table. (UNKNOWN_TABLE) (version 25.1.5.31 (official build))"
```

`errors.As(err, &clickhouse.Exception{})` does **not** match — the HTTP path constructs the error at [`conn_http.go:598`](<https://github.com/ClickHouse/clickhouse-go/blob/main/conn_http.go>) via `fmt.Errorf(...)`, so the typed `*clickhouse.Exception` that Native returns is lost. This was confirmed by @kavirajk in the ClickHouse/clickhouse-go#1670 thread.

### Expected behavior

Have structured error types like the other protocols do currently.

```go
type Exception struct {
    Code       int32   // ClickHouse server error code, e.g. 60
    Name       string  // Symbolic name, e.g. "UNKNOWN_TABLE" — already in the wire message
    Message    string  // The DB::Exception text, with version/scope trimming optional
    StackTrace string  // Optional; present in Native today
}
```

kavirajk:
## Summary

Closes #1877,  #1670

### Changes

Server errors over the HTTP protocol were plain `fmt.Errorf` strings, so `errors.As(err, &chErr)` worked on native but not on HTTP — anyone switching protocols had to regex-parse error messages.

Now the same code works on both protocols:

```
var chErr *clickhouse.Exception
if errors.As(err, &chErr) {
    // chErr.Code (60), chErr.CodeName ("UNKNOWN_TABLE"), chErr.Message
}

var httpErr *clickhouse.HTTPError
if errors.As(err, &httpErr) {
    // httpErr.StatusCode — HTTP only, non-200 responses
}
```

Also added addtional `CodeName` to parse the actual exception name corresponds to the code. Previous Name just represents class of the execption (e.g: `DB::Exceception`, `Net::Exception`). But not it represents the symbolic name (e.g: `UNKNOWN_IDENTIFIER`).

See the examples and tests for more cases

## Checklist


conn_http.go
```diff
@@ -598,16 +630,24 @@ func (h *httpConnect) readRawResponse(response *http.Response) (body []byte, err
...
-		reader = chReader
-	}
 
-	body, err = io.ReadAll(reader)
+	body, err = io.ReadAll(io.LimitReader(reader, maxErrorBodySize))
 	if err != nil && !errors.Is(err, io.EOF) {
 		return nil, err
 	}
+
+	if h.compression == CompressionLZ4 || h.compression == CompressionZSTD {
```
chernser, line 641: how would it handle http compression with turned on the server by default?

chernser (approved):

conn_http.go
kavirajk: Good point :+1:  there was a real gap, now fixed. Here are my understanding.

Default config: the driver sets `DisableCompression: true` on the transport and sends no `Accept-Encoding` AFAIK, so the server never compresses responses regardless of server settings and HTTP compression is opt-in per request.

PR #1913, opened 2026-07-13
fix: replace panic in unhex with error return

### LINKED ISSUE #1908: lib/churl/churl.go:529 — panic("invalid hex character") in unhex() crashes on malformed DSN input
## name: Bug report
about: panic("invalid hex character") in lib/churl unhex() crashes on malformed DSN URL

## Observed

1. `lib/churl/churl.go:529` calls `panic("invalid hex character")` in the `unhex()` helper when it encounters an invalid hex digit (e.g., `%GG`).
2. `unhex()` is called from 3 locations within the same file — lines 324, 335, 361 — all inside the internal URL percent-decoding logic.
3. The URL parsing is invoked when connecting via HTTP protocol with a DSN containing malformed percent-encoding (e.g., `http://user:pass@host:8123/?param=%GG`).
4. Since this is library code, the panic crashes the entire Go process, not just the connection attempt.

## Expected behaviour

`unhex()` should return an error instead of panicking. The error should be propagated up through the URL unescape chain and returned to the caller (either `clickhouse.Open()` or `sql.Open()`), so the application can handle it gracefully.

## Code example

```go
package main

import (
    "log"
)

func main() {
    // Malformed percent-encoding in DSN parameter
        Addr: []string{"localhost:8123"},
        Protocol: clickhouse.HTTP,
        Auth: clickhouse.Auth{
            Database: "default",
            Username: "default",
            Password: "",
        },
        Settings: map[string]interface{}{
            "param": "%GG",  // invalid hex - triggers panic
        },
    })
    if err != nil {
        log.Fatal(err) // never reached - process already panicked
    }
    _ = conn
}
```

## Error log

```
panic: invalid hex character

goroutine 1 [running]:
github.com/ClickHouse/clickhouse-go/v2/lib/churl.unhex(...)
        lib/churl/churl.go:529 +0xbc
github.com/ClickHouse/clickhouse-go/v2/lib/churl.unescape(0xc000..., {0xc000..., 0x3}, 0x0)
        lib/churl/churl.go:361 +0x...
...
```

## Suggested fix

Change unhex signature from func(c byte) byte to func(c byte) (byte, error) — return an error instead of panicking on invalid input.
Update all 3 call sites (lines 324, 335, 361) to handle the error and propagate it.
Ensure the error bubbles up through unescape() → URL parsing → back to the caller.

## Environment
clickhouse-go version: v2.47.0
Interface: ClickHouse API / database/sql
Go version: 1.25.0+

the-onewho-knocks:
## Summary
Replace `panic("invalid hex character")` in `lib/churl.unhex()` with a proper `(byte, error)` return. Fixes #1908.

The `unhex` helper previously panicked on invalid hex input (e.g., `%GG` in a DSN), crashing the entire Go process. Now it returns an error that propagates through `unescape()` → `Parse()` → caller, so malformed URLs are handled gracefully.

## Checklist

### DIFF
--- a/lib/churl/churl.go
+++ b/lib/churl/churl.go
@@ -321,8 +321,14 @@ func unescape(s string, mode encoding) (string, error) {
 			// But https://tools.ietf.org/html/rfc6874#section-2
 			// introduces %25 being allowed to escape a percent sign
 			// in IPv6 scoped-address literals. Yay.
-			if mode == encodeHost && unhex(s[i+1]) < 8 && s[i:i+3] != "%25" {
-				return "", neturl.EscapeError(s[i : i+3])
+			if mode == encodeHost {
+				h, err := unhex(s[i+1])
+				if err != nil {
+					return "", err
+				}
+				if h < 8 && s[i:i+3] != "%25" {
+					return "", neturl.EscapeError(s[i : i+3])
+				}
 			}
 			if mode == encodeZone {
 				// RFC 6874 says basically "anything goes" for zone identifiers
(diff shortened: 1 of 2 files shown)


the-onewho-knocks:
Thanks for the thorough review confirmed the panic is unreachable through churl.Parse since ishex catches everything before unhex. I closed #1908 alongside this PR.

PR #1914, opened 2026-07-13
fix: preserve per-query JWT in QueryOptions.clone()

AndreZiviani:
## Summary

Since v2.42.0, a per-query JWT set with `clickhouse.WithJWT` is silently ignored on the HTTP interface: the request falls back to the connection's basic-auth credentials and the query runs as the wrong principal, with no error surfaced.

**Root cause.** `applyOptionsToRequest` reads the per-query JWT from `queryOptions(ctx)`, which returns a `clone()` of the context's `QueryOptions`. `clone()` copies every field except `jwt`, so the token set via `WithJWT` is always empty by the time the request headers are built: `useJWT` evaluates to false, the `Authorization: Bearer` arm is skipped, and the request authenticates with `X-ClickHouse-User`/`X-ClickHouse-Key` instead.

The regression came in 55bfc3e ("Add context option to append more ClientInfo to the system.query_log", first released in v2.42.0), which switched `applyOptionsToRequest` from `queryOptionsJWT(ctx)` — which read `jwt` straight off the context without cloning — to `queryOpt.jwt` on the clone, without adding `jwt` to `clone()`.

**Impact.** Any HTTPS connection that combines connection-level basic-auth credentials with per-query `WithJWT` tokens (e.g. a shared pool where service queries use a password and user queries carry a JWT) executes the user's queries as the service account on affected versions. Because the query succeeds, the credential downgrade is invisible to the application — results simply reflect the wrong user's grants and row policies.

**Fix.** Copy `jwt` in `QueryOptions.clone()`, plus a regression test covering the round-trip (`Context(ctx, WithJWT(...))` → `queryOptions(ctx).jwt`), including the case where a second `Context(ctx, ...)` call re-clones existing options.

The regression test lives in `context_test.go` rather than `tests/issues/` because it exercises unexported internals (`queryOptions`/`clone`), and JWT auth is a ClickHouse Cloud feature that the Docker-based integration suite can't reproduce.

`make lint` and the unit tests pass locally.

**Provenance.** The root-cause investigation and initial patch were done with an AI assistant (Claude Code, noted as co-author on the commit). I reviewed the change manually and verified it end-to-end against ClickHouse Cloud: with the fix, a query issued with `WithJWT` over HTTPS runs as the JWT principal; without it, it runs as the connection's basic-auth user.

## Checklist

### DIFF
--- a/conn_http_test.go
+++ b/conn_http_test.go
@@ -2,6 +2,7 @@ package clickhouse
 
 import (
 	"context"
+	"crypto/tls"
 	"net/http"
 	"testing"
 )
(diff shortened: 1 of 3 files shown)


kavirajk (approved):
Thanks @AndreZiviani for the PR. changes LGTM. Added one regression test.

PR #1915, opened 2026-07-13
Rework connection liveness checking and connection reuse

SpencerTorres:
## Summary

While investigating a performance issue related to connections, several issues were identified and fixed:

* **`connCheck` is now a non-destructive peek** (`recvfrom(MSG_PEEK)` instead of a 1-byte
  `read`). Pending bytes on an idle connection still mark it bad, which is correct for Native protocol. In these cases a debug message is logged.
* **Liveness results are trusted for 1s** (via `connLivenessWindow`): a connection that
  completed a command or passed a check within the last second skips the socket
  probe. This removes the per-operation connection checks that read from the TCP connection.
* **New `Options.ConnIdlePingThreshold`** (DSN: `conn_idle_ping_threshold`, default 1m,
  negative disables): This is used for proper client Ping commands to validate the connection is valid. For cloud there are often a lot of proxies and other errors, so instead of depending on a client side socket check, we can now use a ping command to validate the connection before executing a command.
* **Server exceptions no longer burn the connection.** `release()` keeps the connection
  and marks it *unverified*; the next acquire confirms liveness with a ping before handing
  it out (covering the rare exception classes where the server does close). This should reduce the number of new connections formed for non-fatal connection errors.
* **`query()` releases the connection before closing the rows channels**, so the
  connection is guaranteed to be back in the pool by the time `rows.Close()` returns.
  Parameter-binding failures now release the connection
  healthy instead of through the error path.
* **`prepareBatch` returns the connection to the pool on validation failure** instead of
  leaking it and its `MaxOpenConns` slot.
* **batch `contextWatchdog` stop is now race-free** (mutex + stopped flag): once `stop()`
  returns, the callback has either completed or will never run. The `Send` watchdog also
  flags the connection closed (`setClosed`) so a cancelled batch's socket can never be
  pooled looking healthy.

### Behavior changes

| Change | Old | New | Escape hatch |
|---|---|---|---|
| Server exception | connection closed, next query re-dials | connection kept, ping-verified on next acquire | (strictly fewer dials) |
| Idle ≥ `ConnIdlePingThreshold` | handed out unverified | one `Ping` round trip on acquire | raise threshold or set negative |
| Socket probe frequency | every acquire + every Put + every std op | at most once per second per connection | (strictly fewer syscalls) |

Context cancellation mid-query still closes the connection by design (the response is not
drained), and this is asserted in the regression tests.

### Testing

* New regression suite `tests/conn_check_churn_test.go`: counts physical dials via a
  custom `DialContext` and probes idle sockets with `MSG_PEEK` between operations. Asserts
  exactly **1 dial** for sequential query/exec/batch/server-logs/early-close/timeout-ctx
  workloads (plain TCP and TLS), including 50 consecutive failing queries and mid-stream
  exceptions (the probe also proves the socket is byte-clean at the exception boundary,
  validating reuse safety). A blackhole-proxy test simulates a silently dropped flow and
  verifies the idle ping discards the half-open connection and recovers within the dial
  timeout instead of hanging.
* Full `tests/`, `tests/std`, and `tests/issues` suites pass against ClickHouse 26.6;
  root unit tests pass with `-race`; cross-compiles clean for all `GOOS` targets covered
  by the conn-check build tags (the `Recvfrom` call matters on solaris/illumos).

## Checklist
- [X] For significant changes, documentation in https://github.com/ClickHouse/clickhouse-docs was updated with further explanations or tutorials

As always, let me know if you have any questions or want to split this into multiple PRs. Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,6 +196,7 @@ The following connection settings are available in both DSN strings and the `cli
 * **max_open_conns** - Maximum number of open connections to the database (default: MaxIdleConns + 5)
 * **max_idle_conns** - Maximum number of idle connections in the pool (default: 5)
 * **conn_max_lifetime** - Maximum amount of time a connection may be reused (default: 1h)
+* **conn_idle_ping_threshold** - How long a pooled connection may sit idle before it is verified with a protocol ping when acquired. Detects half-open connections (e.g. silently dropped by a load balancer or firewall) that a socket-level check cannot see. Connections used more recently than this skip the ping entirely (default: 1m; a negative value disables idle pings)
 
 ### Connection Strategy
 * **connection_open_strategy** - Strategy for selecting servers from the connection pool:
--- a/clickhouse.go
+++ b/clickhouse.go
@@ -90,6 +90,12 @@ type nativeTransport interface {
 	asyncInsert(ctx context.Context, query string, wait bool, args ...any) error
 	ping(context.Context) error
 	isBad() bool
+	// markUnverified requires a successful revalidateIdle before the
+	// connection is handed out again.
+	markUnverified()
+	// revalidateIdle verifies liveness with a protocol ping when the
+	// connection is unverified or has idled past ConnIdlePingThreshold.
+	revalidateIdle(ctx context.Context) error
 	connID() int
 	connectedAtTime() time.Time
 	isReleased() bool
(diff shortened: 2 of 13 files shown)


must108 (changes requested):
LGTM, just address agent review. thanks!

SpencerTorres:
Outdated, closing

PR #1920, opened 2026-07-16
chore: add debug visibility on bad connection pool

kavirajk:
## Summary

Currently, if for some reason, the connection from pool is bad and client end up dialing (establishing) new connection instead, the only visibility in debug log is either "connection acquired from pool" or "new connection is established".

But this lacks "why" the connection was bad that caused to create new connection.

This changes makes sure those reasons are part of debug log

This is built on top of v2.47.0 tag instead of from `main`. DO NOT MERGE THIS as is, this is to help some ClickHouse customer to experiment with to track some performance bugs

Related to #1915 

## Checklist

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -183,20 +183,21 @@ func (c *connect) settings(querySettings Settings) []proto.Setting {
 	return settings
 }
 
-func (c *connect) isBad() bool {
+func (c *connect) healthCheck() error {
 	if c.isClosed() {
-		return true
+		return ErrConnectionClosed
 	}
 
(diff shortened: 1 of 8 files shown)


conn.go
```diff
@@ -183,20 +183,21 @@ func (c *connect) settings(querySettings Settings) []proto.Setting {
 	return settings
 }
 
-func (c *connect) isBad() bool {
+func (c *connect) healthCheck() error {
```
kavirajk, line 186: This is the main change that adds more visibility on debug log on "why connection is not re-used from the pool"

kavirajk:
replaced by #1973

PR #1928, opened 2026-07-20
feat: support arbitrary input/output formats (experimental, HTTP only)

kavirajk:
### Changes:

Two new methods on `driver.Conn`, marked experimental:

```
// stream query results in any ClickHouse format
QueryFormat(ctx, "Parquet", "SELECT * FROM events") (io.ReadCloser, error)

// insert pre-formatted data from any io.Reader
InsertFormat(ctx, "CSV", "INSERT INTO events", file) error
```

Today the driver forces everything through the Native format: it overrides default_format on HTTP and rewrites any `FORMAT` clause in `INSERT` statements. That means you can't ask ClickHouse for `CSV`, `JSONEachRow`, `Parquet`, etc., even though the server over HTTP happily produces and parses all ~70 formats by itself.

With this PR the driver just gets out of the way: the server does all encoding/decoding, and the client streams raw bytes. Query results come back as an `io.ReadCloser` you read the formatted bytes from. 
inserts take an `io.Reader` (a file, an HTTP body, a `strings.Reader`) as payload. No new dependencies.

The immediate use case is Parquet — exporting query results as Parquet files and ingesting Parquet files directly, without going through Rows/Batch.

### Why HTTP only?

Over the native TCP protocol the server always speaks Native blocks, so any other format would have to be encoded/decoded client-side (this is what clickhouse-client does internally). 

I prototyped that with pluggable codecs (as separate package) with CSV/JSONEachRow/Parquet implementations — but it means shipping and forever maintaining format encoders in Go, plus heavy dependencies (arrow-go pulls in grpc, thrift, flatbuffers). 

We deliberately dropped it from this release; the direction we want to explore instead is server-side formatting over the native protocol. 

Over native TCP, both methods return a clear sentinel error (`ErrFormatNativeUnsupported`) pointing users at HTTP.

## Checklist


lib/driver/driver.go
```diff
@@ -46,6 +47,32 @@ type (
...
+		// byte stream. The caller must Close the returned stream; until then it
+		// holds a connection (visible in Stats). Put the format in the format
+		// argument, not as a FORMAT clause in the query.
+		//
+		// Experimental: this API is experimental and may change or be removed
+		// in a future minor release. It is currently only supported over the
+		// HTTP protocol, where the server encodes the stream and every
+		// server-supported format works; over the native protocol it returns
+		// clickhouse.ErrFormatNativeUnsupported.
+		QueryFormat(ctx context.Context, format string, query string, args ...any) (io.ReadCloser, error)
```
chernser, line 102: minor: feels like some glue word missing like `QueryAsFormat`, `InsertAsFormat`.

conn_http_format.go
```diff
@@ -0,0 +1,231 @@
...
+			}
+		}
+	}
+}
+
+// safeLen returns how many pending bytes can be served without risking that
+// their tail is the beginning of an exception marker split across reads.
+func (r *exceptionScanReader) safeLen() int {
+	if r.srcDone || r.err != nil {
+		return len(r.pending)
```
chernser, line 162: this is the branch for done or error. Then how would length be used in this case ?Should it return `0` ?

conn_http_format.go
```diff
@@ -0,0 +1,231 @@
...
+		err = fmt.Errorf("bindQueryOrAppendParameters: %w", err)
+		release(h, err)
+		return nil, err
+	}
+
+	headers := make(map[string]string)
+	switch h.compression {
+	case CompressionGZIP, CompressionDeflate, CompressionBrotli:
+		headers["Accept-Encoding"] = h.compression.String()
+	case CompressionZSTD, CompressionLZ4:
```
chernser, line 276: lz4 and zstd can be used in http compression. 
in java we have dedicated flag `http_compression` to understand what compression should be used.

conn_http_format.go
```diff
@@ -0,0 +1,231 @@
...
+	q.Set("default_format", formatName)
+	req.URL.RawQuery = q.Encode()
+
+	res, err := h.executeRequest(req) //nolint:bodyclose // closed via httpFormatStream.Close
+	if err != nil {
+		release(h, err)
+		return nil, err
+	}
+
+	rw := h.compressionPool.Get()
```
chernser, line 300: how do you detect compression? 
would it work with uncompressed response?

chernser (approved):

lib/driver/driver.go
kavirajk: My initial idea was `QueryWithFormat` and `InsertWithFormat`. Later decided to keep it simple. Goes well with other apis

`Query()` and `QueryFormat()`
`Insert()` and `InsertFormat()`

conn_http_format.go
kavirajk: Good point. REturning `0` here would actually lose data. Rationale being the marker might split across two reads (say if buffer ends in `__excep..`). We cannot serve those bytes yet as next `read()` might complete the marker.

conn_http_format.go
kavirajk: You are right. In Go currently `Compression{Method: CompressionZSTD}` means "native framing" on HTTP, which is why I skip it.

 Supporting HTTP-level `lz4/zstd` needs two new decoders in `HTTPReaderWriter` plus a decision on how to express "HTTP compression" vs "native compression" in the options.  

the dedicated `http_compression` flag you have in Java is probably the right choice. I will leave it up to follow up PR. It now just sends "uncompressed" and not breaks.

conn_http_format.go
kavirajk: Detection is by response header, inside the existing `HTTPReaderWriter.NewReader` (same code the regular query path uses): 

it wraps the body in a decompressor only if the response's `Content-Encoding` matches the method configured on the connection. 

If the server replies uncompressed, ignores `Accept-Encoding`, or `enable_http_compression` is off, there's no `Content-Encoding` header, so the body passes through untouched.

So yes, uncompressed responses work AFAIK (compression is opportunistic). I'll add a small test that pins this (compression configured on the connection, server responding uncompressed).

lib/driver/driver.go
```diff
@@ -46,6 +47,43 @@ type (
...
+		// all-or-nothing semantics set wait_end_of_query=1 via WithSettings:
+		// the server then buffers the complete result, failures surface as an
+		// error from QueryFormat itself, and the stream is served unscanned.
+		//
+		// Experimental: this API is experimental and may change or be removed
+		// in a future minor release. It is currently only supported over the
+		// HTTP protocol, where the server encodes the stream and every
+		// server-supported format works; over the native protocol it returns
+		// clickhouse.ErrFormatNativeUnsupported.
+		QueryFormat(ctx context.Context, format string, query string, args ...any) (io.ReadCloser, error)
```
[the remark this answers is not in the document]
kavirajk: Yes. the break is deliberate. We have changed this interfaces in the past. The main rationale is for easy maintainability and this interface acts mostly like opaque interface for consumers most of the time.

alex-clickhouse:
One lesson learned from doing this in the .NET client: make sure to consider compression in both directions from the start. For the raw arbitrary format queries, I think it's pretty common for users to want to send data that is already compressed on disk/receive compressed data.

So, for sending, let them specify `Content-Encoding` in `InsertFormat()`.

For receiving, let users specify `Accept-Encoding` so they can set an arbitrary compression format, and don't decompress the response for them unless necessary.

kavirajk:
> make sure to consider compression in both directions from the start. 

You saying we should respect the `Content-Encoding` in `INSERT` queries even though the SQL doesn't have say `FORMAT` clause?

For example
```
Content-Encoding: application/json
```

`INSERT INTO table` -> Should send query to server `INSERT INTO TABLE FORMAT JSON`?

@alex-clickhouse

alex-clickhouse:
The scenario I'm describing is eg: user has a `data.jsonl.gzip` file. They want to do `INSERT INTO TABLE FORMAT JSONEachRow`. They should not need to decompress that on the client side unless they want to. They can send the raw compressed stream to the server with `Content-Encoding: Gzip` and the server will handle decompression.

And essentially the same in the other direction, you can go straight from `SELECT ... FORMAT JSONEachRow` to `data.jsonl.gzip` with the right `Accept-Encoding`.

kavirajk:
Good point @alex-clickhouse. 

The problem is more deep and it's independent of Arbitrary format. 

Currently the driver fully controls the "compression" and expects the user to sent plaint formats at all cost. When user wants to use compression via option 

```
		Addr:     []string{"127.0.0.1:8123"},
		Protocol: clickhouse.HTTP,
		Compression: &clickhouse.Compression{
			Method: clickhouse.CompressionGZIP,
		},
	})

```

Now client "compress" and "decompress" before sending to server and sending results back to user. 

This works usually well. Except the case you pointed out "What if user directly want to send already compressed (gzip) format (csv, parquet)". 

That use case needs additional work independent of the arbitrary format to design how to transparently let user send compressed files. UX needs some thinking before I do that.

[I created an issue to track](https://github.com/ClickHouse/clickhouse-go/issues/1936). Would be a better follow up PR.

PR #1934, opened 2026-07-28
fix: close connection when dial setup fails to avoid a file descriptor leak

knQzx:
## Summary

`dial` opens the TCP connection first, then runs the compression / JWT / handshake / addendum setup. if any of those steps fails it returns the error but never closes the socket, so the fd leaks on every failed connection attempt - a reconnect storm against a wrong password, a TLS failure or an unreachable server eventually exhausts the process file descriptors and no new sockets can be opened

reproduce: point a client at an endpoint that accepts the TCP connection but fails the handshake (or just use a wrong password), open connections in a loop and watch the open fd count climb without bound. the added `TestDialClosesConnectionOnSetupFailure` injects a connection whose handshake fails and asserts `dial` closed it - it fails on `main` and passes with this change

the fix closes the freshly dialed connection on any error after the dial succeeds, with a deferred conditional close on the named error return

## Checklist

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -22,11 +22,8 @@ import (
 	"github.com/ClickHouse/clickhouse-go/v2/lib/proto"
 )
 
-func dial(ctx context.Context, addr string, num int, opt *Options) (*connect, error) {
-	var (
-		err  error
-		conn net.Conn
-	)
+func dial(ctx context.Context, addr string, num int, opt *Options) (c *connect, err error) {
+	var conn net.Conn
 
 	switch {
 	case opt.DialContext != nil:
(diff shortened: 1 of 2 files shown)


must108 (changes requested):
@knQzx looks good, just address agent review.

conn.go
```diff
@@ -44,6 +41,15 @@ func dial(ctx context.Context, addr string, num int, opt *Options) (*connect, er
 		return nil, err
 	}
 
+	// Close the freshly dialed connection if any later setup step (compression,
```
must108, line 44: Reduce size of comment

conn_fd_leak_test.go
```diff
@@ -0,0 +1,42 @@
...
+	closed *atomic.Bool
+}
+
+func (c *closeTrackingConn) Close() error {
+	c.closed.Store(true)
+
+	return c.Conn.Close()
+}
+
+// TestDialClosesConnectionOnSetupFailure ensures dial does not leak the underlying
```
must108, line 24: Reduce size of comment

must108 (changes requested):

knQzx:
shortened both comments, and folded the two close paths into one: the defer now sits above the dial error check and guards `conn != nil`, so the case the agent review flagged (a custom `DialContext` returning a connection together with an error) is handled by the same code that handles a failing setup step, with a single close site

the test is a table over both cases now and counts the closes, so it also pins that the connection is closed exactly once

rebased on main, the interserver-secret change landed inside `dial` in the meantime and the named return makes the guard cover its new return paths too

PR #1938, opened 2026-07-29
fix: query parameter detection in quoted SQL

fallintoplace:
## Summary

Server-side query parameter detection currently applies a regular expression to the complete SQL query. Parameter-like text inside a string, quoted identifier, or comment can therefore select the native parameter path and reject ordinary positional arguments with "unsupported query parameter type".

Detect parameter syntax only outside quoted and commented contexts by reusing the existing binding scanner. The detection also requires a nonempty parameter name and type. Regression tests cover strings, quoted identifiers, line and block comments, and a real native parameter with a quoted Enum type.

## Checklist

## Testing

- go test ./...

### DIFF
--- a/query_parameters.go
+++ b/query_parameters.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
-	"regexp"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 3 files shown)


must108 (commented):
@fallintoplace Fix agent comments

PR #1941, opened 2026-07-30
fix: parse Map key types containing commas

fallintoplace:
## Summary

Allow `Map` key types to contain commas in nested parameters. For example, the driver now parses this valid type correctly:

```sql
Map(DateTime64(3, 'UTC'), String)
```

## Problem

`Map.parse` split key and value types at the first comma. For the example above, it treated `DateTime64(3` as the key type and failed before any query could run. The existing Enum-specific workaround covered only one form of nested comma.

## Fix

Find the first top-level comma while ignoring nested parentheses and quoted content. This handles parameterized key types generally, preserves Enum labels containing parentheses, and removes the Enum special case. The parsing flow was also flattened so invalid and non-comparable key types return early.

## Tests

Unit coverage includes:

- `DateTime64` keys with timezone parameters
- Enum keys, including a label containing `)`
- simple Map types
- nested value types
- malformed Map definitions

```bash
go test ./...
```

## Checklist

- [x] Unit tests covering the regression were added


must108 (changes requested):
Overall, LGTM. Please just resolve the nits identified by the agent @fallintoplace. Thanks!

must108 (approved):
LGTM

PR #1949, opened 2026-08-03
perf: make `safeLen()` linear runtime in exception parser

### LINKED ISSUE #1940: QueryFormat: exceptionScanReader re-scans its whole residual buffer on every Read, collapsing small-read throughput to ~3.4 MB/s
## Description

`Conn.QueryFormat` returns an `io.ReadCloser` that, on the default path, **is** `*exceptionScanReader` (`conn_http_format.go:322-333`) — unbuffered. That reader buffers up to `exceptionScanLimit` (32 KiB) of upstream data in `r.pending`, and on **every** `Read` call it invokes `safeLen()`, which runs `bytes.Index` across the entire undecided residual:

```go
// conn_http_format.go:160
func (r *exceptionScanReader) safeLen() int {
	if r.srcDone || r.err != nil {
		return len(r.pending)
	}
	tail := r.pending[r.scanFrom:]
	if i := bytes.Index(tail, exceptionFrame); i >= 0 { // <- O(residual), every Read
		return r.scanFrom + i
	}
	...
	for k := holdback; k > 0; k-- {                     // <- plus the split-marker probe
		if bytes.Equal(tail[len(tail)-k:], exceptionFrame[:k]) {
```

In the common case (no `__exception__` bytes in the payload) `scan()` returns without advancing `scanFrom`, so `scanFrom` stays `0` and `tail` is the **whole** residual on every call. The result is O(residual) work per `Read` regardless of how few bytes the caller asked for, i.e. an amplification factor of roughly `residual / (2 * readSize)`.

The byte-consumption side is already fine — `r.pending = r.pending[n:]` (`conn_http_format.go:89`) is an O(1) reslice, and the `append` refill amortises. `safeLen` is the only per-read cost that scales with the buffer.

### Impact

Small reads against the returned stream are ~1000x slower per byte than large ones. Measured below: `Read` with a 1-byte buffer drains at a flat **3.4 MB/s** — about **300 ns per byte** — while 4 KiB reads reach ~4.8 GB/s. A consumer that pulls a large `TabSeparated`/`RowBinary` result in small steps (`io.ReadFull` for fixed-width `RowBinary` fields, `binary.Read` directly on the stream, or any hand-rolled byte-at-a-time framing) spends essentially all its time in `bytes.Index`, not on the network. A 100 MB result read this way costs ~30 s of pure CPU.

Callers who wrap the stream in `bufio.Reader`/`bufio.Scanner` are largely insulated (4 KiB reads still pay ~8x, but at GB/s that is not a practical bottleneck). The defect is that the returned reader has no internal buffering of its own and is pathological when used directly, which the API does not signal.

## ClickHouse server version

Not applicable / not verified against a running server — the reproduction is a pure in-process exercise of the reader with a `bytes.Reader` as the upstream, and no server round-trip is involved. (The ClickHouse server in my environment was unreachable, so I did not additionally confirm end-to-end through `QueryFormat`; the reader under test is the exact value `queryFormat` returns.)

## Reproduction

`scan_reader_cost_test.go` in the root `clickhouse` package. It drains the same payload twice at each read size — once through a bare `bytes.Reader` (control) and once through `newExceptionScanReader` (the subject) — and reports the ratio.

```go
package clickhouse

import (
	"bytes"
	"fmt"
	"io"
	"testing"
	"time"
)

func drain(r io.Reader, readSize int) (int, time.Duration) {
	p := make([]byte, readSize)
	total := 0
	start := time.Now()
	for {
		n, err := r.Read(p)
		total += n
		if err != nil {
			break
		}
	}
	return total, time.Since(start)
}

func TestScanReaderSmallReadCost(t *testing.T) {
	for _, payload := range []int{1 << 20, 4 << 20} {
		data := bytes.Repeat([]byte("0123456789abcdef"), payload/16)
		for _, readSize := range []int{1, 1024, 4096, 65536} {
			ctrlN, ctrlEl := drain(bytes.NewReader(data), readSize)
			subjN, subjEl := drain(newExceptionScanReader(bytes.NewReader(data), "tag123"), readSize)
			if ctrlN != len(data) || subjN != len(data) {
				t.Fatalf("short read: ctrl=%d subj=%d want=%d", ctrlN, subjN, len(data))
			}
			fmt.Printf("payload=%8d readSize=%6d  control=%12v (%8.1f MB/s)  scanReader=%12v (%8.1f MB/s)  slowdown=%7.1fx\n",
				len(data), readSize,
				ctrlEl, float64(ctrlN)/ctrlEl.Seconds()/1e6,
				subjEl, float64(subjN)/subjEl.Seconds()/1e6,
				subjEl.Seconds()/ctrlEl.Seconds())
		}
	}
}
```

`go test -run TestScanReaderSmallReadCost -v .`

**Actual:**

```
payload= 1048576 readSize=     1  control=  6.028663ms (   173.9 MB/s)  scanReader=306.505615ms (     3.4 MB/s)  slowdown=   50.8x
payload= 1048576 readSize=  1024  control=    78.497µs ( 13358.2 MB/s)  scanReader=   1.59057ms (   659.2 MB/s)  slowdown=   20.3x
payload= 1048576 readSize=  4096  control=    27.301µs ( 38408.0 MB/s)  scanReader=   265.717µs (  3946.2 MB/s)  slowdown=    9.7x
payload= 1048576 readSize= 65536  control=    31.139µs ( 33674.0 MB/s)  scanReader=  903.398µs (  1160.7 MB/s)  slowdown=   29.0x
payload= 4194304 readSize=     1  control= 19.519773ms (   214.9 MB/s)  scanReader=1.222166775s (     3.4 MB/s)  slowdown=   62.6x
payload= 4194304 readSize=  1024  control=   275.576µs ( 15220.1 MB/s)  scanReader=   2.39713ms (  1749.7 MB/s)  slowdown=    8.7x
payload= 4194304 readSize=  4096  control=   110.657µs ( 37903.6 MB/s)  scanReader=   869.908µs (  4821.5 MB/s)  slowdown=    7.9x
payload= 4194304 readSize= 65536  control=  1.084807ms (  3866.4 MB/s)  scanReader=  1.084807ms (  3866.4 MB/s)  slowdown=    8.0x
```

**Expected:** per-byte cost independent of `readSize`. The `1`-byte column should not be ~1000x worse per byte than the `4096`-byte column, and the reader's absolute throughput should not be pinned at 3.4 MB/s. Note the 3.4 MB/s figure is identical at 1 MiB and 4 MiB payloads — the cost is a fixed ~300 ns/byte set by the 32 KiB buffer size, not a growth in total work.

Only the timing differs; the bytes returned are byte-for-byte correct in every configuration (the test asserts full-length delivery), so this is purely a performance defect.

## Suggested fix

Two independent options, either of which removes the amplification:

1. **Memoise the scan in `safeLen`.** The `bytes.Index` result only changes when new bytes arrive from upstream. Cache the computed safe length after each refill and decrement it alongside the `r.pending = r.pending[n:]` reslice in `Read` (`conn_http_format.go:85-116`), rather than recomputing from scratch. `scan()` already runs exactly once per refill and is the natural place to record it.

2. **Serve from the residual without re-deciding.** Track a `safeUntil` offset updated in `scan()`/on refill, and have `Read` hand out `min(len(p), safeUntil)` bytes directly.

Buffering the reader before returning it from `queryFormat` would hide the worst case but leaves the per-`Read` O(residual) scan in place for anyone reading with a buffer smaller than the residual, so it is a mitigation rather than a fix.

## Link

Found while checking whether ClickHouse/clickhouse-connect#917 (chDB `raw_stream` re-copying its residual via `del self._buf[:size]` on every read) applied here. The specific mechanism in that report does **not** apply to clickhouse-go — the consumption side already uses the offset/reslice approach that issue recommends as the fix. This is the same "re-do O(residual) work per read" class arriving through a different path: the exception-marker scan instead of the buffer compaction.

Introduced with the format stream support in #1928.

kavirajk:
## Summary

1. The `read()` loop try to decide if the bytes scanned so far is safe to return to caller (making sure no legit exception is in there)
2. To make this decision, we have `safeLen()` api. But this becomes O(n^2) by not incrementing scanOffset correctly.
3. This PR updates it during `scan()` api that makes the `safeLen` linear runtime.

### Before
screenshot5

### After
screenshot4

Fixes: #1940 

## Checklist

### DIFF
--- a/conn_http_format.go
+++ b/conn_http_format.go
@@ -54,9 +54,8 @@ type exceptionScanReader struct {
 	// terminal error: parsed exception or upstream read error
 	err error
 
-	// srcDone make sure if upstream reader has anything left.
-	// Say any EOF or error from upstream makes this true. Meaning
-	// no more input is coming.
+	// srcDone is true once the upstream reader is exhausted (EOF or error).
+	// no more input is coming
 	srcDone bool
 }
 
@@ -137,8 +136,10 @@ func (r *exceptionScanReader) scan() {
 	for {
 		i := bytes.Index(r.pending[r.scanOffset:], exceptionFrame)
 		if i < 0 {
-			// instead of safeLen() to scale "whole" pending buffer, we can make
-			// safeLen() make it check only new reads.
+			// No complete frame exists at or after scanOffset, so every byte
+			// except the split-frame holdback tail is decided data; advancing
+			// scanOffset keeps safeLen bounded per Read instead of rescanning
+			// all of pending
 			if cleared := len(r.pending) - (len(exceptionFrame) - 1); cleared > r.scanOffset {
 				r.scanOffset = cleared
 			}
--- a/conn_http_format.go
+++ b/conn_http_format.go
@@ -31,19 +31,33 @@ var exceptionFrame = []byte(exceptionMarker + "\r\n")
 //
 // A frame candidate is treated as a genuine exception only if the marker is
 // followed by the per-response random tag the server announced in the
-// X-ClickHouse-Exception-Tag header (tag) - that is what the tag exists for:
-// result data cannot contain a value generated after the response started, so
-// data that happens to contain the marker bytes is served verbatim. On older
-// servers that send no tag header, tag is empty and the marker alone is
-// trusted (best-effort).
+// X-ClickHouse-Exception-Tag HTTP header (tag). On older servers that send no tag header,
+// tag is empty and the marker alone is trusted (best-effort).
 type exceptionScanReader struct {
-	src      io.Reader
-	tag      string // from X-ClickHouse-Exception-Tag; empty disables validation
-	buf      []byte
-	pending  []byte
-	scanFrom int   // pending[:scanFrom] holds no undecided frame candidates
-	err      error // terminal error: parsed exception or upstream read error
-	srcDone  bool
+	// src is the upstream reader
+	src io.Reader
+
+	// tag is from X-ClickHouse-Exception-Tag HTTP header;
+	// empty disables validation (older version)
+	tag string
+
+	// buf is the scratch buffer (max exceptionScanLimit)
+	buf []byte
+
+	// pending is every byte that is not returned to the caller.
+	// it may or may not contain exception.
+	pending []byte
+
+	// pending[:scanOffset] holds plain text and are not exception candidates.
+	scanOffset int
+
+	// terminal error: parsed exception or upstream read error
+	err error
+
+	// srcDone make sure if upstream reader has anything left.
+	// Say any EOF or error from upstream makes this true. Meaning
+	// no more input is coming.
+	srcDone bool
 }
 
 func newExceptionScanReader(src io.Reader, tag string) *exceptionScanReader {
@@ -87,8 +101,8 @@ func (r *exceptionScanReader) Read(p []byte) (int, error) {
 		if safe := r.safeLen(); safe > 0 {
 			n := copy(p, r.pending[:safe])
 			r.pending = r.pending[n:]
-			if r.scanFrom -= n; r.scanFrom < 0 {
-				r.scanFrom = 0
+			if r.scanOffset -= n; r.scanOffset < 0 {
+				r.scanOffset = 0
 			}
 			return n, nil
 		}
@@ -121,11 +135,16 @@ func (r *exceptionScanReader) Read(p []byte) (int, error) {
 // of the stream) settles the question.
 func (r *exceptionScanReader) scan() {
 	for {
-		i := bytes.Index(r.pending[r.scanFrom:], exceptionFrame)
+		i := bytes.Index(r.pending[r.scanOffset:], exceptionFrame)
 		if i < 0 {
+			// instead of safeLen() to scale "whole" pending buffer, we can make
+			// safeLen() make it check only new reads.
+			if cleared := len(r.pending) - (len(exceptionFrame) - 1); cleared > r.scanOffset {
+				r.scanOffset = cleared
+			}
 			return
 		}
-		i += r.scanFrom
+		i += r.scanOffset
 		if r.tag == "" {
 			// Old servers (<= 25.8) announce no tag: the marker alone has to
 			// be trusted.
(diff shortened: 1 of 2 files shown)


conn_http_format.go
```diff
@@ -121,11 +135,16 @@ func (r *exceptionScanReader) Read(p []byte) (int, error) {
 // of the stream) settles the question.
 func (r *exceptionScanReader) scan() {
 	for {
-		i := bytes.Index(r.pending[r.scanFrom:], exceptionFrame)
+		i := bytes.Index(r.pending[r.scanOffset:], exceptionFrame)
 		if i < 0 {
+			// instead of safeLen() to scale "whole" pending buffer, we can make
+			// safeLen() make it check only new reads.
+			if cleared := len(r.pending) - (len(exceptionFrame) - 1); cleared > r.scanOffset {
```
kavirajk, line 143: This is the main change.

Say you have a buffer of bytes from upstream reader (that may or may not contain `__exception__`)

```ascii
buf: [.......<pending>|<scanOffset>........] - len(buf)

Anything from [0..pending], -> plain text guranteed.
Anything from [scanOffset...len(buf)] -> suspicious candidate for exception body.
```

The purpose of `scan()` api is to figure out the exception in `[scanOffset..len(buf)]`. Previously, `safeLen()` api checks for whole buffer instead of only newly appened bytes to buf. That would take quadratic type everytime time.

This PR made it linear time

conn_http_format.go
alex-clickhouse: Leftover scanFrom mention in the comment

alex-clickhouse: I think you can advance `scanOffset` to `i` here(?) Otherwise it's the same issue the PR fixes (although the case is quite rare, with the tag at the boundary).

conn_http_format_bench_test.go
```diff
@@ -0,0 +1,81 @@
...
+	r.scan()
+	require.Equal(t, len(r.pending)-(len(exceptionFrame)-1), r.scanOffset,
+		"clean fill: scanOffset must advance to the holdback boundary")
+
+	// An undecided candidate (complete frame, tag bytes still in flight) must
+	// stay at or right of scanOffset so safeLen keeps holding it back.
+	r = newExceptionScanReader(bytes.NewReader(nil), "tag123")
+	r.pending = append(bytes.Repeat([]byte("x"), 100), exceptionFrame...)
+	r.scan()
+	require.LessOrEqual(t, r.scanOffset, 100,
```
alex-clickhouse: With the change above, this should also be .Equal(), like the first half.

alex-clickhouse (approved):

conn_http_format.go
kavirajk: good catch :+1: addressed

PR #1975, opened 2026-08-11
ci: harden the cloud test path and pin the remaining floating actions

thomasboni:
## Summary

Five focused changes to the workflows, each in its own commit. No test or documentation impact.

**1. The cloud test job runs pull request code with a write-all token** (`run-tests-cloud.yml`)

`claude-code-review.yml` already carries this comment:

> SECURITY: pull_request_target checks out the base branch by default. Do NOT add `ref: ${{ github.event.pull_request.head.sha }}` here, that would run untrusted fork code with access to repository secrets.

`run-tests-cloud.yml` does that. Under `pull_request_target` it checks out the head SHA, and the workflow grants `write-all`, so the job that executes fork code holds a full write token alongside the cloud credentials, on a self-hosted runner.

The label gate is the right control and stays untouched. What changes is what the opted-in code can reach: the workflow default drops to `contents: read`, the one step that needed a write scope (removing the trigger label) moves to its own job on a hosted runner that never checks out the PR, and the checkout stops persisting the git credential. Behaviour is unchanged, the label is still removed after every run including failures.

This does not remove the underlying `pull_request_target` plus head-checkout pattern, because that is what makes cloud testing of a PR work at all. Splitting the credential-bearing part from the fork-code part would be the full fix, and it is your architectural call rather than something to slip into this PR. Happy to open a follow-up if you want it.

**2. Three more workflows default to `write-all`** (`run-tests.yml`, `run-tests-head.yml`, `release.yml`)

None of them uses `GITHUB_TOKEN` for anything that writes. The two test workflows check out and run tests with no API calls, and `release.yml` mints a GitHub App token and uses it for the checkout, the release API call and the push, so the workflow token is unused. All three become `contents: read`.

**3. Five action references were not pinned**

`lycheeverse/lychee-action@v2`, `anthropics/claude-code-action@v1`, `golangci/golangci-lint-action@v9` and two `actions/checkout@main` references now pin a commit with the version in a trailing comment. Two are worth calling out: the CodeQL steps were on `v1`, a tag that last moved in January 2023, so the analysis has been running a three-year-old revision of the action. They move to v4.37.6. And `actions/checkout@main` tracks a moving branch rather than a release.

Left alone deliberately: the two `ClickHouse/integrations-shared-workflows` references. They are your own shared workflows and pinning them would stop central updates reaching this repo, which is presumably the point. Same reasoning for the `clickhouse/docs-builder` image. Happy to pin either if you would rather have them fixed.

**4. The link checker had no `permissions:` block**, so its token fell back to the repository default. It checks out and runs a link checker with no API calls, so `contents: read`.

**5. The benchmark job passed `secrets: inherit`** (`stresshouse-benchmark-compare.yml`), handing the called workflow every secret in the repository. `clickhouse-benchmark.yml` declares exactly two, `WORKFLOW_AUTH_PUBLIC_APP_ID` and `WORKFLOW_AUTH_PUBLIC_PRIVATE_KEY`. Passing those explicitly keeps the cloud credentials out of a workflow that does not ask for them, and means a secret added later is not forwarded by default.

One thing worth knowing if you use a repository-level Actions allowlist: entries written as tag patterns (`owner/action@v1`) stop matching once a workflow pins a SHA, and every workflow using that action then fails at startup. The fix is `owner/action@*` in settings.

Found and fixed by [Plumber](https://github.com/getplumber/plumber)'s analysis, reviewed and submitted by me.


.github/workflows/stresshouse-benchmark-compare.yml
```diff
@@ -41,4 +41,6 @@ jobs:
       client: clickhouse-go
       profile: pr-compare
       pr_number: ${{ github.event.issue.number || inputs.pr_number }}
-    secrets: inherit
+    secrets:
```
kavirajk, line 44: why not inherit the secrets here?

.github/workflows/run-tests-cloud.yml
```diff
@@ -53,8 +57,17 @@ jobs:
         run: |
           CLICKHOUSE_DIAL_TIMEOUT=20 CLICKHOUSE_TEST_TIMEOUT=1200s CLICKHOUSE_QUORUM_INSERT=2 make test
 
+  # Dropping the label needs a write scope, so it runs in its own job on a clean
+  # runner that never checks out the pull request. The job above keeps read-only
+  # access even though it executes fork code.
+  remove-label:
```
kavirajk, line 63: 👍

.github/workflows/run-lint.yml
```diff
@@ -24,6 +24,6 @@ jobs:
           go-version: "1.25"
 
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@v9
+        uses: golangci/golangci-lint-action@ba0d7d2ec06a0ea1cb5fa41b2e4a3ab91d21278a # v9.3.0
```
kavirajk: I'm not super comfortable with git commit hash here. Can't we just do @v9.3.0 here? (similar in other places?)

kavirajk (commented):
Thanks @thomasboni for the PR.

Two things
1. Left few comments to clarify
2. Can you sign the CLA?

.github/workflows/stresshouse-benchmark-compare.yml
thomasboni: `secrets: inherit` forwards every secret into the workflow, while the callee here only declares `WORKFLOW_AUTH_PUBLIC_APP_ID` and `WORKFLOW_AUTH_PUBLIC_PRIVATE_KEY`, so everything beyond those two is exposure with no benefit. 

And this particular callee lives in a different repo (`ClickHouse/integrations-shared-workflows`) and is referenced on `@main`, meaning any future change to that workflow immediately runs here with whatever secrets we hand it.

Passing them explicitly caps what a change on that side can ever reach, and doubles as documentation of what the workflow actually needs. If the shared workflow grows a new secret later, the failure is loud (a `missing-secret` error) instead of a silent broadening of access.

.github/workflows/run-lint.yml
thomasboni: In most cases the git commit hash is the only way to be sure the action you are using is immutable. It's highly recommended when the action comes from an external source that could be compromised and run malicious code in your pipeline (which has access to secrets, code, envs, etc.). That's the route the `tj-actions/changed-files` compromise took.

But in this specific case with `golangci/golangci-lint-action` you are right to prefer `@v9.3.0`, because the repo uses GitHub's immutable releases. The v9.3.0 release is marked immutable, meaning the tag can never be repointed or deleted. So `@v9.3.0` gives the same guarantee as the SHA, and I'm going to switch that one back. Same for `codeql-action`, its releases are immutable too.

The others are different:
- `lychee-action@v2.9.0`: mutable tag from a third-party org. I'd keep the SHA here, but if you prefer the tag for readability I can revert it.
- `claude-code-action@v1.0.190`: mutable tag, maintained by Anthropic. Larger vendor, but still a third party with a repointable tag. Your call, tell me which way you want it.
- `actions/checkout@v5.1.0`: mutable tag, maintained by GitHub itself. If you trust GitHub the tag is reasonable, I will revert to `@v5.1.0` (tell me if not).

.github/workflows/run-tests-cloud.yml
```diff
@@ -53,8 +57,17 @@ jobs:
...
+    needs: [test]
+    if: always() && github.event_name == 'pull_request_target'
```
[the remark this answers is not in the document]
thomasboni: Right, `always()` was meant to guarantee cleanup when the test job fails but as written it also fires for unrelated labels. Restored the label guard on the job condition.

.github/workflows/run-tests-cloud.yml
```diff
@@ -35,6 +36,9 @@ jobs:
         uses: actions/checkout@v4
```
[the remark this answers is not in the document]
thomasboni: I disagree with the bot here: even if it doesn't look consistent, we should keep the easy human-readable tag except for risky inclusion where we have to use the SHA.

thomasboni:
Hi @kavirajk 

- CLA is signed
- Let me know if my answers are clear :)

.github/workflows/check-pr-links.yml
```diff
@@ -5,14 +5,17 @@ on:
     branches: [main]
   workflow_dispatch:
 
+permissions:
```
kavirajk, line 8: thanks :) ❤️

.github/workflows/claude-code-review.yml
```diff
@@ -47,7 +47,7 @@ jobs:
             > existing-threads.json || echo '{"threads":[]}' > existing-threads.json
 
       - name: Review PR with Claude (produce findings JSON)
-        uses: anthropics/claude-code-action@v1
+        uses: anthropics/claude-code-action@5ef2e550a465a721f4f45e4a7d3c340c873e1dcc # v1.0.190
```
kavirajk, line 50: I'm failing to understand the exactly rationale for SHA based versioning for `claude-code-action` and `lyche` from the PR description.

Can you help me understand @thomasboni ? This is my only concern with the PR. After than we can merge

PR #1978, opened 2026-08-12
fix(bind): treat []byte as String and time.Duration as Time

### LINKED ISSUE #1942: bind: []byte is bound as Array(UInt8) and time.Duration as a Go duration string, so binary and Time values cannot be bound as parameters

sankalpsthakur:
## Summary
Fixes client-side parameter binding for two Go types that previously produced unusable SQL literals in `formatValue` (`bind.go`):

1. `[]byte`: was `Array(UInt8)`. Now a quoted String literal (same escaping as string).
2. `time.Duration` / `*time.Duration`: was Go duration form. Now Time/Time64-style `HH:MM:SS[.frac]`, quoted like `formatTime`.

This replaces #1945, which I had to close after an accidental rewrite of fork history. Same branch, same change.

## Validation
- `go test -run 'TestBindDuration|TestBindBytes|TestFormatDuration|TestBind' .`

Fixes #1942

### AI/LLM disclosure
- AI coding tools (including Grok and/or Codex agent-assisted editing) were used to help draft or modify code and this PR description.
- I reviewed the complete change, understand the reasoning, and ran the reported local tests before submitting.
- This submission is original work of authorship under the project CLA / contributor terms; AI output was not pasted unreviewed.


CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+# Unreleased
```
kavirajk: Can you remove the changelog entries fully?

We auto-generate changelog notes before the release (from PR title and description).

sankalpsthakur: Removed in f92ff6f.

tests/issues/issue_1942_test.go
```diff
@@ -0,0 +1,92 @@
...
+					require.NoError(t, conn.QueryRow(timeCtx, "SELECT toString(CAST(? AS Time64(3)))", duration).Scan(&got))
+					assert.Equal(t, "14:30:00.123", got)
+					require.NoError(t, conn.QueryRow(timeCtx, "SELECT toString({p:Time64(3)})", clickhouse.Named("p", duration)).Scan(&got))
+					assert.Equal(t, "14:30:00.123", got)
+				}
+			})
+		}
+	})
+
+	t.Run("database/sql", func(t *testing.T) {
```
kavirajk, line 67: can you move this to `tests/std` instead? Just to stick to the format we already have.

kavirajk: Never mind. I see this is under "issues". We usually combine native and std interface there. No problem

bind.go
```diff
@@ -483,7 +483,9 @@ func formatTime(tz *time.Location, scale TimeUnit, value time.Time) (string, err
 	return fmt.Sprintf("toDateTime64('%s', %d, '%s')", value.Format(fmt.Sprintf("2006-01-02 15:04:05.%0*d", int(scale*3), 0)), int(scale*3), escapedTimezone), nil
 }
 
-var stringQuoteReplacer = strings.NewReplacer(`\`, `\\`, `'`, `\'`)
+// Escape order: backslash first so later replacements are not re-escaped.
+// NUL is written as \0 so binary String values survive client-side bind.
+var stringQuoteReplacer = strings.NewReplacer(`\`, `\\`, `'`, `\'`, "\x00", `\0`)
```
kavirajk, line 488: do we need this binary `\x00` and `\0`?. I don't see even any of the tests cover this case.

I'm bit uncomfortable having to consider the binary string for this query param. Can you help me understand your exact use case here?

query_parameters.go
```diff
@@ -54,12 +54,21 @@ func bindQueryOrAppendParameters(paramsProtocolSupport bool, options *QueryOptio
 				case *string:
 					options.parameters[p.Name] = *v
 					continue
+				case []byte:
+					options.parameters[p.Name] = string(v)
+					continue
```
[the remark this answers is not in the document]
sankalpsthakur: added the *[]byte case to the raw-value switch in query_parameters.go, mirrors the []byte one above it. pinned with a test case.

bind.go
sankalpsthakur: The use case is preserving the normal `database/sql` meaning of a top-level `[]byte` bound to ClickHouse `String`: `String` is byte-oriented and may contain NULs, so silently truncating or changing `A\x00B` would make the new byte binding incomplete. This is covered both at the formatter level (`bind_test.go`, `TestBindBytes`: `A\x00B` -> `A\0B`) and end-to-end in `tests/issues/issue_1942_test.go`, which round-trips `[]byte{65, 0, 66}` through native + database/sql, native + HTTP, and both `?` and `{p:String}` paths. The NUL escape is only used for the SQL-literal path; server-side parameters send the raw bytes.

kavirajk (approved):
thanks @sankalpsthakur

PR #1985, opened 2026-08-19
Replace github.com/google/uuid with stdlib uuid (Go 1.27)

Laotree:
- switch UUID column driver paths to the standard library uuid package
- go directive 1.25 -> 1.27, toolchain go1.27rc3
- google/uuid retained only as an indirect dep for ch-go proto boundary (proto.ColUUID is typed google/uuid.UUID; converted at the column edge)

## Checklist

### DIFF
--- a/examples/clickhouse_api/batch.go
+++ b/examples/clickhouse_api/batch.go
@@ -2,7 +2,7 @@ package clickhouse_api
 
 import (
 	"context"
-	"github.com/google/uuid"
+	"uuid"
 	"time"
 )
 
--- a/examples/clickhouse_api/context.go
+++ b/examples/clickhouse_api/context.go
@@ -8,7 +8,7 @@ import (
 
 	"github.com/ClickHouse/clickhouse-go/v2"
 	clickhouse_tests "github.com/ClickHouse/clickhouse-go/v2/tests"
-	"github.com/google/uuid"
+	"uuid"
 )
 
 func UseContext() error {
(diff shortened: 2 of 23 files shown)


must108 (commented):
ping @Laotree

Laotree:
anything I need to do?

must108:
@Laotree it is drafted, wanted you to take a look to open it if you are satisfied with the PR

Laotree:
> @Laotree it is drafted, wanted you to take a look to open it if you are satisfied with the PR

I’m considering making a larger change: moving UUID into a standalone package under pkg, and then using Go’s build mechanism to support both versions below 1.27 and 1.27+.

However, this would effectively make UUID a customized implementation. The fact that UUID was added to the Go standard library indicates that it is considered a fundamental type that should be supported natively. We wouldn’t use different builds for int or float depending on the Go version, so having to do that for UUID feels a bit odd.

I haven’t fully figured out the right approach yet, so I’ll keep this as a draft for now.

must108:
> > @Laotree it is drafted, wanted you to take a look to open it if you are satisfied with the PR
> 
> I’m considering making a larger change: moving UUID into a standalone package under pkg, and then using Go’s build mechanism to support both versions below 1.27 and 1.27+.
> 
> However, this would effectively make UUID a customized implementation. The fact that UUID was added to the Go standard library indicates that it is considered a fundamental type that should be supported natively. We wouldn’t use different builds for int or float depending on the Go version, so having to do that for UUID feels a bit odd.
> 
> I haven’t fully figured out the right approach yet, so I’ll keep this as a draft for now.

No worries then. Just wanted to check up, take your time!

PR #1989, opened 2026-08-21
Improve support of different uuid implementations

ShoshinNikita:
## Summary

Go 1.27 added a new standard `uuid` package. However, `uuid.UUID` doesn't implement `sql.Scanner` interface, so it can't be used with the ClickHouse driver. Fix it by adding support for interface `encoding.TextUnmarshaler` (which is implemented by the standard uuid type) to `column.UUID.ScanRow`.

## Checklist


kavirajk:
duplicate of #1985 ?

ShoshinNikita:
I don't think this PR is a duplicate of #1985:

- this PR adds support not only for Go 1.27's uuid, but all types that implement interface `encoding.TextUnmarshaler`
- this PR doesn't change go version in go.mod, allowing users to keep their version of Go

Also, I am not sure about `clickhouse-go`, but AFAIK, many packages try to support at least 2 last versions of Go. Considering this approach, #1985 will have to wait until Go 1.28

Laotree:
I agree that raising the minimum Go version to 1.27 is a compatibility consideration, but I don't think this makes #1985 and #1989 duplicates.

#1989 adds support for scanning into stdlib uuid.UUID while keeping the current Go version support.

#1985 is a broader migration: moving clickhouse-go from github.com/google/uuid to stdlib uuid, while retaining google/uuid only at the ch-go boundary for now.

So the key question for #1985 is whether we're ready to require Go 1.27, rather than whether it duplicates #1989.

kavirajk:
Thanks @Laotree @ShoshinNikita for clarification. 

> So the key question for https://github.com/ClickHouse/clickhouse-go/pull/1985 is whether we're ready to require Go 1.27, rather than whether it duplicates https://github.com/ClickHouse/clickhouse-go/pull/1989.

I don't think we are ready for that. We usually support two earlier versions. So probably we will be ready once we have Go 1.29.

tests/uuid_127_test.go
```diff
@@ -0,0 +1,16 @@
+//go:build go1.27
+
+package tests
+
+import (
+	"testing"
+	"uuid"
+)
+
+func TestUUIDStd(t *testing.T) {
```
kavirajk, line 10: @ShoshinNikita can we combine `TestUUIDStd` and `TestUUIDValuer` as two different cases in single Test under `uuid_test.go` (no new separate `uuid_127_test.go` file) ?

That is more intuitive and easy to maintain. Plus the helper `testUUIDImplementation` can be some internal to the test.

PR #1995, opened 2026-08-28
Do not hold pool mutex while waiting for drain goroutine

### LINKED ISSUE #1990: connPool.Close() can deadlock with runDrainPool ticker tick (mutex held across finished-wait)
## Summary

`connPool.Close()` can **deadlock permanently** against the background `runDrainPool` goroutine: `Close` holds the pool's write mutex while blocking on `<-i.finished`, while the drain goroutine — if it has just woken on its ticker and is blocked acquiring that same mutex — can only observe the close signal *after* it finishes the ticker branch, which requires the mutex `Close` is holding and will not release until `<-i.finished` returns. Circular wait; neither side makes progress.

## Location

- File: [`conn_pool.go`](https://github.com/ClickHouse/clickhouse-go/blob/0c09672ddf7252afcdc32db324a2b53a78113377/conn_pool.go)
- `Close()` (lines 117–133) and `runDrainPool()` (lines 144–160)

```go
func (i *connPool) Close() error {
	i.mu.Lock()            // (1) writer lock held for the whole function
	defer i.mu.Unlock()
	if i.closed() { return nil }
	close(i.finish)
	<-i.finished           // (2) waits for the drain goroutine to exit
	i.drainPool()
	return nil
}

func (i *connPool) runDrainPool() {
	defer func() { i.ticker.Stop(); close(i.finished) }()
	for {
		select {
		case <-i.ticker.C:
			i.mu.Lock()    // (3) can block behind (1)
			i.drainPool()
			i.mu.Unlock()
		case <-i.finish:
			return
		}
	}
}
```

## Problem

Deadlock sequence:

1. The drain goroutine wakes on `ticker.C` and calls `i.mu.Lock()` at line 153.
2. Before it acquires the mutex, `Close()` barges in and takes it (`sync.Mutex` in normal mode permits barging ahead of queued waiters).
3. `Close()` sets `closed` state via `close(i.finish)` and then blocks on `<-i.finished` (line 127) **while still holding `mu`**.
4. The drain goroutine, parked at line 153, cannot acquire the mutex because `Close()` holds it; it therefore never reaches the `case <-i.finish:` branch, so its deferred `close(i.finished)` never runs.
5. `Close()` never returns; every subsequent `Get`/`Put` also blocks on the held mutex.

The window is between the goroutine waking on the ticker and acquiring the mutex. It is narrow per event, but the ticker fires once per connection lifetime for the entire process lifetime, and Go's non-FIFO mutex behavior makes the "Close jumps the queue" interleaving realistic rather than theoretical.

## Trigger / Reproduction

Static analysis finding — not confirmed by execution; derived from the control flow above at `main` (`0c09672d`). Conceptual repro:

- Create a client with a short `ConnMaxLifetime` (to make the ticker fire frequently).
- Concurrently call `clickhouse.Close()` (or let finalizers/HTTP-server shutdown do it) in a loop.
- Eventually `Close()` hangs forever with all pool operations stuck.

A deterministic unit test can force the interleaving by injecting a hook/pause between the select wake-up and `mu.Lock()`, or by replacing the ticker channel with a controllable channel and gating `Close()` until the drain goroutine is provably blocked on `Lock` (observable via runtime stack inspection).

## Expected Behavior

Closing the pool should always terminate. The close signal must be observable regardless of where the drain goroutine currently sits, e.g.:

- take the ticker work outside the mutex conflict path: in the ticker branch, use a non-blocking try-lock or re-check `i.closed()` after acquiring the lock and bail out;
- or have `Close()` release the mutex before waiting: set closed flag under lock, unlock, `close(finish)`, wait for `finished`, then drain under lock.

## Actual Behavior

Permanent hang of `Close()` and of all subsequent pool operations when the unlucky interleaving occurs.

## Impact

Any code path shutting down a `clickhouse.OpenDB` client (server shutdown, tests, CLI tools) can freeze indefinitely with all connections leaked-busy; goroutine dumps show `Close` parked on `<i>.finished` and `runDrainPool` parked on `Lock`.

## Evidence

- Mutex hold spans the `<-i.finished` wait: lines 118–127.
- Drain goroutine's only exit check happens in `select`; the ticker branch unconditionally locks the same mutex first: lines 150–158.
- `closed()` is checked nowhere inside the ticker branch before locking.

AshSgDe29071999:
Fixes #1990

`connPool.Close` held `mu` across `<-i.finished`. `runDrainPool` acquires the same mutex in the ticker branch. If Close ran while that goroutine was already queued on `Lock`, Close waited for `finished` and the drain goroutine waited for `mu`: permanent hang, and every later `Get`/`Put` stuck behind the held lock.

Close now records `closing` and signals `finish` under the lock, releases it before waiting, then drains remaining connections after the background goroutine has exited. A second `Close` is still a no-op.

The drain loop also returns immediately if it observes `closing` after taking the lock on a ticker tick.

### DIFF
--- a/conn_pool.go
+++ b/conn_pool.go
@@ -19,6 +19,7 @@ type connPool struct {
 	ticker   *time.Ticker
 	finish   chan struct{}
 	finished chan struct{}
+	closing  bool
 
 	maxConnLifetime time.Duration
 }
@@ -116,17 +117,21 @@ func (i *connPool) Put(conn nativeTransport) {
 
 func (i *connPool) Close() error {
 	i.mu.Lock()
-	defer i.mu.Unlock()
-
-	if i.closed() {
+	if i.closed() || i.closing {
+		i.mu.Unlock()
 		return nil
 	}
-
+	i.closing = true
 	close(i.finish)
+	// Release the mutex before waiting for the drain goroutine. Holding it
+	// across <-finished deadlocks if that goroutine is blocked on Lock in
+	// the ticker branch.
+	i.mu.Unlock()
 
 	<-i.finished
 
-	// Drain all remaining connections from the pool
+	i.mu.Lock()
+	defer i.mu.Unlock()
 	i.drainPool()
 
 	return nil
@@ -151,6 +156,10 @@ func (i *connPool) runDrainPool() {
 		select {
 		case <-i.ticker.C:
 			i.mu.Lock()
+			if i.closing {
+				i.mu.Unlock()
+				return
+			}
 			i.drainPool()
 			i.mu.Unlock()
 		case <-i.finish:
--- a/conn_pool_sync_test.go
+++ b/conn_pool_sync_test.go
@@ -67,3 +67,16 @@ func TestConnPool_ExpiredConnectionsAreDrained(t *testing.T) {
 		synctest.Wait()
 	})
 }
+
+func TestConnPool_CloseDuringTickerTick(t *testing.T) {
+	synctest.Test(t, func(t *testing.T) {
+		pool := newConnPool(50*time.Millisecond, 5)
+		pool.Put(&mockTransport{connectedAt: time.Now(), id: 1})
+
+		time.Sleep(50 * time.Millisecond)
+		synctest.Wait()
+
+		require.NoError(t, pool.Close())
+		require.NoError(t, pool.Close())
+	})
+}
--- a/conn_pool_test.go
+++ b/conn_pool_test.go
@@ -529,6 +529,30 @@ func TestConnPool_FIFOOrdering(t *testing.T) {
 	}
 }
 
+func TestConnPool_CloseDoesNotDeadlockWithDrainTicker(t *testing.T) {
+	// Close used to hold mu while waiting for runDrainPool. If the drain
+	// goroutine had already woken on ticker.C and was blocked on that
+	// same mutex, neither side made progress.
+	for range 50 {
+		pool := newConnPool(time.Millisecond, 5)
+		pool.Put(&mockTransport{connectedAt: time.Now(), id: 1})
+
+		done := make(chan struct{})
+		go func() {
+			assert.NoError(t, pool.Close())
+			close(done)
+		}()
+
+		select {
+		case <-done:
+		case <-time.After(3 * time.Second):
+			t.Fatal("Close deadlocked against runDrainPool")
+		}
+
+		assert.NoError(t, pool.Close())
+	}
+}
+
 // mockTransport implements nativeTransport for testing
 type mockTransport struct {
 	connectedAt   time.Time


herefindalex:
Thanks for working on this. I independently tested the change against the parent commit.

  The core fix is correct: releasing `connPool.mu` before waiting on `finished` removes the deadlock described in #1990. I also forced the problematic ordering with a controlled ticker, and the old implementation deadlocks while this PR completes
  successfully.

  However, I found two concerns before merging.

  ### 1. The added regression tests do not fail against the old implementation

  I kept the tests from this PR but restored `conn_pool.go` from the parent commit (`ea072533`). Both new tests still passed, including 20 repeated runs of each test.

  In `TestConnPool_CloseDuringTickerTick`, `synctest.Wait()` runs before `Close()`. This allows the ticker handler to finish before shutdown starts, so it removes the mutex contention the test intends to reproduce.

  `TestConnPool_CloseDoesNotDeadlockWithDrainTicker` is probabilistic: the immediate `Close()` normally completes before the 1 ms ticker fires.

  A deterministic test can use a controlled ticker channel to guarantee this ordering:

  1. Queue `Close()` as the first writer waiting for `mu`.
  2. Send an unbuffered ticker event, proving `runDrainPool()` selected the ticker branch.
  3. Release the existing read lock.
  4. Assert that `Close()` completes.

  This test fails on `ea072533` and passed 100/100 times on this PR.

  ### 2. `closing` is not treated as closed by the rest of the pool

  `Get()` and `Put()` only check `closed()`, not `closing`.

  After `Close()` sets `closing = true` and releases `mu`, another goroutine can therefore acquire an idle connection while shutdown is in progress. At the public API level, `clickhouse.closed` is not set until `idle.Close()` returns, so a concurrent
  query can take an idle connection or potentially dial a new connection after `Close()` has started.

  A second direct call to `connPool.Close()` also returns immediately when `closing` is true, before the first close has finished draining the pool. The outer `clickhouse.Close()` currently masks this with `sync.Once`, but it makes the pool lifecycle
  semantics inconsistent.

  A possible minimal adjustment would be:

  - Have `Get()` and `Put()` reject operations when either `closed()` or `closing` is true.
  - Let only the first `Close()` close `finish`.
  - Make subsequent `Close()` calls wait for `finished` instead of returning immediately.

  I tested that adjustment locally with the deterministic regression tests. The package passed normally and under `-race`.

  I have not opened a competing PR; these are investigation results for consideration here.

conn_pool_sync_test.go
```diff
@@ -67,3 +67,16 @@ func TestConnPool_ExpiredConnectionsAreDrained(t *testing.T) {
...
 	})
 }
+
+func TestConnPool_CloseDuringTickerTick(t *testing.T) {
+	synctest.Test(t, func(t *testing.T) {
+		pool := newConnPool(50*time.Millisecond, 5)
+		pool.Put(&mockTransport{connectedAt: time.Now(), id: 1})
+
+		time.Sleep(50 * time.Millisecond)
+		synctest.Wait()
```
herefindalex, line 77: Does calling `synctest.Wait()` here allow the ticker handler to finish before `Close()` starts, preventing this test from exercising the target deadlock interleaving? I restored `conn_pool.go` to the parent implementation while
keeping this test, and the test still passed.

conn_pool.go
```diff
@@ -116,17 +117,21 @@ func (i *connPool) Put(conn nativeTransport) {
 
 func (i *connPool) Close() error {
 	i.mu.Lock()
-	defer i.mu.Unlock()
-
-	if i.closed() {
+	if i.closed() || i.closing {
```
herefindalex, line 120: Returning immediately when `closing` is true means a concurrent `connPool.Close()` can report success before the drain goroutine has exited and the final pool drain has completed. 
Should concurrent callers instead wait for the in-progress shutdown to complete, while ensuring that `finish` is closed only once?

PR #2002, opened 2026-09-03
fix(bind): handle fmt.Stringer in server-side query parameter switch

adamgonen:
## Summary
The switch in bindQueryOrAppendParameters handles some types explicitly but not fmt.Stringer. A named argument {p:UUID} fell through to formatValue which does quote(v.String()) so this would lead to an error like:

code: 457, message: Value '11111111-1111-1111-1111-111111111111' cannot be parsed as UUID for query parameter 'p' because it isn't parsed completely: only 32 of 38 bytes was parsed

This adds an additional case to the recent change in https://github.com/ClickHouse/clickhouse-go/pull/1978.

## Checklist

### DIFF
--- a/query_parameters.go
+++ b/query_parameters.go
@@ -80,6 +80,11 @@ func bindQueryOrAppendParameters(paramsProtocolSupport bool, options *QueryOptio
 				case *time.Duration:
 					options.parameters[p.Name] = formatDurationBody(*v)
 					continue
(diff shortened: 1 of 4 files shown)


must108 (changes requested):
Hi @adamgonen, thanks for the contribution. Just address that final nit flagged by the agent. Thanks!

must108 (approved):
LGTM

PR #2005, opened 2026-09-06
fix(churl): validate each host of a comma separated DSN authority

MaxFreedomPollard:
## Summary
`churl.Parse` is the DSN parser behind `clickhouse.ParseDSN`, and it exists so a ClickHouse HA DSN can list several hosts in the authority, for example `clickhouse://host1:9000,host2:9000/db`. `parseHost` in `lib/churl/churl.go` was still net/url's single-host validator though, so it read the whole comma separated list as one host.

That broke two things. The bracket search `strings.LastIndex(host, "[")` finds only the last IP-literal in the list, so `clickhouse://[::1]:9000,[fe80::2]:9001/db` came back as `[fe80::2]:9001` and `Options.fromDSN` ended up with `Addr: ["[fe80::2]:9001"]`. The first replica was dropped silently, with no error, so a DSN written for failover quietly connected to a single host. Separately, the optional-port check ran on the text after the list's last colon, so `clickhouse://host1:9000,host2/db` was rejected with `invalid port ":9000,host2" after host`, and so was any list whose earlier element was a bracketed literal and whose last element was named, such as `clickhouse://[fe80::1]:9000,host2:9000/db`.

`parseHost` now splits the authority on `,` and runs the unchanged single-host validation, moved to `parseSingleHost`, over each element before joining the results back. That is the same separator `Options.fromDSN` already splits the returned value on. A single-host authority takes exactly the same path as before, and a malformed element is still rejected with the same error, which the two new cases in `TestParseErrors` pin down.

Verification, all unit tests, no ClickHouse server needed:

```
$ go test -count=1 -race ./lib/... ./internal/... .
ok  	github.com/ClickHouse/clickhouse-go/v2/lib/churl	2.048s
ok  	github.com/ClickHouse/clickhouse-go/v2	1.915s
(and ok for lib/binary, lib/chcol, lib/cityhash102, lib/column, lib/column/orderedmap, lib/proto, lib/timezone, internal/circular)

$ git stash push -- lib/churl/churl.go   # the new tests against the old parseHost
$ go test -count=1 -run 'TestParseHosts|TestParseErrors' ./lib/churl/ ; go test -count=1 -run TestParseDSN .
    --- FAIL: TestParseHosts/multi-host_mixing_explicit_and_default_ports
    --- FAIL: TestParseHosts/multi-host_IPv6
    --- FAIL: TestParseHosts/multi-host_mixing_IPv6_and_named_hosts
    --- FAIL: TestParseHosts/multi-host_IPv6_with_zone_identifiers
    --- FAIL: TestParseDSN/multiple_IPv6_hosts_in_HA_mode
    --- FAIL: TestParseDSN/multiple_hosts_in_HA_mode_with_mixed_explicit_and_default_ports

$ golangci-lint run     # v2.11.0, the version .github/workflows/run-lint.yml pins
0 issues.

$ gofmt -l lib/churl/churl.go lib/churl/churl_test.go clickhouse_options_test.go
(no output)
```

## Checklist


lib/churl/churl.go
```diff
@@ -182,7 +182,28 @@ func parseAuthority(authority string) (user *neturl.Userinfo, host string, err e
 	return user, host, nil
 }
 
+// parseHost validates the host subcomponent of the authority. A ClickHouse DSN
```
must108, line 185: Please reduce the size of large comments like these. You can generally do this in a personal `AGENTS.md`

must108 (changes requested):
Overall, LGTM. Please address the 3 failing tests as well

MaxFreedomPollard:
Thanks for the review. On the three red checks, I don't think any of them are coming from this branch:

- Two are the `pull_request_target` workflow refusing to check out fork code (`Refusing to check out fork pull request code from a 'pull_request_target' workflow`). That's a repo-side setting, I can't do anything about it from a fork.
- The third, `test (1.25, 26.3)`, looks like a flake. The same server version passed on Go 1.26 and the other seven matrix cells are green. `go test ./lib/churl/...` passes locally, and the change is a pure parser fix that doesn't touch the server path.

Happy to tighten the empty-element case (`host1,,host2` and trailing commas) if you want that in this PR rather than a follow-up.

On the long review comments, I'm fairly sure those are this repo's own review workflow posting as `github-actions`, not something on my end, so I don't think an `AGENTS.md` of mine would change them. Let me know if I've got that wrong.

must108:
@MaxFreedomPollard you may address that in a follow-up, no problem. As for the comments, I am referring to the comments in code. We are working to specifying shorter in-code comments through our own `AGENTS.md`, but for now, please try to ensure they follow the rules outlined in this issue: https://github.com/ClickHouse/clickhouse-go/issues/2013.

Thanks!

must108 (approved):
LGTM! Failing test case is not due to these changes

PR #2006, opened 2026-09-06
fix(chcol): return true from HasType when the Variant carries a type

MaxFreedomPollard:
## Summary
`Variant.HasType` in `lib/chcol/variant.go` returned `v.chType == ""`, so it reported true for a `Variant` or `Dynamic` built with `NewVariant` or `NewDynamic`, which carry no type, and false for one built with `NewVariantWithType`, `NewDynamicWithType` or `WithType`, which do. That is the inverse of its doc comment ("returns true if the value has a type ClickHouse included") and of the sibling accessor `Type()`, which returns `v.chType`. This PR flips the comparison to `v.chType != ""`.

The inversion is as old as the method. `HasType` was added in 6c5abcbf as part of #1453 and first shipped in v2.30.1, so no release has ever returned the documented value.

Nothing inside this repository calls `HasType`. On `main`, `grep -rn HasType` matches only the declaration, and the method appears nowhere in `examples/`, `tests/` or the ClickHouse docs repository. A GitHub code search for Go files containing both `.HasType()` and `clickhouse-go` returns no results, and the search API's hits for `HasType` together with `chcol` or `clickhouse` are vendored copies of `variant.go` itself plus unrelated identifiers in other projects.

Breaking change assessment: this is a bug fix, not a breaking change. The signature is unchanged, the new behavior is the one the doc comment has promised since v2.30.1, no caller inside the repository depends on the inverted result, and a GitHub code search finds none outside it. The one caller who could notice is one who observed the inversion and negated the result to work around it; they would remove that negation after upgrading, which is the normal consequence of a bug fix and worth one line in the release notes.

Tests: `lib/chcol/variant_test.go` adds `TestVariantHasType`, a table test over `NewVariant`, `NewVariantWithType` (including a nil value with a type), `WithType`, `NewDynamic` and `NewDynamicWithType`. It asserts `HasType` directly and that it agrees with `Type() != ""`. Before the fix all six subtests fail with the inverted value; after the fix `go test -race -count=1 ./lib/chcol/` passes and `golangci-lint run` with the repository config reports zero issues. There is no `tests/issues/` regression test because the method is a pure in-memory accessor that never talks to a server, and there is no open issue for it.

## Checklist
- [x] Unit tests covering the common scenarios were added (no integration test: the change never touches a server)


must108 (approved):
LGTM

Issue #2016, opened 2026-09-10
missing error check in appendRowsBlocks


cyrilou242:
## Observed

appendRowsBlocks may return nil even if an error happened.
in our project we use Append with blocks, so it's getting into the appendRowsBlocks codepath. We observe incomplete streams being appended without errors. 

## Expected behaviour

Return error if a mid-stream error happens in the appendRowsBlocks codepath.

In 
https://github.com/ClickHouse/clickhouse-go/blob/a1ee16be2f6db74a3d2bd3c6f0dada2e0725871f/conn_batch.go#L138
when returning after looping on r.Next()
https://github.com/ClickHouse/clickhouse-go/blob/a1ee16be2f6db74a3d2bd3c6f0dada2e0725871f/conn_batch.go#L162

return `r.Err()` instead of `nil`.

PR #2017, opened 2026-09-10
fix: return error from appendRowsBlocks on mid-stream failure

cyrilou242:
see https://github.com/ClickHouse/clickhouse-go/issues/2016

### DIFF
--- a/conn_batch.go
+++ b/conn_batch.go
@@ -159,7 +159,7 @@ func (b *batch) appendRowsBlocks(r *rows) error {
 		lastReadLock = r.block
 	}
 
-	return nil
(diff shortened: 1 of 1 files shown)


conn_batch.go
```diff
@@ -159,7 +159,7 @@ func (b *batch) appendRowsBlocks(r *rows) error {
 		lastReadLock = r.block
 	}
 
-	return nil
+	return r.Err()
```
kavirajk, line 162: can you add some test to avoid future regression? @cyrilou242

must108 (commented):
@cyrilou242 ping

PR #2018, opened 2026-09-10
fix: missing err check after rows.Next()

cyrilou242:
Adds the missing `rows.Err()` check after `Next()` in `fetchColumnNamesAndTypesForInsert`, where a mid-stream error was silently swallowed.

note: following https://github.com/ClickHouse/clickhouse-go/pull/2017, I asked claude to look for other places with missing rows.Err() check


must108 (commented):
@cyrilou242 address agent review reasonably

PR #2024, opened 2026-09-13
fix: treat // as a single-line comment when binding query parameters

### LINKED ISSUE #1953: bind: // line comments are not recognized, so ?/$N/@name inside a // comment are substituted as placeholders
## Description

`bindQuoteState.update` (`bind.go:190-211`) recognizes ClickHouse's `--`, `#` and `#!` line comments
and `/* … */` block comments, but not `//`:

```go
case query[pos] == '#':
    // "#" and "#!" both start a single-line comment.
    s.inLineComment = true
case query[pos] == '-' && pos+1 < len(query) && query[pos+1] == '-':
    s.inLineComment = true
    return pos + 1
case query[pos] == '/' && pos+1 < len(query) && query[pos+1] == '*':
    s.blockComment++
    return pos + 1
```

The ClickHouse lexer accepts `//` as a line comment (`src/Parsers/Lexer.cpp` treats `//` the same as
`--`). Because `//` is missing from the scanner, everything after a `//` is treated as ordinary query
text by `bind()`, so `?`, `$N` and `@name` appearing inside a `//` comment are counted and substituted
as bind placeholders. Depending on the argument count the user either gets a spurious error or —
worse — a silently corrupted query where a value is interpolated into the comment body:

* too few args → `have no arg for param ? at last 1 positions` / `have no arg for "@b" param` /
  `have no arg for $2 param` for a query the server would happily run;
* enough args → the extra placeholder inside the comment is replaced with a real value and the
  positional arguments after the comment shift by one.

`bindParamsFormats` uses the same scanner, so a `?` in a `//` comment next to a `$N` elsewhere also
trips the `mixed named, numeric or positional parameters` check.

This is the Go counterpart of ClickHouse/clickhouse-java#3023, where the `jdbc-v2` ANTLR4 lexer has
the same `//` gap. Related but distinct: #1537 (same class of bug for `--`, whose repro passes today)
and #1950 (the `//` gap in the separate `PrepareBatch` insert-query regex). This issue is about
`bind.go` only.

## ClickHouse server version

Code analysis only; not verified against a running server — no ClickHouse instance was reachable in
this environment (nothing listening on `:8123`/`:9000`). The unit test below exercises `bind()`
directly, i.e. the exact function every `Query`/`Exec` call runs before any bytes reach the wire, so
the defect is confirmed independently of a server.

## Reproduction

`bind_slash_comment_test.go` in the root package:

```go
package clickhouse

import (
	"testing"

)

func TestBindSlashSlashLineComment(t *testing.T) {
	t.Run("positional", func(t *testing.T) {
		query, err := bind(nil, "SELECT ? // comment ?\n, ?", 1, 2)
		require.NoError(t, err)
		assert.Equal(t, "SELECT 1 // comment ?\n, 2", query)
	})
	t.Run("named", func(t *testing.T) {
		query, err := bind(nil, "SELECT @a // note @b\n", Named("a", 1))
		require.NoError(t, err)
		assert.Equal(t, "SELECT 1 // note @b\n", query)
	})
	t.Run("numeric", func(t *testing.T) {
		query, err := bind(nil, "SELECT $1 // $2\n", 1)
		require.NoError(t, err)
		assert.Equal(t, "SELECT 1 // $2\n", query)
	})
	t.Run("control-dash-dash", func(t *testing.T) {
		query, err := bind(nil, "SELECT ? -- comment ?\n, ?", 1, 2)
		require.NoError(t, err)
		assert.Equal(t, "SELECT 1 -- comment ?\n, 2", query)
	})
}
```

`go test -run TestBindSlashSlashLineComment -v .`:

```
--- FAIL: TestBindSlashSlashLineComment (0.00s)
    --- FAIL: TestBindSlashSlashLineComment/positional (0.00s)
        Error: Received unexpected error: have no arg for param ? at last 1 positions
    --- FAIL: TestBindSlashSlashLineComment/named (0.00s)
        Error: Received unexpected error: have no arg for "@b" param
    --- FAIL: TestBindSlashSlashLineComment/numeric (0.00s)
        Error: Received unexpected error: have no arg for $2 param
    --- PASS: TestBindSlashSlashLineComment/control-dash-dash (0.00s)
```

The `--` control case passes, which isolates the defect to the missing `//` handling.

Expected: all four subtests pass — the text after `//` is passed through untouched, exactly like
`--`, `#` and `#!`.

The same thing is reachable through the public API, e.g.

```go
row := conn.QueryRow(ctx, "SELECT $1 // why not $2\n", "hello")
// row.Err(): clickhouse [bind]: have no arg for $2 param
```

## Suggested fix

In `bindQuoteState.update` (`bind.go:208`), extend the `/` case so `//` opens a line comment while
`/*` keeps opening a block comment and a lone `/` stays division:

```go
case query[pos] == '/' && pos+1 < len(query) && query[pos+1] == '/':
    s.inLineComment = true
    return pos + 1
case query[pos] == '/' && pos+1 < len(query) && query[pos+1] == '*':
    s.blockComment++
    return pos + 1
```

The existing `inLineComment` branch already terminates on `\n`, and `//` inside a string literal is
unaffected because the `default:` arm only runs outside quoted contexts. The doc comment on
`bindQuoteState` (`bind.go:120`) should be updated to list `//` as well.

## Link

Relayed from https://github.com/ClickHouse/clickhouse-java/issues/3023

RealPratham21:
## Summary

Fixes #1953 — `bind()` did not recognize `//` line comments, so `?`, `$N` and `@name` inside a `//` comment were substituted as bind placeholders.

ClickHouse's lexer accepts `//` as a single-line comment like `--`, but `bindQuoteState.update` only handled `--`, `#`, `#!` and `/* */`.

Depending on argument count, users got a spurious `have no arg ...` error for a query the server would run fine, or a value silently interpolated into the comment body:

```go
svgconn.QueryRow(ctx, "SELECT $1 // why not $2\n", "hello")
// error: clickhouse [bind]: have no arg for $2 param
```

The fix adds a `//` case to the scanner's raw-context switch, before the `/*` case so block comments still nest and a lone `/` stays division.

`bindParamsFormats` shares the scanner, so a `?` in a `//` comment no longer trips the mixed-parameters check either.

## Tests

- `bind_test.go`: new cases in the existing positional / numeric / named / mixed-format comment suites, plus guard rails (`//` inside a block comment, lone `/` division).
- `tests/issues/issue_1953_test.go`: end-to-end regression test for all three placeholder styles across native and HTTP protocols (verified against ClickHouse 26.8.2).

## Checklist


RealPratham21:
Hi, @kavirajk & @chernser, I've accepted the CLA. I've tested and raised a PR addressing the issue, could you guys please review and let me know your feedback? Thanks.

